PR #19, opened 2014-08-04
Support for data structure that can not be used in the property list.

dealforest:
I want to use this as.
```
[Bugsnag notify: [NSException exceptionWithName:@"net.dealforest.exception" reason:@"sample" userInfo:nil]
       withData: @{ @"error": [NSError errorWithDomain:@"net.dealforest.demo" code:200 userInfo:nil] }
];
```
but this method call is failed.
and the developer will not be able to notice that it has failed.

The reason for this is using a data structure that is not supported in the property list.
(https://github.com/dealforest/bugsnag-cocoa/blob/master/bugsnag/BugsnagNotifier.m#L160)

I think the unsported object is call `-[NSObejct description]`.

Regards best.

### DIFF
--- a/bugsnag/BugsnagEvent.m
+++ b/bugsnag/BugsnagEvent.m
@@ -197,7 +197,7 @@
 
 - (NSDictionary *) toDictionary {
     @synchronized(self) {
-        [self.dictionary setObject:[self.metaData toDictionary] forKey:@"metaData"];
+        [self.dictionary setObject:[self.metaData toDescriptionDictionary] forKey:@"metaData"];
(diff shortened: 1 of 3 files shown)


snmaynard:
Now we have rewritten the notifier to use KSCrash this PR is no longer mergeable (the entire notifier was rewritten). 

I know that KSCrash has its own custom serialisation code so this might work with the new notifier straight away. If not let us know and we can have a look into adding support for this if it is appropriate.

Thanks!

dealforest:
No problem! :tada:

PR #20, opened 2014-08-17
added method for starting BugSnag with out exception handler

dannyshmueli:
Aniways is an SDK for mobile apps which have chat. If a client application is already using a bug reporting framework we do not wish to de-register their's exception handler (`NSSetUncaughtExceptionHandler`)


snmaynard:
Now that we have re-written the notifier to use KSCrash this is actually possible using a custom BugsnagConfiguration with autoNotify set to false - take a look at https://github.com/bugsnag/bugsnag-cocoa/blob/master/Source/Bugsnag/BugsnagNotifier.m#L97

Thanks for the inspiration!

PR #23, opened 2014-09-08
Fixes a bug when crash occurs in didFinishLaunchingWithOptions

mll:
When context is not set or guessed from topmost view controller ((e.g. didFinishLaunhcingWithOptions) bugsnag itself crashes and no crash report is sent


snmaynard:
We have re-written the notifier to use KSCrash and so this is now fixed!

Thanks

PR #35, opened 2015-01-09
Unity support

ConradIrwin:
:

### DIFF
--- a/Source/Bugsnag/BugsnagMetaData.h
+++ b/Source/Bugsnag/BugsnagMetaData.h
@@ -36,7 +36,7 @@
 - (void) mergeWith:(NSDictionary*)data;
 - (NSDictionary*) toDictionary;
 - (void) addAttribute:(NSString*)attributeName withValue:(id)value toTabWithName:(NSString*)tabName;
-@property(weak) id<BugsnagMetaDataDelegate> delegate;
+@property(unsafe_unretained) id<BugsnagMetaDataDelegate> delegate;
 
 @end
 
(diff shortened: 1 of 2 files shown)


Source/Bugsnag/BugsnagSink.m
```diff
@@ -144,6 +144,16 @@ - (NSDictionary*) formatEvent:(BugsnagCrashReport*) report {
     [exception safeSetObject: report.errorClass forKey: @"errorClass"];
     [exception setObjectIfNotNil: report.errorMessage forKey: @"message"];
 
+    // HACK: For the Unity Notifier. We don't include ObjectiveC exceptions or threads
```
snmaynard: This isnt a hack :-)

You could have notifier specific sink code that you call from here (so you call out to the notifier to make final adjustments to the report before its sent). This could then be used by users to solve some of the other ios issues we have.

Source/Bugsnag/BugsnagMetaData.h
```diff
@@ -36,7 +36,7 @@
 - (void) mergeWith:(NSDictionary*)data;
 - (NSDictionary*) toDictionary;
 - (void) addAttribute:(NSString*)attributeName withValue:(id)value toTabWithName:(NSString*)tabName;
-@property(weak) id<BugsnagMetaDataDelegate> delegate;
+@property(unsafe_unretained) id<BugsnagMetaDataDelegate> delegate;
```
snmaynard: what are the implications of this change - could it cause issues on 4.3?

PR #40, opened 2015-03-03
NSNotificationCenter dealloc

snmaynard:
This is a possible fix for bugsnag/bugsnag-unity#6

I think it can happen if people call start multiple times in a single app run. This is safer regardless and could possibly fix. I will try and reproduce tomorrow.


snmaynard:
Reproduced the issue by calling start multiple times, confirmed that this fixes.

Issue #43, opened 2015-04-15
Support for breadcrumbs


vocaro:
Bugsnag has a cool breadcrumb feature to help determine the sequence of events leading up to the crash:

https://bugsnag.com/blog/introducing-mobile-breadcrumbs

Breadcrumbs have been available on Android since December, but they are absent from iOS. It would be great if iOS developers could also start logging breadcrumbs.

jessicard:
Hey @vocaro - thanks for opening this! It would be awesome to add this to the notifier. We do have plans to add it eventually, but currently we have no ETA. In the meantime, of course, we're open to PRs.

BananaNeil:
+1

tinman821:
+1

joeblau:
:+1:

trobrock:
:+1:

joshdholtz:
:+1:

PR #56, opened 2015-09-17
Initial pass of breadcrumbs for iOS

kattrali:
Breadcrumbs are [supported](https://bugsnag.com/blog/introducing-mobile-breadcrumbs) on Android, as noted in #43. These changes add breadcrumb support to the iOS library, and updates the `README` with the basic usage. 

Aww yiss

## Implementation notes

* The breadcrumb interface mirrors the [android implementation](https://github.com/bugsnag/bugsnag-android/blob/97eca13c7aa604a6578a231abb83cdfa5e14bf49/src/main/java/com/bugsnag/android/Bugsnag.java#L365-L391), adjusting the selector naming slightly for platform conventions.
* `breadcrumbs` is added to the `crash` tab if any are recorded.
* I assumed breadcrumbs are intended for recording user interactions and/or view loading. As such, the `BugsnagBreadcrumbs` container enforces use of the main thread for modification.
* Breadcrumbs are stored in `BugsnagConfiguration` alongside `metaData`. I'm not sure if there is a better place for this.

## Usage questions

* Should breadcrumbs be cleared after recording an exception? What about multiple exceptions occuring without crashing the app, such as on OS X?

### DIFF
--- a/Bugsnag.xcodeproj/project.pbxproj
+++ b/Bugsnag.xcodeproj/project.pbxproj
@@ -96,6 +96,12 @@
 		79FD0E301A0C1AAB0050DBC5 /* Bugsnag.h in Headers */ = {isa = PBXBuildFile; fileRef = 79FD0D541A0C168B0050DBC5 /* Bugsnag.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		79FD0E311A0C1AAF0050DBC5 /* BugsnagMetaData.h in Headers */ = {isa = PBXBuildFile; fileRef = 79FD0D5A1A0C168B0050DBC5 /* BugsnagMetaData.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		79FD0E321A0C1AB30050DBC5 /* BugsnagConfiguration.h in Headers */ = {isa = PBXBuildFile; fileRef = 79FD0D561A0C168B0050DBC5 /* BugsnagConfiguration.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		8A4C96F11BAA2DEA007B2BFC /* BugsnagBreadcrumbsTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 8A4C96F01BAA2DEA007B2BFC /* BugsnagBreadcrumbsTest.m */; settings = {ASSET_TAGS = (); }; };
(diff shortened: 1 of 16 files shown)


snmaynard:
I love the appropriateness of that gif!

The idea for breadcrumbs is that they would largely be used by people to see how a crash occurred and to give them some ideas on reproduction. So you could use them for log like messages or user interactions - so it helps answer the how question when looking at a new crash.

Therefore I think these are independent of whether there has been a crash or not, so we shouldn't clear them when there is a crash or whatever, it should just snapshot them and send that with the bug report - so I think this is perfect!

I think that we need to make a small change to the BugsnagSink to send the data through to our servers. The code that is currently here will get KSCrash to write the crash report to disk, but then its the sink's responsibility to take that crash report and send it to Bugsnag. So you can see how severity is handled at https://github.com/bugsnag/bugsnag-cocoa/blob/master/Source/Bugsnag/BugsnagSink.m#L122 and https://github.com/bugsnag/bugsnag-cocoa/blob/master/Source/Bugsnag/BugsnagCrashReport.m#L83 for example. 

Other than that I think it's all good! I also need to take a look into RFC3339 and its compatibility with how we are parsing dates on the backend.

Awesome work @kattrali!

kattrali:
> I think that we need to make a small change to the BugsnagSink to send the data through to our servers. The code that is currently here will get KSCrash to write the crash report to disk, but then its the sink's responsibility to take that crash report and send it to Bugsnag.

Ah, got it. That makes perfect sense. I will update the request with a few adjustments to `BugsnagSink` to upload the breadcrumb data. :bread: 

> I also need to take a look into RFC3339 and its compatibility with how we are parsing dates on the backend.

Is there a better/more preferable format for this? It would be a small change to send unix timestamps instead, for example.

snmaynard:
I think we use iso8601 so that it is still human readable when inspecting payloads and it looks like rfc 3339 is actually fairly [compatible](https://en.wikipedia.org/wiki/ISO_8601#Usage) with that so we might be good as we are

kattrali:
Patched the sink to archive the breadcrumbs along with the rest of the event payload.

kattrali:
Merged manually in 7c70b627cc62117e2d6ef81951ac3b80cc4b0e65

PR #57, opened 2015-09-24
Nil name

snmaynard:
This deals appropriately with a nil name. It is hard but possible to get a nil name into NSException, and KSCrash enforces that it should not be nil.


martin308:
:+1:

PR #66, opened 2015-10-23
Use Bugsnag.config.notifyReleaseStages instead of bugsnagReport.notifyReleaseStages in BugsnagSink

hayashi311:
Bugsnag.config.notifyReleaseStages dosn't work.
bugsnagReport.notifyReleaseStages returns always nil in my enviroment.


kattrali:
Hi @hayashi311, thank you for the contribution! I took a look at this issue and it makes sense that the `BugsnagSink` reporting filtering should fall back to the global configuration if the report-specific configuration option is not set for `notifyReleaseStages`. I'm merging this pull request and making a small change to override the global config when users choose to set the report-specific `notifyReleaseStages` option.

PR #68, opened 2015-11-16
Fixes typo

ranunez:
Name of module is 'Bugsnag', not 'BugSnag'


kattrali:
Good catch! Thanks, @ranunez.

PR #69, opened 2015-11-17
Cleans up syntax with modern Objective-C

ranunez:


kattrali:
Looks good, thanks @ranunez. I plan to do a full pass at modernizing to "ObjC 2.0" where possible in the codebase.

PR #71, opened 2015-11-19
Update to the crash handler to send breadcrumbs and severity when uncaught errors occur

kattrali:
## Changes

* Refactor `serializedDictionary:toJSON:` into a function, `BSSerializeJSONDictionary`, for reuse
* Move #defines into string constants
* Rename onCrash handler to indicate purpose
* Add notifier state metadata to crash report

### DIFF
--- a/Source/Bugsnag/BugsnagNotifier.m
+++ b/Source/Bugsnag/BugsnagNotifier.m
@@ -37,8 +37,17 @@
 #import "KSSafeCollections.h"
 #import "NSDictionary+Merge.h"
 
-#define NOTIFIER_VERSION @"4.0.9"
-#define NOTIFIER_URL @"https://github.com/bugsnag/bugsnag-cocoa"
+NSString *const NOTIFIER_VERSION = @"4.0.9";
+NSString *const NOTIFIER_URL = @"https://github.com/bugsnag/bugsnag-cocoa";
+NSString *const BSTabCrash = @"crash";
+NSString *const BSTabConfig = @"config";
+NSString *const BSAttributeSeverity = @"severity";
+NSString *const BSAttributeDepth = @"depth";
+NSString *const BSAttributeBreadcrumbs = @"breadcrumbs";
+
+@interface Bugsnag ()
++ (BugsnagNotifier*)notifier;
+@end
 
 struct bugsnag_data_t {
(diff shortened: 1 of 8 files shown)


Source/Bugsnag/BugsnagNotifier.m
```diff
@@ -51,13 +64,44 @@
...
+ *  @param writer report writer which will receive updated metadata
+ */
+void BSSerializeCrashData(const KSCrashReportWriter *writer) {
     if (g_bugsnag_data.configJSON) {
         writer->addJSONElement(writer, "config", g_bugsnag_data.configJSON);
     }
     if (g_bugsnag_data.metaDataJSON) {
         writer->addJSONElement(writer, "metaData", g_bugsnag_data.metaDataJSON);
     }
+    BugsnagNotifier* notifier = [Bugsnag notifier];
```
snmaynard: I believe that this function cant call objective c code as its not async safe.

Source/Bugsnag/BugsnagNotifier.m
```diff
@@ -51,13 +64,44 @@
...
+    [notifier addCrashDataToStateWithSeverity:BugsnagSeverityError andDepth:4];
+    [notifier metaDataChanged:notifier.state];
```
snmaynard: We wouldnt need to send this message because we shouldnt actually change the shared state in this case - that would be my guess

kattrali: It wasn't the state modification which was important, just [writing the `state` JSON](https://github.com/bugsnag/bugsnag-cocoa/blob/kattrali/send-severity-breadcrumbs-with-uncaught/Source/Bugsnag/BugsnagNotifier.m#L206). It could be called directly/referenced though.

snmaynard: Yeah we could just write it out slightly modified here. I dont know how many json manipulation functions we have here though.

kattrali:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Q60HtFgo&#x2F;23-cocoa-platform-improvement"> Cocoa Platform Improvement</a>

kattrali:
Closing this out, may reopen once I get around to an alternative approach.

PR #75, opened 2015-12-16
Adding proper support for notifyReleaseStages

mikelikespie:
Notify release stages don't actually work in the current version.

Before this fix, in `[BugsnagSink filterReports:completion:]`, `bugsnagReport.notifyReleaseStages` would return nil because it was never encoded in the JSON.


kattrali:
Thanks, @mikelikespie. Looks good.

Issue #72, opened 2015-11-28
Duplicate interface definition for class 'KSCrashReportFilterPassthrough' and other


thepoortom:
Can't run Bugsnag.

I've added Bugsnag via Cocoapods and got several error messages "Duplicate interface definition for class 'KSCrashReportFilter..." in KSCrash target.

I have YandexMetrica (https://github.com/yandexmobile/metrica-sdk-ios) in my project which also uses KSCrash. In Podfile.lock YandexMetrica's dependences explicitly written:
  - YandexMobileMetrica (2.1.1):
    - FMDB (~> 2.2)
    - KSCrash/Recording (~> 0.0.5)
    - KSCrash/Reporting/Filters/AppleFmt (~> 0.0.5)
    - protobuf-c (~> 1.0.1)

I think Bugsnag pod should have similar dependences in its settings to run with other pods such as YandexMetrica which use same subpods.

kattrali:
Thanks for the report, @thepoortom. We're planning to update the library to use a prefixed version of KSCrash since this is a commonly reported issue. This should be released later this week.

thepoortom:
Thanks! I've checked update (4.1.0) but issue remains the same.

Issue #52, opened 2015-07-27
KSCrashReport.c:1996:67: error: conflicting types for 'time'


crossle:
```
Pods/Bugsnag/KSCrash/Source/KSCrash/Recording/KSCrashReport.c:1996:67: error: conflicting types for 'time'
242         writer->addIntegerElement(writer, KSCrashField_Timestamp, time(NULL));                                                                                                                                                      
243                                                                   ^                                                             
244 In module 'Darwin' imported from /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS8.4.sdk/usr/include/sys/_structs.h:29:
245 /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS8.4.sdk/usr/include/time.h:116:8: note: previous declaration is here
246 time_t time(time_t *);                                                                                                          
247        ^                                                                                                                        
248 1 error generated.    
```

crossle:
How about this?

crossle:
Version 4.0.7

crossle:
/Pods/KSCrash/Source/KSCrash/Recording/KSCrashReport.c:1998:67: Declaration of 'time' must be imported from module 'Darwin.C.time' before it is required

still have this error

PR #77, opened 2015-12-19
Add a custom prefix to KSCrash to avoid conflict, improve pod specification

### LINKED ISSUE #41: Use KSCrash dependency in CocoaPods instead of self-hosting
screen shot 2015-03-17 at 3 33 23 pm

[KSCrash](http://cocoapods.org/pods/KSCrash) is it's own independent CocoaPod, people are less likely to get class collisions in their apps. If they are including both ( I am in the tests for [ARAnalytics](http://cocoapods.org/pods/ARAnalytics) )

kattrali:
### Changes

* Updated KSCrash to latest upstream changes, applying Bugsnag's patches on top
* Adds a Bugsnag prefix to vendored classes from [KSCrash](https://github.com/bugsnag/KSCrash/tree/namespacing)
* Updated the podspec to use the new dependency as well as moving the target configuration to [cocoapods-bugsnag](https://github.com/bugsnag/cocoapods-bugsnag), a cocoapods plugin.

### Rationale

Until we/someone has the bandwidth to develop a solution to https://github.com/kstenerud/KSCrash/pull/84, this change allows client applications to simultaneously use other libraries dependent on KSCrash. The namespacing is done via a script committed to a patch of KSCrash, and it finds and updates unprefixed classes as needed.

Fixes #41, #52, #72


snmaynard:
LGTM

PR #79, opened 2015-12-22
Ignore deprecation warning for selector in BugsnagSink

endoze:
In order to avoid causing deprecation warnings in projects that include
this library, this commit adds some compiler instructions to ignore the
deprecated selector used on NSURLConnection
(sendSynchronousRequest:returningResponse:error:).


kattrali:
This looks good, thanks @endoze. I'll add a new issue for removing these deprecations for users on newer/current platforms as well.

PR #81, opened 2015-12-28
Customize/skip build phase

joshdholtz:
Three new environment variables (can define in top of `Podfile`)
- `BUGSNAG_SKIP_PHASE_SCRIPT` - don't install phase script to upload dSYM
- `BUGSNAG_PRE_PHASE_SCRIPT` - prepend stuff to script
- `BUGSNAG_POST_PHASE_SCRIPT` - append stuff to script

### Don't install phase script
```ruby
ENV['BUGSNAG_SKIP_PHASE_SCRIPT'] = 'true'
```

### Configure phase script
```ruby
ENV['BUGSNAG_PRE_PHASE_SCRIPT'] = <<-EOS
if ENV["CONFIGURATION"] != "Release"
	exit 0
end
EOS
```

### DIFF
--- a/Bugsnag.podspec
+++ b/Bugsnag.podspec
@@ -31,8 +31,8 @@ RUBY
 
     def add_bugsnag_upload_script_phase
 
-      pre_script = ENV['BUGSNAG_PRE_PHASE_SCRIPT'] || ""
-      post_script = ENV['BUGSNAG_POST_PHASE_SCRIPT'] || ""
+      pre_script = ENV['BUGSNAG_PRE_PHASE_SCRIPT'] || "\n"
+      post_script = ENV['BUGSNAG_POST_PHASE_SCRIPT'] || "\n"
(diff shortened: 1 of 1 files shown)


kattrali:
Hi, @joshdholtz. Thank you for the contribution. Unfortunately, I can't merge this directly here as we are moving to separate the build phase integration into a [cocoapods plugin](https://github.com/bugsnag/cocoapods-bugsnag) in the next release. I'm open to merging the skip phase change there however. 

The customization changes may require a bit more thought. Perhaps there is a cleaner way to specify before and after hooks rather than putting scripts into environment variables, since failing to set the variable before running `pod install` would result in multiple script phases being generated.

joshdholtz:
@kattrali No problem at all. I was just running into issues where Bugsnag was trying to upload the dsym with every build (including development) which was destroying my home internet. So I quickly through this solution together to solve my problem and figured I'd share it.

kattrali:
Gotcha, I'm glad you found the problem. :+1:

PR #86, opened 2016-02-02
Send severity and breadcrumbs with uncaught exceptions

### LINKED ISSUE #58: Missing Severity
The severity is not being sent in some situations

kattrali:
## Changes

* Set and serialize default severity as `error` in case of an uncaught exception. It is still overridden in the case of a handled exception.
* Serialize breadcrumbs after each change
* Refactor `serializedDictionary:toJSON:` into a function, `BSSerializeJSONDictionary`
* Rename onCrash handler to indicate purpose

Fixes #58

### DIFF
--- a/Source/Bugsnag/BugsnagNotifier.m
+++ b/Source/Bugsnag/BugsnagNotifier.m
@@ -39,6 +39,11 @@
 
 NSString *const NOTIFIER_VERSION = @"4.1.0";
 NSString *const NOTIFIER_URL = @"https://github.com/bugsnag/bugsnag-cocoa";
+NSString *const BSTabCrash = @"crash";
+NSString *const BSTabConfig = @"config";
+NSString *const BSAttributeSeverity = @"severity";
+NSString *const BSAttributeDepth = @"depth";
+NSString *const BSAttributeBreadcrumbs = @"breadcrumbs";
 
 struct bugsnag_data_t {
     // Contains the user-specified metaData, including the user tab from config.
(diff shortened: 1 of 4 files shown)


Source/Bugsnag/BugsnagNotifier.m
```diff
@@ -114,27 +148,29 @@ - (void)notify:(NSException *)exception withData:(NSDictionary *)metaData atSeve
...
     [self.metaDataLock unlock];
     [self metaDataChanged: self.configuration.metaData];
-    [[self state] clearTab:@"crash"];
+    [[self state] clearTab:BSTabCrash];
 
     [self performSelectorInBackground:@selector(sendPendingReports) withObject:nil];
+    [self resetDefaultSeverity];
+}
+
+- (void) resetDefaultSeverity {
```
snmaynard: Could we instead say that if severity is nil then we should set it to error in the sink?

snmaynard:
This LGTM!

Source/Bugsnag/BugsnagNotifier.m
kattrali: I wondered about that. Is it weird to modify the payloads during the filter step though?

snmaynard: I dont think so. I think we'd just be saying that if one wasnt written it was in fact an error. I dont mind which direction we head in but wanted to point it out as an option.

kattrali: Yeah, I could see that. and then I can move all of this logic as one unit. Patched.

snmaynard:
:+1: :+1: :deciduous_tree:

PR #87, opened 2016-02-04
Simplify report payload creation by putting it in one place

kattrali:
Add a `serializableValue`-type method to crash reports to create the required payload without [exposing fields](https://github.com/bugsnag/bugsnag-cocoa/pull/87/files#diff-400100c91ac4c22beb92952cd57807f8L37) unnecessarily or spreading out the logic of the internal structure. In addition, split up and add more tests for fields within the payload and format the changed files in LLVM style.

This is most easily verifiable by running the tests for the commits individually, as the suite passes before and after the serialization change.


snmaynard:
This LGTM - looks like mostly this was a reorg and the tests in this area do give me confidence as well

Issue #17, opened 2014-06-13
Flag when an app opens for the first time after it previously crashed


jessicard:
Flag when an app opens for the first time after it previously crashed. For example, the flag could be used as a way to prompt the user for help and also to avoid asking the user to rate an app if it just crashed.

trobrock:
:+1:

PR #88, opened 2016-02-04
Add user-available post-crash handler

### LINKED ISSUE #34: Support end user 'onCrash' handling
Looks like KSCrash's onCrash handler is being used up for BugSnag to do its thing. We would like a way to add another handler for recording errors through our own custom tooling.

kattrali:
Fixes #34.

### DIFF
--- a/Source/Bugsnag/BugsnagNotifier.m
+++ b/Source/Bugsnag/BugsnagNotifier.m
@@ -52,6 +52,8 @@ struct bugsnag_data_t {
     char *configJSON;
     // Contains notifier state, under "deviceState" and crash-specific information under "crash".
     char *stateJSON;
+    // User onCrash handler
+    void (*onCrash)(const KSCrashReportWriter* writer);
 };
 
 static struct bugsnag_data_t g_bugsnag_data;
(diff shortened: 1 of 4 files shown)


snmaynard:
As a separate feature/pr if required. We could add a pre-send handler. In there you could safely run objective-c code, but it wouldnt be in the crashing context. This would be good for adding user information or editing the stacktrace/grouping hash etc.

Source/Bugsnag/BugsnagNotifier.m
```diff
@@ -72,6 +74,9 @@ void BSSerializeDataCrashHandler(const KSCrashReportWriter *writer) {
     if (g_bugsnag_data.stateJSON) {
         writer->addJSONElement(writer, "state", g_bugsnag_data.stateJSON);
     }
+    if (g_bugsnag_data.onCrash) {
+        g_bugsnag_data.onCrash();
```
snmaynard: Can we sensibly pass it anything? The writer? Could we do something sensible with the return value?

Usually we pass the payload in and you can edit it. And then the return value is a bool true/false to send it or not.

kattrali:
That's a good point. I was wondering how to integrate #17 with this, and I bet that would be cleanest.

PR #89, opened 2016-02-04
Add symlink option to framework archive

### LINKED ISSUE #55: Bugsnag OS X and iOS frameworks ZIP files are created wrong
The framework ZIP archives that are downloadable for iOS and OS X do not have the sym links, instead they have copies of the files which makes my code signing fail, among other problems. 
screen shot 2015-09-10 at 12 37 27 pm

All OS X and iOS frameworks have lots of symlinks, as per https://developer.apple.com/library/mac/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/FrameworkAnatomy.html

Also it seems that downloading the source and building the frameworks seems to make the frameworks correctly, so it looks like an issue where your ZIP step is following the symlinks instead of leaving them alone.

screen shot 2015-09-10 at 12 40 12 pm

kattrali:
Fixes #55

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,8 @@ build/Release/%.framework:
 	xcodebuild -target $* build $(FORMATTER)
 
 build/Release/%-$(VERSION).zip: build/Release/%.framework
-	zip -r build/Release/$*.zip build/Release/$*.framework
+	cd build/Release; \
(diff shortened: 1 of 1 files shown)


snmaynard:
LGTM

PR #90, opened 2016-02-05
Add pre-notify hooks

### LINKED ISSUE #17: Flag when an app opens for the first time after it previously crashed

### LINKED ISSUE #47: Add Notifier Access To GroupingHash
There doesn't seem to be a way to set the groupingHash on iOS.

kattrali:
Actual change content is in a2ddc2b, previous commit is just formatting.

Fixes #17 
Fixes #47


kattrali:
Patched.

snmaynard:
LGTM!

Issue #91, opened 2016-02-08
'KSCrashReportWriter.h' file not found


ben-j69:
The project was working well, I ran pod update and then I got this error despite the file is in my project folder.

/Users/XXX/Desktop/XXXX/XXX-ios-client/Pods/Headers/Public/Bugsnag/BugsnagConfiguration.h:30:9: 'KSCrashReportWriter.h' file not found

kattrali:
Hi @ben-j69, thank you for the report. Can you post your Podfile here? I'd like to double check the version used to ensure I'm debugging the correct issue.

carrarodev:
Same error here. I've removed & installed bugsnag again after #92, but the error persist. I've also cleaned project.
Attached my podfile
![podfile](https://cloud.githubusercontent.com/assets/402719/12921547/163d4fd0-cf4e-11e5-9fab-3461e205f5ee.png)

PR #92, opened 2016-02-08
Fix missing header KSCrashReportWriter.h

mattyohe:
Fixes https://github.com/bugsnag/bugsnag-cocoa/issues/91

This resolves an issue introduced in 43683817. BugsnagConfiguration.h (a
public header) gained a new include of the KSCrashReportWriter.h
(private) header. This fix includes updates to the podpsec so that
KSCrashReportWriter.h is now public, and also reverts the project
setting "CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES" which I
assume was set to allow the test target to compile.

- Add public header in Bugsnag.podspec
- Revert project setting allowing non modular includes
- Add public header in Bugsnag target

```bash
make test
...
Executed 40 tests, with 0 failures (0 unexpected) in 0.533 (0.579) seconds
```

### DIFF
--- a/Bugsnag.podspec
+++ b/Bugsnag.podspec
@@ -22,7 +22,10 @@ Pod::Spec.new do |s|
 
   s.requires_arc = true
 
-  s.public_header_files = ["Source/Bugsnag/*.h", "KSCrash/Source/KSCrash/Reporting/Filters/KSCrashReportFilter.h"]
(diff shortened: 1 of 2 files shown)


kattrali:
Thanks, @mattyohe. I must have zoned out there. :)

PR #105, opened 2016-04-20
Made Cartfile private to stop carthage build errors

samritchie:
Currently installing Bugsnag via Carthage leads to the dreaded “Parse error: expected submodule commit SHA in output of task (ls-tree -z 1.0.0 KSCrashServer) but encountered: ”

Pretty sure this is due to KSCrashReporter being included as a submodule as well as in the Cartfile. I’ve just made the `Cartfile` private so it’s not pulled in as a transitive dependency - this seems to work.


samritchie:
Hmm, I had to bump the IPHONEOS_DEPLOYMENT_TARGET setting otherwise iTunes upload fails. Looks like tests are failing now, will try to fix.

Errortype520:
This still results in the error on my end

samritchie:
@Errortype520 until there’s a new release you’ll need to use `github "samritchie/bugsnag-cocoa" "master"`

kattrali:
Unfortunately, the deployment target cannot be changed as we support iOS 6.0+, looking into this issue though.

samritchie:
@kattrali the deployment target does not affect CocoaPods usage - 6.0 is still possible there. It will affect Carthage, but Carthage uses dynamic frameworks and needs iOS 8+ anyway. 
It will also affect the downloadable framework, but again, this is dynamic. BTW I don’t think that framework is building properly at the moment, I got missing symbols errors when I tried to use it.

kattrali:
>  BTW I don’t think that framework is building properly at the moment, I got missing symbols errors when I tried to use it.

There are missing symbols when building the workspace? Can you give an example?

> It will also affect the downloadable framework, but again, this is dynamic.

Yeah, this is the part I'm concerned about - those using the downloaded framework and building for 6.0 will see some strangeness.

samritchie:
> There are missing symbols when building the workspace? Can you give an example?

No, the built framework attached to the release is missing architectures. If you download the Bugsnag iOS zip and run:

    lipo -info Bugsnag.framework/Bugsnag

you get `Architectures in the fat file: Bugsnag.framework/Bugsnag are: armv7 arm64` - ie you won’t be able to build an app for the simulator if you have this framework linked. 

>  those using the downloaded framework and building for 6.0 will see some strangeness

You can’t use the downloaded framework and build for 6.0 - dynamic frameworks require iOS 8+ (technically you can, but you won’t be able to submit to the store). If you want to support iOS 6 without CocoaPods via a downloadable binary, you’ll need to add a static library target. This can have a separate  `IPHONEOS_DEPLOYMENT_TARGET`.

samritchie:
Will it be possible to get this PR merged, or are there other changes required? I’ve merged latest into my branch and it’s all working fin from the Carthage perspective.

kattrali:
I think this is good, thanks for the reminder and the patience @samritchie

kattrali:
I'll need to update the documentation for Carthage installation, though I haven't had much success with using Carthage in the past. I'll post an update when its live.

Issue #122, opened 2016-11-04
Expose more `BugsnagCrashReport` methods


spikebrehm:
We're upgrading `bugsnag-cocoa` from 5.3.0 to 5.6.3 in order to use the new custom stacktrace feature.   As part of that, `addBeforeNotifyHook` was deprecated, so we're updating our code to use `addBeforeSendBlock`.   However, we are running into an issue.   In that block we also log to our internal event tracking service, and we were looking at the `NSDictionary` report for metadata about the crash, including the keys `error`, `device`, `appState`.  The problem is that using the `BugsnagBeforeSendBlock` we have access to the `BugsnagCrashReport` instead of the raw `NSDictionary`, and the `BugsnagCrashReport` class doesn't expose `error`, `device`, and `appState` in its header file, so we can't access them.

What do you think about exposing `error`, `device`, and `appState` as public methods, so we can hook into them?

Thanks!

spikebrehm:
PR here: https://github.com/bugsnag/bugsnag-cocoa/pull/123

PR #123, opened 2016-11-07
Make a few BugsnagCrashReport properties public

spikebrehm:
Move a few `BugsnagCrashReport` property definitions to the header file
to make them public:

* `error`
* `device`
* `deviceState`
* `app`
* `appState`

Perhaps some more of the properties should be made public as well?

We need this in order to access this data in a `addBeforeSendBlock`
block. While upgrading from 5.3.0 to 5.6.3, we changed our code to
use `addBeforeSendBlock` instead of `addBeforeNotifyBlock`, which has
been deprecated. However, `addBeforeSendBlock` provides
`BugsnagCrashReport` as an argument, rather than a raw `NSDictionary`
containing the `error`, `device`, `deviceState`, `appState`, etc.
fields. We're consuming this information in order to log analytics data
to our own servers. By exposing these properties on
`BugsnagCrashReport`, it allows us to keep logging this information in
the `addBeforeSendBlock` block.

to: @kattrali 
cc: @lelandrichardson

### DIFF
--- a/Source/BugsnagCrashReport.h
+++ b/Source/BugsnagCrashReport.h
@@ -132,4 +132,24 @@ NSString *_Nonnull BSGFormatSeverity(BSGSeverity severity);
  *  Number of frames to discard at the top of the stacktrace
  */
 @property(nonatomic, readwrite) NSUInteger depth;
+/**
+ *  Raw error data
+ */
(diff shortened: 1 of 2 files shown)


spikebrehm:
Addresses #122

kattrali:
Looks good. Thanks, @spikebrehm!

PR #126, opened 2016-11-10
Filter reports should perform onCompletion on the filtered reports

chiahan1123:
It seems that ```BugsnagSink``` ```filterReports``` should be passing in the ```bugsnagReports``` instead of the ```reports``` to the ```KSCrashReportFilterCompletion``` in line 81.

This way returning false from the ```BugsnagBeforeSendBlock``` will actually not send the Bugsnag report as expected.

### DIFF
--- a/Source/BugsnagSink.m
+++ b/Source/BugsnagSink.m
@@ -78,7 +78,7 @@
     
     if (bugsnagReports.count == 0) {
         if (onCompletion) {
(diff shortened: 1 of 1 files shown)


kattrali:
Thank you for the contribution, @chiahan1123. I think there may have been some misunderstanding here, as at the point that the completion block is called, no more action will take place on reports, but the logged message was incorrect in saying that reports were sent.

PR #134, opened 2017-02-20
[Doc] Fixed @param names

louisdh:
Fixed two @param names. Xcode was generating warnings in a project of mine that uses Bugsnag.


kattrali:
Thanks, @louisdh!

PR #142, opened 2017-04-14
Update outdated docstrings

keith:
This updates a few docstrings that had gotten out of sync. This also
enables the Xcode build setting to warn about invalid docstrings.


keith:
Thanks 😄

PR #147, opened 2017-05-26
Fix comparison warning

sberrevoets:
After changing the deployment target for the library to 6.0,
`UIApplicationUserDidTakeScreenshotNotification` was not guaranteed to
be available anymore. However, our project's deployment target is set to
8.0 which _does_ guarantee the presence of the symbol and leads to a
warning in Xcode.


kattrali:
Thanks, scott!

PR #148, opened 2017-05-29
Ignore warning: address always evaluating to true

anderscarling:
`Warning: "Address of 'UIApplicationUserDidTakeScreenshotNotification' will always evaluate to 'true'"`

This should suppress that warning.


kattrali:
Thanks, @anderscarling! There was a prior PR for the same issue, which I just merged.

PR #154, opened 2017-07-26
Add free disk space to crash reports

### LINKED ISSUE #74: Missing disk free/total size reporting
Disk size reporting seems to be edited out since version 3.1.3, and Bugsnag site shows 0 bytes free disk for all crashes

fractalwrench:
This adds free disk space to the device state JSONObject in the crash report, mirroring the current behaviour of the Android platform.

Fixes #74


kattrali:
Looks fine. The catch here is that collecting disk space in Report is that this will be the size of the disk when the report is being sent, rather than at `notify` time. This isn't a problem (as it was how it worked before as well), just something to note in the docs.

fractalwrench:
I've opened a PR to document this in the Docs repo.

PR #155, opened 2017-07-26
Refine iOS Automatic Breadcrumbs

fractalwrench:
This alters automatic breadcrumbs so that they use a short human-readable String, rather than the Notification name.

This also enables automatic breadcrumbs for TableView selection, which appeared not to have any observers registered.

### DIFF
--- a/Source/BugsnagNotifier.m
+++ b/Source/BugsnagNotifier.m
@@ -316,7 +316,9 @@ void BSSerializeJSONDictionary(NSDictionary *dictionary, char **destination) {
 
 - (void)orientationChanged:(NSNotification *)notif {
   NSString *orientation;
-  switch ([UIDevice currentDevice].orientation) {
+  UIDeviceOrientation deviceOrientation = [UIDevice currentDevice].orientation;
+    
+  switch (deviceOrientation) {
   case UIDeviceOrientationPortraitUpsideDown:
     orientation = @"portraitupsidedown";
     break;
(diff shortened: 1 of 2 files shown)


Source/BugsnagNotifier.m
```diff
@@ -335,10 +420,16 @@ - (void)orientationChanged:(NSNotification *)notif {
...
     break;
-  case UIDeviceOrientationUnknown:
   default:
-    orientation = @"unknown";
+    return; // always ignore unknown breadcrumbs
+  }
+  
+  NSDictionary *lastBreadcrumb = [[self.configuration.breadcrumbs arrayValue] lastObject];
+    
+  if (lastBreadcrumb && [orientation isEqualToString:lastBreadcrumb[@"name"]]) {
```
kattrali: This check seems to not be working, though I think its because the last breadcrumb name isn't the correct field for a comparison. It looks like the name needs to indicate that this is an orientation breadcrumb and the orientation in `metadata` should be equal to `orientation` in the current scope.

kattrali:
Other than the one critique, this looks good! I like the new formatting for the breadcrumb names.

Source/BugsnagNotifier.m
fractalwrench: Good catch - I've updated this to check for the Orientation Change notification, and to inspect the breadcrumb metadata instead.

kattrali:
🎉

Issue #152, opened 2017-07-19
notifyError is _really_ slow


edenman:
This is the code I'm using:
```
	let isMainThead = Thread.isMainThread
	let error = NSError(domain: "co.recharge", code: 1)
	Bugsnag.notifyError(error, block: { report in
		report.errorMessage = message
		report.depth += 2 // Don't group by LogNonFatal always being at the top of the stack.
		report.addAttribute("isMainThread", withValue: isMainThead, toTabWithName: "extra")
	})
``` 
On average, 494ms on an iPhone SE.  I'd stick it in an Async block but then the stacktrace wouldn't be correct :/

fractalwrench:
Thanks for the report @edenman. It looks as though the KSCrash library which we use within the SDK is taking a while to record all the thread traces. I'll see if we can find an iPhone SE in the office to reproduce the issue more closely.

edenman:
Any update on this, @fractalwrench?  I'm happy to drop by the Bugsnag offices if you need to borrow my SE to test :)

fractalwrench:
@edenman I've profiled sending a non-fatal exception on an iPad Mini 2 running iOS 10.3.2 (Instruments trace attached).

The bottleneck is definitely in KSCrash - it takes around 600ms to write a report, of which around 1/3 was spent in writing thread information (`kscrw_i_writeAllThreads`), and the rest of which was spent serializing binary images (`kscrw_i_writeBinaryImages`). We're also currently using `kscrashreport_writeStandardReport` to record non-fatal crashes, when `kscrashreport_writeMinimalReport` and `kscrashreport_logCrash` exist, which presumably are faster at the expense of recording less information.

I'll investigate how we can speed up these two methods, or whether we can avoid them entirely. This could potentially result in a trade-off of having less information for non-fatal crashes, which we'll have to weigh up.

I'm in Bugsnag's UK office, so it may take you a while to drop by with your iPhone :)

[non_fatal_exception_trace_ipad.trace.zip](https://github.com/bugsnag/bugsnag-cocoa/files/1182505/non_fatal_exception_trace_ipad.trace.zip)

edenman:
😂 

1) I'd be totally happy with non-fatal exceptions only grabbing the current thread's stacktrace.  Not sure this is true of everyone, but it's true for my app at least.
[...]

PR #160, opened 2017-08-22
Update crash reporter

fractalwrench:
Not ready for review yet, opening so that I can review changes and also keep a log of anything that would break an upgrading user:

- Users now need to link their targets (Test app + custom keyboard) with libc++.tbd and libz.tbd (is there a way around this)?
- Only iOS workspace has been updated so far
- Tests are failing


fractalwrench:
I've updated all the workspaces and fixed the flaky iOS test (an issue with the project.pbxproj file). A summary of changes:

- Removes git submodules
- Removes KSCrash as a pod dependency, and includes the source files within the Bugsnag framework instead
- Reduce Bugsnag dependency on KSCrash to 3 files (BugsnagCrashSentry, BugsnagSink, BugsnagErrorReportApiClient), by creating a couple of wrapper classes/moving functionality into our code
- Fixes performance in #152. This turned out to be an issue with crash report JSON serialisation calling `write` thousands of times, which took about 200ms on an iPhone 7. Using a 128kb buffer has got that down to 27ms. Changes can be seen in the last few commits.

There may be some additional performance gains in allowing the user to configure whether all the thread traces are recorded, as that now seems to be the bottleneck. This could help out on older hardware. However, I think it would be best to open that in a separate ticket for reviewability purposes.

fractalwrench:
This also requires for user to link against libc++.tbd and libz.tbd for a manual installation, I believe we can set this up somehow for Carthage/Cocoapods?

kattrali:
Yep, you'll need to add it to the spec, [like so](https://github.com/kstenerud/KSCrash/blob/master/KSCrash.podspec#L15)

Source/BugsnagCrashReport.h
```diff
@@ -37,9 +37,9 @@ NSString *_Nonnull BSGFormatSeverity(BSGSeverity severity);
 @interface BugsnagCrashReport : NSObject
 
 /**
- *  Create a new crash report from a JSON crash report generated by KSCrash
+ *  Create a new crash report from a JSON crash report generated by BugsnagCrashSentry
```
kattrali: I don't think renaming this makes sense yet, since the underlying interface hasn't changed.

kattrali:
It passes! 🎉

fractalwrench:
@snmaynard I wasn't too sure what needed changing with the Podspec License/License in general. KSCrash also uses MIT - do we need to update our license, or add them as an author? Is there any action we need to take on this for other installation methods (i.e. manual/carthage)

Current Podspec: https://github.com/bugsnag/bugsnag-cocoa/blob/update-crash-reporter/Bugsnag.podspec.json

snmaynard:
The license just says include a copy of it with the source and we are doing that here so we are covered

PR #162, opened 2017-08-29
Send report on connection change

fractalwrench:
Detects when a connection is available and attempts to send any reports which are stored on disk.


Source/BSGReachability.h
```diff
@@ -0,0 +1,102 @@
...
+-(BOOL)connectionRequired; // Apple's routine.
+// Dynamic, on demand connection?
+-(BOOL)isConnectionOnDemand;
+// Is user intervention required?
+-(BOOL)isInterventionRequired;
+
+-(NetworkStatus)currentReachabilityStatus;
+-(SCNetworkReachabilityFlags)reachabilityFlags;
+-(NSString*)currentReachabilityString;
+-(NSString*)currentReachabilityFlags;
```
kattrali: There seems to be several things here we aren't using, which is adding more complexity than we need. For example, we don't have a preferred network type, so any kind of connectivity is fine.

Source/BugsnagNotifier.m
```diff
@@ -287,6 +290,21 @@ - (void) start {
...
+    self.apiReachable = [BSGReachability reachabilityWithHostName:hostUrl];
+    [self.apiReachable startNotifier];
+    
+    __weak id weakSelf = self;
+    self.apiReachable.reachableBlock = ^(BSGReachability *reachability) {
+        [weakSelf flushPendingReports];
+    };
+}
+
+- (NSString *)apiHostUrl {
```
kattrali: Its a bit unconventional for a method which returns a string to be named 'url'. This could also be scoped to `setupConnectivityListener` since its only used in the single place.

Source/BSGReachability.h
```diff
@@ -0,0 +1,102 @@
+/*
+ Copyright (c) 2011, Tony Million.
```
kattrali: Vendoring this in complicates the license. I'm also not entirely sure the license for this version of Reachability is valid given that its derived from Apple code. We could depend on it by adding it to our podspec/Cartfile/submodules, but my overall perception is that for what we need, it might be overkill.

Source/BugsnagNotifier.m
```diff
@@ -287,6 +290,21 @@ - (void) start {
 #endif
 }
 
+- (void)setupConnectivityListener {
+    NSString *hostUrl = [self apiHostUrl];
+    self.apiReachable = [BSGReachability reachabilityWithHostName:hostUrl];
```
kattrali: `hostUrl` is URL, not a hostname. It definitely works 🤔, but I'd be concerned that if the implementation of `SCNetworkReachabilityCreateWithName` ever changes, this code would break unexpectedly.

fractalwrench:
Thanks for the feedback @kattrali - I've updated the PR to use SCNetworkReachability directly.

Source/BSGConnectivity.m
```diff
@@ -0,0 +1,115 @@
...
+}
+
+- (void)connectivityChanged:(SCNetworkReachabilityFlags)flags {
+    BOOL connected = YES;
+    
+    if (!(flags & kSCNetworkReachabilityFlagsReachable)) {
+        connected = NO;
+    }
+    if (connected) { // notify via block
+        self.connectivityChangeBlock(self);
```
kylef: Since this is optional and nil by default. Might be worth check if its nil before calling it. If `connectivityChangeBlock` is nil then this will crash.

Source/BugsnagNotifier.m
```diff
@@ -287,6 +290,17 @@ - (void) start {
 #endif
 }
 
+- (void)setupConnectivityListener {
+    NSURL *url = self.configuration.notifyURL;
+    self.networkReachable = [[BSGConnectivity alloc] initWithURL:url];
+    [self.networkReachable startWatchingConnectivity];
```
kylef: Related to the other comment, since `BSGConnectivity` observes reachability on a different queue. There could be a timing problem if reachability becomes connected after starting watching and before `connectivityChangeBlock` is set below. Since `connectivityChangeBlock` will be nil it would cause a crash.

Source/BSGConnectivity.m
```diff
@@ -0,0 +1,115 @@
...
+#import <netinet6/in6.h>
+#import <arpa/inet.h>
+#import <ifaddrs.h>
+#import <netdb.h>
+
+@interface BSGConnectivity ()
+
+@property (nonatomic, assign) SCNetworkReachabilityRef reachabilityRef;
+@property (nonatomic, strong) dispatch_queue_t serialQueue;
+@property (nonatomic, strong) id reachability;
```
kylef: Why not use instancetype instead of `id`? Since this is always set to be an instance of self. Clearer for documentation purposes.

Source/BSGConnectivity.m
```diff
@@ -0,0 +1,115 @@
...
+    
+    if (self = [super init]) {
+        self.reachabilityRef = ref;
+        self.serialQueue = dispatch_queue_create("com.bugsnag.cocoa", NULL);
+    }
+    return self;    
+}
+
+- (void)dealloc {
+    [self stopWatchingConnectivity];
```
kylef: This is effectively a noop, since starting watching connectivity is retaining the `BSGConnectivity` instance. While the instance is watching connectivity will not be released until `stopWatchingConnectivity` is called and the retain count goes down (`self.reachability = nil`). Therefore, to get here, we would no longer be watching and `stopWatchingConnectivity` has no effect.

Source/BSGConnectivity.m
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Sets a callback with SCNetworkReachability
+ */
+- (void)startWatchingConnectivity {
+    SCNetworkReachabilityContext context = { 0, NULL, NULL, NULL, NULL };
+    context.info = (__bridge void *) self;
+
+    if (SCNetworkReachabilitySetCallback(self.reachabilityRef, BSGConnectivityCallback, &context)) {
+        if (SCNetworkReachabilitySetDispatchQueue(self.reachabilityRef, self.serialQueue)) {
+            self.reachability = self;
```
kylef: I think you could simplify memory management by passing a closure (which retains necessary objects or properties) into `SCNetworkReachabilitySetCallback` and a release handler.

That way you will not need to have the `reachability` property.

Something along the lines of:

```objective-c
void (^callbackBlock)(SCNetworkReachabilityFlags) = ^(SCNetworkReachabilityFlags flags) {
    // ...
};

SCNetworkReachabilityContext context = {
    .version = 0,
    .info = (callbackBlock copy),
    .release = CFRelease
};
```

Source/BSGConnectivity.m
kylef: It also means that if the someone forgot call `stopWatchingConnectivity` and the reference to the `BSGConnectivity`, `BSGConnectivity` would be leaked due to the retain cycle.

fractalwrench:
Thanks for all the review comments @kylef - they've been really useful!

@kattrali would you be able to have a look through this PR again? The way we're handling SCNetworkReachability callbacks has changed quite a bit.

kattrali:
This will require updating the manual configuration docs to add SystemConfiguration to frameworks during installation, but I think this is otherwise ready to go.

PR #164, opened 2017-09-08
Track whether an error is handled/unhandled

fractalwrench:


Source/BugsnagCrashReport.h
```diff
@@ -60,7 +60,8 @@ NSString *_Nonnull BSGFormatSeverity(BSGSeverity severity);
                               errorMessage:(NSString *_Nonnull)message
                              configuration:(BugsnagConfiguration *_Nonnull)config
                                   metaData:(NSDictionary *_Nonnull)metaData
-                                  severity:(BSGSeverity)severity;
+                                  severity:(BSGSeverity)severity
+                                 unhandled:(BOOL)unhandled;
```
kattrali: Since this interface is public, we may want to add a second constructor to avoid making this a breaking change.

Source/BugsnagCrashReport.m
```diff
@@ -309,6 +310,19 @@ - (instancetype)initWithKSReport:(NSDictionary *)report {
...
+      
+      
+      NSDictionary *recordedState = [report valueForKeyPath:@"user.eventHandledState"];
+      
+      if (recordedState) {
+          _eventHandledState = recordedState;
+      } else { // the event was unhandled.
+          _eventHandledState = @{
+                                 @"unhandled": @YES,
+                                 @"originalSeverity": BSGFormatSeverity(BSGSeverityWarning),
```
kattrali: Should this be `error`? I'm not sure when this case would happen.

Source/BugsnagCrashReport.h
fractalwrench: I've reverted the signature change here, as the extra parameter wasn't being used within the constructor, and any client calling this method will be passing a handled error.

Source/BugsnagCrashReport.m
fractalwrench: Yes - I've updated this so that unhandled events default to a severity of Error.

Source/BugsnagCrashReport.m
```diff
@@ -452,6 +471,20 @@ - (NSDictionary *)serializableValueWithTopLevelData:
...
   BSGDictInsertIfNotNil(event, self.groupingHash, @"groupingHash");
+    
+    BSGDictSetSafeObject(event, @(self.handledState.unhandled), @"unhandled");
+    
+    // serialize handled/unhandled into payload
+    NSMutableDictionary *severityReason = [NSMutableDictionary new];
+    NSString *reasonType = [BugsnagHandledState stringFromSeverityReason:self.handledState.calculateSeverityReasonType];
+    severityReason[@"type"] = reasonType;
+    
+    if (self.handledState.attrKey) {
```
kattrali: It looks like `attrValue` should be checked too, since a dict value can't be nil

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -2182,7 +2182,11 @@ void bsg_kscrashreport_writeStandardReport(BSG_KSCrash_Context* const crashConte
                                 crashContext->config.crashID,
                                 crashContext->config.processName);
 
-        bsg_kscrw_i_writeBinaryImages(writer, BSG_KSCrashField_BinaryImages);
+        // Don't write the binary images for user reported crashes to improve performance
+        if (crashContext->crash.threadTracingEnabled == true ||
+            crashContext->crash.crashType != BSG_KSCrashTypeUserReported) {
```
kattrali: Disabling writing binary images will need to be behind a flag for unity -- we still need binary images for user reported issues on regular cocoa.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -2205,13 +2209,17 @@ void bsg_kscrashreport_writeStandardReport(BSG_KSCrash_Context* const crashConte
...
-            bsg_kscrw_i_writeAllThreads(writer,
-                                    BSG_KSCrashField_Threads,
-                                    &crashContext->crash,
-                                    crashContext->config.introspectionRules.enabled,
-                                    crashContext->config.searchThreadNames,
-                                    crashContext->config.searchQueueNames);
-            bsg_kscrw_i_writeError(writer, BSG_KSCrashField_Error, &crashContext->crash);
+            // Don't write the threads for user reported crashes to improve performance
+            if (crashContext->crash.threadTracingEnabled == true ||
+                crashContext->crash.crashType != BSG_KSCrashTypeUserReported) {
```
kattrali: Same as above

Issue #170, opened 2017-09-22
Forced crash not uploaded when using 5.11.2


edenman:
### Expected behavior
I'm using `fatalError` to force a crash during a very specific part of my app that only I can access.  I expect the crash to be uploaded to Bugsnag.

### Observed behavior
I just pushed a build to TestFlight that uses Bugsnag pod version `5.11.2`, and this crash does not seem to be uploaded.  https://app.bugsnag.com/recharge-labs/ios/errors/59c430573ed28e0019e60b42 is a previous occurrence of the issue.  I relaunched the app after the forced crash, waited a few minutes, and the new occurrence still does not show up in Bugsnag's web UI.

fractalwrench:
Thanks for the report @edenman - I've confirmed that this is a regression in v5.11.2. I'm looking into this as a priority.

PR #171, opened 2017-09-26
Fix error class parsing

fractalwrench:
The `errorClass` field can be nil if data is not supplied by KSCrash. This alters how we parse the data so that the method will never return nil, which would reject the payload.

See #170


Source/BugsnagCrashReport.m
```diff
@@ -64,18 +64,24 @@
 }
 
 NSString *BSGParseErrorClass(NSDictionary *error, NSString *errorType) {
```
kattrali: We could annotate this return value as non-null to change having null returned to a build error.

PR #174, opened 2017-09-27
Remove use of deprecated NSProcessInfo property

sberrevoets:
I'm getting a warning on this line because `operatingSystemName` has been deprecated since iOS 8, apparently because it always returns `NSMACHOperatingSystem` instead.

If you're more comfortable wrapping this in a `#if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_8_0` conditional I'd be happy to do that as well, but given the warning it didn't seem like this value is of much use.


kattrali:
@sberrevoets Thanks, Scott! This seems fine.

snmaynard:
@kattrali can we still set an osname please?

kattrali:
Yep! was going to set each platform to something human-readable (macOS, etc), though its worth noting that we don't show this in related project types.

Issue #151, opened 2017-07-18
Stacktrace on notifyError


edenman:
I'm trying to migrate to Bugsnag and I'm having trouble figuring out the right way to log non-fatal errors.  Here's what I've got:

```
func LogNonFatal(_ message: String) {
	let isMainThead = Thread.isMainThread
	let error = NSError(domain: "co.recharge", code: 1)
	Bugsnag.notifyError(error, block: { report in
		report.errorMessage = message
		report.depth += 1 // Don't group by LogNonFatal always being at the top of the stack.
		report.addAttribute("isMainThread", withValue: isMainThead, toTabWithName: "extra")
	})
}
```
I've tried changing that `depth` value between 1 and 3, but none of them give me the desired result (displaying and grouping by stacktrace).  https://app.bugsnag.com/recharge-labs/ios/errors?filters[event.since][0]=30d&filters[error.status][0]=fixed is my project if you want to look at what it's generated.  

Do I need to manually attach the stacktrace?  Or somehow otherwise indicate that it should be captured?

kattrali:
Thanks for the report, @edenman. I'm taking a look now. My guess is the depth still is off a bit, which might be a miscalculation in the library. Based on the [most recent instance](https://app.bugsnag.com/recharge-labs/ios/errors/596e67b9b1f76f0018660c09), it seems to need to be incremented by 1. Can you try 3 or 4?

edenman:
I changed the messages so it's easier to tell what happens with various depths:

depth 4: https://app.bugsnag.com/recharge-labs/ios/errors/596fa660e303b40019d75739 (two different calls grouped together, no useful stacktrace)
depth 3: https://app.bugsnag.com/recharge-labs/ios/errors/596e63fd825d9100180bfc02 (two different calls grouped together, no useful stacktrace)
depth 2: https://app.bugsnag.com/recharge-labs/ios/errors/596e626bde94bf00198d4659 and https://app.bugsnag.com/recharge-labs/ios/errors/596e67b9b1f76f0018660c09 (grouping is fixed, but stacktraces are somehow even less useful?)
depth 1: https://app.bugsnag.com/recharge-labs/ios/errors/596e61739ec2d600193dc0ee (two different calls grouped together, no useful stacktrace)

fractalwrench:
[...]

PR #175, opened 2017-09-28
Update stack frame depth

fractalwrench:
By default we now discard 5 frames from the stacktrace in `[BugsnagNotifier notify:message:block]`. Calls in `Bugsnag.m` have also been updated to increase that depth by 2, as there will be a frame for both the public API and the internal notifier API.

I tested out manually changing the depth as in #151, and increasing the depth by 1 for each function call worked on both Objective-C and iOS. We'll need to update our docs on this unless I've missed something: https://docs.bugsnag.com/platforms/ios-objc/reporting-handled-exceptions/#depth


kattrali:
This makes sense. Given this changeset, the end user would need to instead do this, correct:

```
void logError(NSError *error) {
    [Bugsnag notifyError:error block:^(BugsnagCrashReport *report) {
        // Increment for the logError function
        report.depth += 1;
    }];
}
```

?

fractalwrench:
@kattrali yep, that's what I've tested out (in Swift)

PR #185, opened 2017-10-10
Update podspec ios to min supported version of 8.0

### LINKED ISSUE #184: Xcode 8.3.3 warning due to targeting iOS 6.0
### Expected behavior
No warnings when building an Xcode project that uses CocoaPods to depend on Bugsnag.

### Observed behavior
Xcode displays a "Buildtime" warning on Pods.xcodeproj stating "Update to recommended settings".  Clicking on the warning displays a view with an issue in Bugsnag:  "An iOS Deployment Target earlier than 8.0 is not supported by this version of Xcode. This will update the value for Target 'Bugsnag' to '8.0'."

### Steps to reproduce
Add `pod 'Bugsnag'` to your Podfile
Run `pod install`
Load your Xcode project and build

### Version
Xcode 8.3.3
Bugsnag 5.11.2

fractalwrench:
Fixes #184


DavidPotter:
Thanks.  When do you think this will be published?

PR #186, opened 2017-10-11
Simplify release versioning

kattrali:
Changes the release handling to add a `make bump` target, so nobody needs to manually edit the files to update version numbers. Usage:

```sh
make VERSION=5.13.0 bump
```

It could be fancy and read the version number from the top entry of the changelog, what do you think, @fractalwrench?


fractalwrench:
I think this makes sense as is - lgtm!

Issue #159, opened 2017-08-17
Propagate the message from fatalError/preconditionFailure/assertionFailure into the Bugsnag issue


edenman:
I commented on #132 but got no response.  It'd be really helpful to do something like https://github.com/getsentry/sentry-cocoa/pull/188/files to get the message from forced crashes.

kattrali:
Thanks for the bump, @edenman. We'll take a look.

kattrali:
Fixed in v5.13.0

edenman:
This rules.  Thanks so much!

PR #188, opened 2017-10-12
Parse notable addresses for Swift Assertion failure messages

fractalwrench:
Enables memory introspection in KSCrash, which allows us to parse the notable addresses for Swift Assertion failure messages, as requested in #159. The message should now be captured for `fatalError`, `preconditionFailure`, `assertionFailure`, and `assert`.

This approach relies on heuristics - we'll inspect the variable values, and set the error message in the following conditions:

- "fatal error" or "assertion failure" is present in at least one variable
- At least one variable is a string that contains 2 or less "/"

If this isn't the case, we won't alter the error message.

### DIFF
--- a/Source/BugsnagCrashSentry.m
+++ b/Source/BugsnagCrashSentry.m
@@ -23,7 +23,6 @@ NSUInteger const BSG_MAX_STORED_REPORTS = 12;
 
     BugsnagSink *sink = [[BugsnagSink alloc] initWithApiClient:apiClient];
     [BSG_KSCrash sharedInstance].sink = sink;
-    // We don't use this feature yet, so we turn it off
     [BSG_KSCrash sharedInstance].introspectMemory = YES;
     [BSG_KSCrash sharedInstance].deleteBehaviorAfterSendAll =
         BSG_KSCDeleteOnSucess;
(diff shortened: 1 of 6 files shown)


Source/BugsnagCrashSentry.m
```diff
@@ -24,7 +24,7 @@ - (void)install:(BugsnagConfiguration *)config
     BugsnagSink *sink = [[BugsnagSink alloc] initWithApiClient:apiClient];
     [BSG_KSCrash sharedInstance].sink = sink;
     // We don't use this feature yet, so we turn it off
```
kattrali: Should change/remove the comment here

Tests/BugsnagCrashReportTests.m
```diff
@@ -127,4 +127,73 @@ - (void)testNotifyReleaseStagesSkipsSendFromConfig {
...
+                          @"value": @"Whoops - fatalerror"
+                          };
+    XCTAssertEqualObjects(@"Whoops - fatalerror", [errorReport enhancedErrorMessageForThread:thread]);
+    
+    
+    // ignores additional dict if more than 2 "/" present
+    addresses[@"r24"] = @{
+                          @"address": @4511084983,
+                          @"type": @"string",
+                          @"value": @"/usr/include/lib/something.swift"
```
kattrali: Is there any value in including these addresses in other metadata?

fractalwrench: Quite possibly - I'll look into what sort of values are available for our Swift/Obj-C examples.

PR #191, opened 2017-10-13
Sync crash report

fractalwrench:
Sends the crash report synchronously if the crash occurred within a configurable time period since app launch


Source/BugsnagErrorReportApiClient.m
```diff
@@ -114,35 +95,56 @@ - (void)sendReports:(NSArray<BugsnagCrashReport *> *)reports
     }
 }
 
-@end
-
-@implementation BSGDelayOperation
```
kattrali: This was added in https://github.com/bugsnag/bugsnag-cocoa/commit/42944e0032c0b31011430ae7795c6df6f10de14b#diff-3ae0af9a6224b0b591a4a57c922e273e to avoid a case where several repeated notify calls or exceptions were either dropped or delivered twice. Is there comparable handling for this now elsewhere for the asynchronous delivery case?

Tests/BugsnagConfigurationTests.m
```diff
@@ -52,4 +52,16 @@ - (void)testNotifyReleaseStagesExcludedSkipsSending {
...
 }
 
+- (void)testLaunchThreshold {
+    BugsnagConfiguration *config = [BugsnagConfiguration new];
+    XCTAssertEqual(10000, config.launchCrashThresholdMs);
+    
+    config.launchCrashThresholdMs = 0;
+    XCTAssertEqual(0, config.launchCrashThresholdMs);
+    
+    int expected = 1500;
```
kylef: `expected` here is `int` type yet we are comparing it to `launchCrashThresholdMs` which is `NSUInteger`.

PR #192, opened 2017-10-15
Update supported iOS Platform in documentation

kylef:
In #184 the minimum iOS target was moved to 8.0 but the README was not updated. This PR does just that, I've also fixed other platforms:

- macOS not `OS X`.
- Added tvOS.
- iOS 8 not 6.


fractalwrench:
Thanks for catching this @kylef! I'll try and look through the other PRs soon.

PR #193, opened 2017-10-15
Fix potential memory leaks inside BSGConnectivity

kylef:
I'll comment on the cases below in source.

### DIFF
--- a/Source/BSGConnectivity.m
+++ b/Source/BSGConnectivity.m
@@ -50,7 +50,7 @@ static void BSGConnectivityCallback(SCNetworkReachabilityRef target,
 
 @property(nonatomic, assign) SCNetworkReachabilityRef reachabilityRef;
 @property(nonatomic, strong) dispatch_queue_t serialQueue;
-@property(nonatomic) CallbackBlock callbackBlock;
+@property(nonatomic, copy) CallbackBlock callbackBlock;
 
 @end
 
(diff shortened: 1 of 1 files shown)


Source/BSGConnectivity.m
```diff
@@ -50,21 +50,19 @@ @interface BSGConnectivity ()
...
+@property(nonatomic, copy) CallbackBlock callbackBlock;
 
 @end
 
 @implementation BSGConnectivity
 
 - (instancetype)initWithURL:(NSURL *)url
                 changeBlock:(ConnectivityChange)changeBlock {
-    NSString *hostName = [url absoluteString];
-    SCNetworkReachabilityRef ref =
```
kylef: In the case that `[super init]` returns nil, the `ref` would never be released and thus cause a leak.

Source/BSGConnectivity.m
```diff
@@ -50,21 +50,19 @@ @interface BSGConnectivity ()
...
-        SCNetworkReachabilityCreateWithName(NULL, [hostName UTF8String]);
-
     if (self = [super init]) {
-        _connectivityChangeBlock = changeBlock;
-        self.reachabilityRef = ref;
+        NSString *hostName = [url absoluteString];
+        _reachabilityRef = SCNetworkReachabilityCreateWithName(NULL, [hostName UTF8String]);
+
+        _connectivityChangeBlock = [changeBlock copy];
```
kylef: `connectivityChangeBlock` property is annotated as copy and thus should be a copy.

Source/BSGConnectivity.m
```diff
@@ -50,21 +50,19 @@ @interface BSGConnectivity ()
 
 @property(nonatomic, assign) SCNetworkReachabilityRef reachabilityRef;
 @property(nonatomic, strong) dispatch_queue_t serialQueue;
-@property(nonatomic) CallbackBlock callbackBlock;
+@property(nonatomic, copy) CallbackBlock callbackBlock;
```
kylef: Just as `connectivityChangeBlock`, the `callbackBlock` set should be copied from stack to heap.

Source/BSGConnectivity.m
```diff
@@ -84,9 +82,6 @@ - (void)dealloc {
         CFRelease(self.reachabilityRef);
         self.reachabilityRef = nil;
     }
-
-    _connectivityChangeBlock = nil;
```
kylef: We're using ARC, there is no need for setting these to `nil`.

Source/BSGConnectivity.m
```diff
@@ -95,8 +90,10 @@ - (void)dealloc {
 - (void)startWatchingConnectivity {
     SCNetworkReachabilityContext context = {
         .version = 0,
-        .info = (void *)CFBridgingRetain(self.callbackBlock),
-        .release = CFRelease};
+        .info = (__bridge void *)self.callbackBlock,
```
kylef: This change will prevent the static analyser from detecting this area as a potential memory leak. As we're passing in the retain AND release function to `SCNetworkReachabilitySetCallback` it will be responsible for both retaining and releasing.

Otherwise the static analyser will detect that we are calling `CFBridgingRetain` ourselves and flag as a potential leak as it never sees it released.

PR #197, opened 2017-10-24
Externalise string literals

fractalwrench:
Refactors string literals to consolidate definitions which appear in multiple places.


Tests/BugsnagConfigurationTests.m
```diff
@@ -1,5 +1,6 @@
 #import "Bugsnag.h"
 #import "BugsnagConfiguration.h"
+#import "BugsnagKeys.h"
```
kattrali: We should not use the keys file within the tests. The tests should validate values rather than show equality with the values being tested. For example, if a typo occurred in the keys file changing the value of `BSGKeyType` to `"typ"`, the test would not break but should.

fractalwrench:
Thanks @kattrali - I've resolved the issue with testing.

Issue #199, opened 2017-10-27
Are breadcrumbs not thread-safe?


edenman:
https://app.bugsnag.com/recharge-labs/ios/errors/59f2781c96d78c00199afa65

### Expected behavior
non-fatal error is logged to the server, even if somebody tries to add a breadcrumb while we're in the middle of a `notifyError` call.

### Observed behavior
SIGABRT
https://app.bugsnag.com/recharge-labs/ios/errors/59f2781c96d78c00199afa65
The top level stacktrace just has one useful line #: `AppDelegate.swift:226` which is a call to `Bugsnag.leaveBreadcrumb(withMessage: logMessage.message)`
In the "threads" view, thread 19 is in the middle of doing a `Bugsnag.notifyError`

### Steps to reproduce
Unclear.  I have not been able to repro this myself.  But I imagine if you made an infinite loop that just called `leaveBreadcrumb` and then occasionally fired off calls to `notifyError`, you'd eventually trigger it.

### Version
5.13.0

### Additional information
I made the change to run notifyError from a background thread in a version of our app that went live on 10/15, and it just today (10/26) triggered this error, so I'm guessing it's a pretty rare race condition-y situation.

fractalwrench:
Thanks for the report @edenman - I'll give those reproduction steps a go.

fractalwrench:
The following code reproduces the issue after a few thousand iterations:
    
    for (int k = 0; k < 10000; k++) {
        NSLog(@"Attempt: %d", k);
        [Bugsnag leaveBreadcrumbWithMessage:@"generate non-fatal exception"];
        
        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0), ^{
            NSError *error = [NSError errorWithDomain:@"thread test" code:0 userInfo:nil];
            [Bugsnag notifyError:error];
        });
    }

This also seems to apply to other methods in our public API.

fractalwrench:
Fixed in v5.13.2

PR #201, opened 2017-10-30
Thread safety fixes

fractalwrench:
Adds locks for methods which could crash, and makes public API properties atomic. Crashes in the library can be reproduced by using the following approach:

```
for (int k = 0; k < 10000; k++) {
    NSLog(@"Attempt: %d", k);
    [Bugsnag leaveBreadcrumbWithMessage:@"generate non-fatal exception"];
    
    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0), ^{
        NSError *error = [NSError errorWithDomain:@"thread test" code:0 userInfo:nil];
        [Bugsnag notifyError:error];
    });
}
```


fractalwrench:
This should address the following issues:

https://github.com/bugsnag/bugsnag-cocoa/issues/199
https://github.com/bugsnag/bugsnag-react-native/issues/174

sberrevoets:
On Xcode 9.1 we're seeing a number of warnings related to this change:

`Writable atomic property 'capacity' cannot pair a synthesized getter with a user defined setter. Setter and getter must both be synthesized, or both be user defined,or the property must be nonatomic.`

There are 5 other properties this applies to as well.

kattrali:
@sberrevoets I've got a fix for this in https://github.com/bugsnag/bugsnag-cocoa/commit/7670931aa056227ffcc1caa96c60e3a52f1360ed, also turned on warnings as errors.

sberrevoets:
Awesome, thanks! 🙏

PR #202, opened 2017-10-30
Make dateformatter thread local

fractalwrench:
https://github.com/bugsnag/bugsnag-react-native/issues/173

### DIFF
--- a/Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_RFC3339DateTool.m
+++ b/Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_RFC3339DateTool.m
@@ -29,22 +29,19 @@
 static NSString *const kDateFormatterKey = @"RfcDateFormatter";
 
 + (NSDateFormatter *)sharedInstance {
-    @synchronized(self) {
+    NSMutableDictionary *threadDict = [[NSThread currentThread] threadDictionary];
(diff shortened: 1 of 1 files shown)


Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_RFC3339DateTool.m
```diff
@@ -26,29 +26,39 @@
...
 
-+ (void)initialize {
-    bsg_g_formatter = [[NSDateFormatter alloc] init];
-    NSLocale *locale =
-        [[NSLocale alloc] initWithLocaleIdentifier:@"en_US_POSIX"];
-    [bsg_g_formatter setLocale:locale];
-    [bsg_g_formatter setDateFormat:@"yyyy'-'MM'-'dd'T'HH':'mm':'ss'Z'"];
-    [bsg_g_formatter setTimeZone:[NSTimeZone timeZoneForSecondsFromGMT:0]];
++ (NSDateFormatter *)sharedInstance {
+    @synchronized(self) {
```
kattrali: Doesn't the thread dict negate the need to synchronize access? `self` here is the class rather than an instance.

fractalwrench: @kattrali that's true - I've updated the PR.

PR #208, opened 2017-11-06
Fix for xcodebuild tool compilation error on version 8.1(8B62)

idcuesta:
## Description
With latest changes introduced around wrapping constant names into external files, Bugsnag library stopped compiling on old Xcode versions (I was using `8.1(8B62)`). We still use old Xcode version to run Appium tests, as there is some migration needed to get test running on Xcode 9 and we are facing other issues, like slow and unstable builds.

We only need to remove the unneeded wrapper string that is initialised with another non-constant at compile time string and then everything works fine.


ppaulojr:
The same problem happens on Circle CI, please merge this PR.

fractalwrench:
Merged and released in v5.13.3 - thanks for the contribution @idcuesta!

idcuesta:
Thank you for ultra fast response @fractalwrench 👍

PR #210, opened 2017-11-08
Fix potential buffer overflow when writing large reports

### LINKED ISSUE #209: Buffer Overflow Crash when sending large amounts of `report.metaData`
### Expected behavior
Bugsnag properly logs error

### Observed behavior
bugsnag-cocoa crashes with a buffer overflow

### Steps to reproduce
Crash occurs [here](https://github.com/bugsnag/bugsnag-cocoa/blob/master/Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSFileUtils.c#L88) in `BSG_KSFileUtils.c`on when sending large amounts of data
on a logged snag's `report.metaData` field.

```c
bool bsg_ksfuwriteBytesToFD(const int fd, const char *const bytes,
                            ssize_t length) {
    ssize_t newLen = bufferLen + length;

    if (newLen >= BUFFER_SIZE) { // flush the current buffer
        bsg_ksfuflushWriteBuffer(fd);
        newLen = bufferLen + length; // recalculate new position
    }

    for (ssize_t k = bufferLen; k < newLen;
         k++) { // place in buffer for future write
        ssize_t j = k - bufferLen;
        charBuffer[k] = bytes[j]; // CRASHES HERE
    }
    bufferLen += length;
    return true;
}
```

Getting a crash in the above mentioned function in `BSG_KSFileUtils.c`.

### Version
At least starting at 5.11.2 - 5.13.3 (but could have started earlier)

### Additional information
Sorry for the brief issue - will come back with more details soon.

fractalwrench:
Fixes #209 

The original issue can be fixed by passing a very large byte array buffer to the fileUtils function - in practice this can be triggered with metaData as below:

```
[Bugsnag notify:[NSException exceptionWithName:@"wat" reason:nil userInfo:nil] 
block:^(BugsnagCrashReport * _Nonnull report) {
        
    for (int k = 0; k < 10000; k++) {
        NSString *val = [NSString stringWithFormat:@"user %d", k];
        [report addMetadata:@{@"tab": @{val: @"rar\"watrar"}} toTabWithName:@"tab"];
    }
}];
```

The implementation has been simplified so that the buffer size is checked each time before copying a byte.

### DIFF
--- a/Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSFileUtils.c
+++ b/Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSFileUtils.c
@@ -44,7 +44,7 @@
 #define BSG_KSFU_WriteFmtBufferSize 1024
 #endif
 
-#define BUFFER_SIZE 131072
+#define BUFFER_SIZE 65536
 
 char charBuffer[BUFFER_SIZE];
 ssize_t bufferLen = 0;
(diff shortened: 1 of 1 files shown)


Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSFileUtils.c
```diff
@@ -75,19 +75,14 @@ bool bsg_ksfuflushWriteBuffer(const int fd) {
...
-    }
-
-    for (ssize_t k = bufferLen; k < newLen;
-         k++) { // place in buffer for future write
-        ssize_t j = k - bufferLen;
-        charBuffer[k] = bytes[j];
+    
+    for (ssize_t k = 0; k < length; k++) {
+        if (bufferLen >= BUFFER_SIZE) {
+            bsg_ksfuflushWriteBuffer(fd); // flush to disk
```
benasher44: If `bsg_ksfuflushWriteBuffer` fails and returns `false`, `bufferLen` won't be updated, and then it'll overflow on the next write to `charBuffer `. Maybe we should bail out and return `false` in that case?

anayini: Ah @benasher44 beat me to it.  Yeah I realized this function always returns `true`

PR #212, opened 2017-11-15
Exclude address, mach, and signal from non-fatal error reports

fractalwrench:
Excludes the address, mach, and signal fields from the metadata of non-fatal error reports. These were always reported as EXC_CRASH and SIGABRT by default, which could be misleading.


kattrali:
Tested this a bit and moved the metadata down even further. The set values are correct in most cases but it should reduce confusion.

PR #219, opened 2017-12-07
Add kscrash tests

fractalwrench:
Adds the KSCrash test suite to the project, bringing code coverage up to 42% from 24%.

It's worth noting that the following [test case](https://github.com/bugsnag/bugsnag-cocoa/blob/5ba7fe2452afc8c06d9d6e7ccd208cf64c4083ec/Tests/KSCrash/KSObjC_Tests.m#L800) is failing - possibly due to some change in the underlying NSArray implementation? A couple of the JSON encoding tests have also been updated to handle control characters. The KSFileUtils test case for large read/writes has also been removed, as it caused the build to continue for 3 hours and throw an internal error on BuddyBuild 😭


fractalwrench:
@kattrali It looks like there's an issue with how KSCrash represents `NSMutableArray` in iOS 11. The offending piece of code looks like this:

```
NSMutableArray* array = [@[@"1", @"2", @"3", @"4"] mutableCopy];
void* arrayPtr = (__bridge void*)array;
size_t expectedCount = [array count];
size_t count = bsg_ksobjc_arrayCount(arrayPtr);

// called by bsg_ksobjc_arrayCount
static inline size_t nsarrayCount(const void *const arrayPtr) {
    const struct NSArray *array = arrayPtr;
    CFIndex cfIndex = array->basic.count;
    return cfIndex < 0 ? 0 : (size_t) cfIndex;
}

struct NSArray {
    struct {
        void *isa;
        CFIndex count;
        id firstEntry;
    } basic;
};
```

The issue is that `array->basic.count` always returns 0 for a mutable array, which suggests that something changed in the implementation between iOS 10 and 11. This could potentially cause us to miss data from crash reports as arrays will erroneously be thought of as empty.

fractalwrench:
@kattrali I've extracted the failing nsmutablearray test to its [own PR](https://github.com/bugsnag/bugsnag-cocoa/pull/221), so that there aren't any obstacles to getting this reviewed and merged to master.

fractalwrench:
@kattrali I've moved all the failing tests into a separate branch so we can address them separately - it'd be good to get this one reviewed as it will improve the code coverage quite a bit.

PR #220, opened 2017-12-16
Avoid generating and parsing invalid JSON files

### LINKED ISSUE #218: [Crash] EXC_BAD_ACCESS in `bsg_kscrashstate_i_onIntegerElement`on Bugsnag startup, attempting to load persistent state of crash context
### Expected behavior

Calling `-[Bugsnag startBugsnagWithApiKey:]` doesn't crash.

### Observed behavior

Calling `-[Bugsnag startBugsnagWithApiKey:]` crashes.

### Steps to reproduce

Unsure, but this is happening on application launch. We call `-[Bugsnag startBugsnagWithApiKey:]` in the `AppDelegate`. 

The crash is happening [here](https://github.com/bugsnag/bugsnag-cocoa/blob/516333a8553b3417a857a702b43698c97df9ced9/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashState.c#L106), in `bsg_kscrashstate_i_onIntegerElement` in the call to [`strcmp`](http://www.cplusplus.com/reference/cstring/strcmp/). `name` is `NULL`, which seems to be the issue.

### Version

`5.13.3`, `5.14.1` (possible that earlier version are affected)

### Stack trace

screen shot 2017-12-04 at 11 25 47 am

kattrali:
In the case that the crash state file fails to be written completely, avoid flushing the buffer as the file contents will be inconsistent. In addition, when parsing the file, add validation to avoid comparing a string to NULL.

To reproduce:
* Install the sample app with v5.14.1 onto a device
* Download the application container and show package contents
* Change the contents of `AppData/Library/Caches/KSCrashReport/{app name}-CrashState.json` to `3 }`. This triggers the code path where the file is invalid but it is not rejected before crashing.
* Replace the app container on the device with the altered one
* Restart app

Fixes #218

### DIFF
--- a/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashState.c
+++ b/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashState.c
@@ -266,11 +266,11 @@ bool bsg_kscrashstate_i_saveState(const BSG_KSCrash_State *const state,
         goto done;
     }
     result = bsg_ksjsonendEncode(&JSONContext);
-
-done:
     if (!bsg_ksfuflushWriteBuffer(fd)) {
         BSG_KSLOG_ERROR("Failed to flush write buffer");
     }
+
+done:
     close(fd);
 
     if (result != BSG_KSJSON_OK) {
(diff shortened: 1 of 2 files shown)


jessesquires:
Thanks so much @kattrali ! 🙌

Issue #224, opened 2018-01-10
Crashes on app startup don't have stacktrace/version/anything


edenman:
### Expected behavior
Crashes have a minimum set of information about what went wrong

### Observed behavior
Certain crashes soon after app startup have no information: https://app.bugsnag.com/recharge-labs/ios/errors/5a53e4e1931c7b0018ff868a (missing stacktrace, device information, app information, user information) 

### Steps to reproduce
Crash on startup

### Version
5.14.0

I contacted support about this and got this response:

_Errors like these show up when Bugsnag detects a crash that does not have an associated stacktrace. It's likely that the crash occurred so immediately that a stacktrace could not be generated before the application exited._

_Unfortunately, there's no reliable way to gather more information when an iOS application crashes on startup so Bugsnag just sends an event to at least highlight the occurrence._

So
1) I don't understand the technical reason why the stacktrace and app/device info are unable to be captured.  Can somebody elaborate?
2) Can you point me to the code that causes the "just sends an event to at least highlight the occurence" procedure to happen?  

- It was super-frustrating to know that the app was crashing, but not have literally any useful information to figure out why.  
- Also frustrating: I have since released an app version that fixes the issue, but marking the Bugsnag issue as "Fixed" doesn't have any effect because there's no app version attached, so Bugsnag's web interface doesn't know that new occurrences are from the old app.  I could "Ignore" the issue, but then I'm at risk of a subsequent app version having this sort of crash, and never know that it's happening.

fractalwrench:
There are a few scenarios where stacktraces can't be captured:

### Failed allocation errors
`std::bad_alloc` can be thrown when the system is out of memory. Bugsnag uses the KSCrash library internally, which cannot capture C++ exceptions from linked libraries. This is because these linked libraries use the system `__cxa_throw` function rather than a custom one installed by KSCrash, meaning no stacktrace is available.

### Missing unwind data
This [technical note](https://developer.apple.com/library/content/technotes/tn2151/_index.html) details another scenario where stack unwind data is not available.

[...]

PR #239, opened 2018-02-07
Improve minimal report app/device information

fractalwrench:
Changes report generation so that when a minimal or incomplete crash is recorded, essential app/device information is included in the report on the next startup.

Partially addresses #224

### DIFF
--- a/Source/BugsnagSink.m
+++ b/Source/BugsnagSink.m
@@ -29,7 +29,7 @@
 #import "BugsnagCollections.h"
 #import "BugsnagNotifier.h"
 #import "BugsnagKeys.h"
-#import "BugsnagHandledState.h"
+#import "BSG_KSSystemInfo.h"
 
 // This is private in Bugsnag, but really we want package private so define
 // it here.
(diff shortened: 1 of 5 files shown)


Source/BugsnagSink.m
```diff
@@ -59,9 +60,38 @@ - (void)filterReports:(NSArray *)reports
...
+            NSDictionary *sysInfo = [BSG_KSSystemInfo systemInfo];
+            
+            // reset any existing data as it will be corrupted/nil
+            bugsnagReport.appState = @{};
+            bugsnagReport.deviceState = @{};
+            
+            bugsnagReport.app = @{
+                                  @"bundleVersion": sysInfo[@BSG_KSSystemField_BundleVersion],
+                                  @"id": sysInfo[@BSG_KSSystemField_BundleID],
+                                  @"releaseStage": configuration.releaseStage,
```
kattrali: There is a risk here for any of these values to be nil, particularly release stage if overridden by the user, so inserting values only if not nil would be prudent here as this will otherwise crash if any of these values is not set.

snmaynard:
Should we add tests for this functionality?

fractalwrench:
We plan to add a mazerunner scenario which will cover this test case.

PR #246, opened 2018-03-06
Boilerplate for running maze scenarios

fractalwrench:
Sets up a basic handled exception scenario, with the boilerplate required to run all the other scenarios. This broadly follows the same approach as the Android repo.


features/scripts/send_request.sh
```diff
@@ -0,0 +1,11 @@
...
+
+require 'net/http'
+
+# Sends a request to the mock server running on the port
+# specified by the MOCK_API_PORT environment variable
+http = Net::HTTP.new('localhost', ENV['MOCK_API_PORT'])
+request = Net::HTTP::Post.new('/')
+request['Content-Type'] = 'application/json'
+request.body = '{"dessert":"' + ENV['menu_item'] + '"}'
+http.request(request)
```
kattrali: This file should be deleted. `init` was run on the branch prior to the PR.

features/steps/website_steps.rb
```diff
@@ -0,0 +1,7 @@
+When("I select {string} on the website") do |menu_item|
+steps %Q{
+  When I set environment variable "menu_item" to "#{menu_item}"
+  And I run the script "features/scripts/send_request.sh"
+  And I wait for 1 second
+}
+end
```
kattrali: This file should be deleted.

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/Scenario.swift
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Sets a NOP implementation for the Error Tracking API and the Session Tracking API,
+     * preventing delivery
+     */
+    func disableAllDelivery() {
+        self.disableReportDelivery()
+        self.disableSessionDelivery()
+    }
+
+    func generateException() -> NSException {
```
kattrali: Each individual scenario should specify its own exception rather than deriving it from a helper - this doesn't reflect how exception handling would work in a real app.

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/Scenario.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    required init(config: BugsnagConfiguration) {
+        self.config = config
+    }
+
+
+    /**
+     * Sets a NOP implementation for the Session Tracking API, preventing delivery
+     */
+    func disableSessionDelivery() {
```
kattrali: There may be special cases at some time in the future, but for now, there should not be override mechanisms for changing request delivery. If unexpected requests are happening, then either:

* There is a bug which should be fixed.
* The scenario should be made more specific to account for it
* Properties on the `BugsnagConfiguration` should be changed to exclude them (e.g. disabling session tracking)

Gemfile
```diff
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 
+gem 'cocoapods'
```
kattrali: This shouldn't be needed. `cocoapods` is already a dependency inside the [test fixture project Gemfile](https://github.com/bugsnag/bugsnag-cocoa/blob/add-maze-harness/features/fixtures/ios-swift-cocoapods/Gemfile).

features/unhandled_exception.feature
```diff
@@ -2,17 +2,17 @@ Feature: NSException handling
 
 Scenario: Uncaught NSException is raised
     When I set environment variable "BUGSNAG_API_KEY" to "a35a2a72bd230ac0aa0f52715bbdc6aa"
-    And I set environment variable "EVENT_TYPE" to "NSException"
+    And I set environment variable "EVENT_TYPE" to "HandledExceptionScenario"
```
kattrali: There should be a scenario to handle an NSException being raised. Its not on the test list since it was written already for the branch. The `Scenario:` title should also always match the contents.

features/fixtures/ios-swift-cocoapods/iOSTestApp/AppDelegate.swift
```diff
@@ -13,19 +13,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         return true
     }
 
-    func triggerEvent(name: String, delay: TimeInterval) {
```
kattrali: Allowing the specification of a time delay was intentional to account for cases where exceptions happen after a specific duration, like a crash on launch vs triggering click events on a normal human time scale which lead to a crash.

features/fixtures/ios-swift-cocoapods/iOSTestApp/AppDelegate.swift
```diff
@@ -45,11 +36,30 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    internal func scenarioForEventType(eventType: String, config: BugsnagConfiguration) -> Scenario {
+        let eventTypeForClass = "iOSTestApp." + eventType // prefix with app name
+        let type = NSClassFromString(eventTypeForClass) as! Scenario.Type
+        return type.init(config: config)
+    }
 
-        triggerEvent(name: eventType, delay: delay)
+    func triggerEvent(scenario: Scenario, delay: TimeInterval) {
+        let when = DispatchTime.now() + delay
+        DispatchQueue.main.asyncAfter(deadline: when) {
```
kattrali: Given the scenario structure, I'd move the delay into the `run` method because otherwise Bugsnag is initialized after the delay, which might cause some interesting edge cases.

features/fixtures/ios-swift-cocoapods/iOSTestApp/AppDelegate.swift
fractalwrench: The method signature has changed but this behaviour now remains in place.

features/fixtures/ios-swift-cocoapods/iOSTestApp/AppDelegate.swift
fractalwrench: I've added an `initBugsnag` method which is called prior to `run`.

fractalwrench:
Addressed previous review comments

PR #247, opened 2018-03-07
Handled exception scenarios

fractalwrench:
Adds scenarios for:

- Handled exception
- Handled error
- Override error class/message


features/handled_errors.feature
```diff
@@ -0,0 +1,28 @@
+Feature: Handled Error
+
+Scenario: Handled Error report override
```
kattrali: I needed to look at the source code of the test to know what this test does. A better title would be more explicit, e.g. "Override errorClass and message from a notifyError() callback"

features/handled_errors.feature
```diff
@@ -0,0 +1,28 @@
+Feature: Handled Error
```
kattrali: This is reporting handled errors and exceptions, correct?

features/steps/ios_steps.rb
```diff
@@ -1,3 +1,13 @@
+When("I run {string} with the defaults") do |eventType|
```
kattrali: I'd change this to pass through the simulator value as well, as that is how the tests will likely be split in the near future. One option:

```ruby
When("I run {string} on the {string}") do |event_type, simulator|
```

fractalwrench:
Addressed review comments and updated with latest basic-maze-harness branch.

PR #248, opened 2018-03-13
Add timezone to device payload

fractalwrench:
The timezone was being serialised from the wrong keyPath when mapping the KSCrash report to the Bugsnag crash report.


kattrali:
There should be a test to ensure that timezone is non-null in the report.

fractalwrench:
Added the payload field to the serialisation test.

PR #250, opened 2018-03-14
Update minimal reports serialisation

fractalwrench:
Adds device info to correct dictionary for minimal reports


snmaynard:
No tests?

fractalwrench:
@snmaynard I've added a note to the Cocoa Maze Runner ticket to add this as a scenario.

PR #263, opened 2018-03-28
Add scenario for released object

fractalwrench:
Triggers a crash by sending a message to an object whose memory is freed

### DIFF
--- a/features/crashprobe.feature
+++ b/features/crashprobe.feature
@@ -25,3 +25,4 @@ Scenario: Send a message to an object whose memory has already been freed
     And the exception "message" equals "Attempted to dereference null pointer."
     And the exception "errorClass" equals "SIGSEGV"
     And the "method" of stack frame 0 equals "objc_msgSend"
+    And the "method" of stack frame 1 equals "-[ReleasedObjectScenario run]"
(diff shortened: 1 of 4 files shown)


features/crashprobe.feature
```diff
@@ -14,4 +14,14 @@ Scenario: Abort is reported
...
+Scenario: Send a message to an object whose memory has already been freed
+    When I set environment variable "BUGSNAG_API_KEY" to "a35a2a72bd230ac0aa0f52715bbdc6aa"
+    And I configure the app to run on "iPhone 8"
+    And I crash the app using "ReleasedObjectScenario"
+    And I relaunch the app
+    Then I should receive a request
+    And the request is a valid for the error reporting API
+    And the exception "message" equals "Attempted to dereference null pointer."
+    And the exception "errorClass" equals "SIGSEGV"
+    And the "method" of stack frame 0 equals "objc_msgSend"
```
kattrali: We should check all frames leading back to the scenario runner in the expectations. So this one should include

```gherkin
And the "method" of stack frame 1 equals "-[ReleasedObjectScenario run]"
```

fractalwrench: I've updated the scenario to check for this additional frame.

PR #265, opened 2018-03-28
Add c++ exception scenario

fractalwrench:
Adds a scenario which triggers a C++ exception. This cannot currently be caught by most in-process crash reporters, so the expected behaviour is for 0 requests to be received.

### DIFF
--- a/features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/CxxExceptionScenario.mm
+++ b/features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/CxxExceptionScenario.mm
@@ -41,6 +41,10 @@ const char *kaboom_exception::what() const throw() {
  */
 @implementation CxxExceptionScenario
 
+- (void)run {
+    [self crash];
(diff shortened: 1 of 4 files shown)


features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/CxxExceptionScenario.mm
```diff
@@ -0,0 +1,48 @@
...
+
+const char *kaboom_exception::what() const throw() {
+    return "If this had been a real exception, you would be cursing now.";
+}
+
+/**
+ * Throw an uncaught C++ exception. This is a difficult case for crash reporters to handle,
+ * as it involves the destruction of the data necessary to generate a correct backtrace.
+ */
+@implementation CxxExceptionScenario
```
kattrali: `run` is unimplemented and should call `crash` here.

fractalwrench:
I've updated the scenario to check that a report is generated with an empty stacktrace.

PR #266, opened 2018-03-28
Add swift crash scenario

fractalwrench:
Adds a scenario which triggers a crash in Swift by unwrapping a nil optional.

N.B. this scenario is "imprecise" on CrashProbe due to line number info, which is not currently tested in MazeRunner as this would require symbolication.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,8 @@ matrix:
         - osx_image: xcode9.2
           env: SDK=appletvsimulator11.2 BUILD_TV=1
           script: make test
-        - osx_image: xcode9.2
-          env: SDK=iphonesimulator11.2
+        - osx_image: xcode9.3beta
+          env: SDK=iphonesimulator11.3
           script: bundle exec maze-runner -c
 install: make bootstrap
 
--- a/features/crashprobe.feature
+++ b/features/crashprobe.feature
@@ -26,3 +26,8 @@ Scenario: Swift crash is reported
     And the request is a valid for the error reporting API
     And the exception "message" equals "Unexpectedly found nil while unwrapping an Optional value"
     And the exception "errorClass" equals "Fatal error"
+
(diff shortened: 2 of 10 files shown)


fractalwrench:
I've updated the scenario to check the method names.

fractalwrench:
The Swift scenario appears to fail when comparing the demangled method names when using `xcrun swift-demangle`.

Local:
```
_T0s18_fatalErrorMessages5NeverOs12StaticStringV_A2E4fileSu4lines6UInt32V5flagstFTf4nnddn_n
_T0s18_fatalErrorMessages5NeverOs12StaticStringV_A2E4fileSu4lines6UInt32V5flagstFTf4nnddn_n ---> function signature specialization <Arg[2] = Dead, Arg[3] = Dead> of Swift._fatalErrorMessage(Swift.StaticString, Swift.StaticString, file: Swift.StaticString, line: Swift.UInt, flags: Swift.UInt32) -> Swift.Never
```

CI:
```
_T0s18_fatalErrorMessages5NeverOs12StaticStringV_A2E4fileSu4lines6UInt32V5flagstFTfq4nnddn_n
_T0s18_fatalErrorMessages5NeverOs12StaticStringV_A2E4fileSu4lines6UInt32V5flagstFTfq4nnddn_n ---> function signature specialization <preserving fragile attribute, Arg[2] = Dead, Arg[3] = Dead> of Swift._fatalErrorMessage(Swift.StaticString, Swift.StaticString, file: Swift.StaticString, line: Swift.UInt, flags: Swift.UInt32) -> Swift.Never
```

The difference I can observe is that the CI build method has a [fragile argument specialisation](https://github.com/apple/swift/blob/master/docs/ABI/Mangling.rst). I'm not immediately unsure why this is, but have altered the Travis image to use XCode 9.3-beta so that it's closer to my local version.

fractalwrench:
@kattrali the Swift scenario successfully passes on XCode 9.3-beta, so there must have been a change in Swift demangling since 9.2.

fractalwrench:
@kattrali I've updated the scenario to match the start of the Swift method names, rather than their entirety. Can you confirm whether that's ok?

PR #268, opened 2018-03-28
Add scenario which crashes within the crash handler

fractalwrench:
Triggers a crash within the crash handler exposed by Bugsnag, which should generate a minimal report.

This scenario ensures that useful app/device information is collected regardless on the next launch, and sent in the report.

### DIFF
--- a/features/minimal_crash.feature
+++ b/features/minimal_crash.feature
@@ -1,0 +1,15 @@
+Feature: Minimal Crash report
+    "Minimal" crash reports are generated when an error condition occurs
+    while a crash is already being handled. Minimal reports include basic
+    app and device metadata while skipping the more complex parts of reports,
+    like breadcrumbs and custom post-crash handlers.
+
+Scenario: Crash within the crash handler
+    When I set environment variable "BUGSNAG_API_KEY" to "a35a2a72bd230ac0aa0f52715bbdc6aa"
+    And I configure the app to run on "iPhone8-11.2"
+    And I crash the app using "MinimalCrashReportScenario"
+    And I relaunch the app
+    Then I should receive a request
+    And the request is a valid for the error reporting API
+    And the "Bugsnag-API-Key" header equals "a35a2a72bd230ac0aa0f52715bbdc6aa"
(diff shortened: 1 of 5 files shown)


features/minimal_crash.feature
```diff
@@ -0,0 +1,38 @@
...
+    And the event "app.type" equals "iOS"
+    And the event "app.id" equals "com.bugsnag.iOSTestApp"
+    And the event "app.releaseStage" equals "development"
+
+    And the event "device.locale" is not null
+    And the event "device.osVersion" is not null
+    And the event "device.model" is not null
+    And the event "device.jailbroken" is not null
+    And the event "device.modelNumber" is not null
+    And the event "device.osName" is not null
```
kattrali: The OS name should be a constant value here, so we can use "equals" rather than "is not null"

features/minimal_crash.feature
```diff
@@ -0,0 +1,38 @@
...
+
+    And the event "app.bundleVersion" equals "1"
+    And the event "app.version" equals "1.0"
+    And the event "app.type" equals "iOS"
+    And the event "app.id" equals "com.bugsnag.iOSTestApp"
+    And the event "app.releaseStage" equals "development"
+
+    And the event "device.locale" is not null
+    And the event "device.osVersion" is not null
+    And the event "device.model" is not null
```
kattrali: Since the scenario specifies "iPhone 8", the model here should be the same. When adding other devices, we can tweak as needed.

features/minimal_crash.feature
```diff
@@ -0,0 +1,38 @@
...
+    And the event "severityReason.type" equals "unhandledException"
+
+    And the event "app.bundleVersion" equals "1"
+    And the event "app.version" equals "1.0"
+    And the event "app.type" equals "iOS"
+    And the event "app.id" equals "com.bugsnag.iOSTestApp"
+    And the event "app.releaseStage" equals "development"
+
+    And the event "device.locale" is not null
+    And the event "device.osVersion" is not null
```
kattrali: Similar to `device.model`, this is set as a part of the device selection, so we should specify the expected OS version here (11.2?), and adjust when expanding to other versions.

PR #271, opened 2018-04-05
[WIP] Wait for simulator at start of scenarios

fractalwrench:
Adding a wait for the simulator to finish launching appears to have helped pass scenarios locally. Without this change, the first scenario in a run tends to fail with 0 requests sent.

This value may tweaking depending on the speed of Travis CI.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -12,6 +12,7 @@ Dir.chdir('features/fixtures/ios-swift-cocoapods') do
     ['bundle', 'exec', 'pod', 'install'],
     ['../../scripts/build_ios_app.sh'],
     ['../../scripts/launch_ios_simulators.sh'],
+    ['sleep', '15']
   ])
 end
 


kattrali:
Hmm, still seems to happen for some reason.

fractalwrench:
I'm not too sure what's going on here. I've attempted increasing the wait to 60 seconds, so presumably that's not the issue, but it does seem to fix things locally. Perhaps the launch time and wait between scenarios requires increasing?

It's odd that only the first couple of scenarios seem to fail on CI.

fractalwrench:
A bit of background on my investigations so far, as I don't think the latest commit is going to work on CI.

If a simulator isn't already booted, then no report will be received on the first scenario run by mazerunner, and an extra report is received by the second scenario. This only occurs on the crashprobe-scenarios branch (16af423), and not add-maze-harness (1d535d5).

I've attempted the following fixes, all of which pass locally but fail on CI:

- Rerunning successful build (passes) and failed build (fails) on each commit mentioned above
- Increasing sleep in support/env.rb commands to 3 minutes
- Increasing timeout on CI to 20 seconds for each iOS step
- Adding sleep of 30 in between install_ios_app.sh commands
- Running only the HandledErrorScenario (passes)
- Changing the feature run order so that handled_errors.feature is first (these pass, but the first one in crashprobe.feature still fails)

My gut feeling at this point is that something is going wrong with the relaunch script when it's run for the first time, but I've not investigated this yet.

fractalwrench:
An additional issue is that if mazerunner is immediately interrupted locally, then often an extra report will be sent on the first scenario run, as the simulator data will not have been cleared.

PR #272, opened 2018-04-12
__builtin_trap Mazerunner scenario

fractalwrench:
Adds a mazerunner scenario for __builtin_trap


fractalwrench:
Updated to check for `SIGILL` and that a stack frame is present.

PR #274, opened 2018-04-19
ci: Enable testing iOS/tvOS 11.3

kattrali:
* Upgrade Swift functions to 4.1 mangled names


features/crashprobe.feature
```diff
@@ -2,7 +2,7 @@ Feature: Reporting crash events
 
 Scenario: Calling __builtin_trap()
     When I set environment variable "BUGSNAG_API_KEY" to "a35a2a72bd230ac0aa0f52715bbdc6aa"
-    And I configure the app to run on "iPhone8-11.2"
+    And I configure the app to run on "iPhone8-11.3"
```
fractalwrench: One of our original goals was to setup mazerunner so that it can run on multiple devices, with different models + OS versions. Would now be a good time to extract this to a variable/command line option for iOS?

features/crashprobe.feature
```diff
@@ -147,20 +147,22 @@ Scenario: Send a message to an object whose memory has already been freed
...
     And the request is a valid for the error reporting API
     And the exception "message" equals "Unexpectedly found nil while unwrapping an Optional value"
     And the exception "errorClass" equals "Fatal error"
 
-    And the "method" of stack frame 0 starts with "_T0s18_fatalErrorMessages5NeverOs12StaticStringV_A2E4fileSu4lines6UInt32V5"
-    And the "method" of stack frame 1 starts with "_T010iOSTestApp10SwiftCrashC3run"
+    And the "method" of stack frame 0 equals "_T0s18_fatalErrorMessages5NeverOs12StaticStringV_A2E4fileSu4lines6UInt32V5flagstFTf4nnddn_n"
+    And the "method" of stack frame 1 equals "_T0s18_fatalErrorMessages5NeverOs12StaticStringV_A2E4fileSu4lines6UInt32V5flagstF"
+    And the "method" of stack frame 2 equals "_T010iOSTestApp10SwiftCrashC3runyyF"
+    And the "method" of stack frame 3 equals "_T010iOSTestApp10SwiftCrashC3runyyFTo"
```
fractalwrench: I'm concerned that this step might break in future (even though supposedly the ABI is stable). A similar equality comparison gave different results on my local machine vs Travis, when using XCode 9.3, which is why the current version uses 'starts with'.

features/crashprobe.feature
kattrali: Yes. I marked it as in progress to pull the pieces around a bit. Rather than running all devices at once though, which thus far has been flaky, I'm planning to configure it so you can run it all on a particular device. That way we can also divide up the CI builds for reasonable time lengths.

features/crashprobe.feature
fractalwrench: Expected:
      +++ actual
      @@ -1 +1 @@
      -`"_T0s18_fatalErrorMessages5NeverOs12StaticStringV_A2E4fileSu4lines6UInt32V5flagstFTf4nnddn_n"`
      +"_T0s18_fatalErrorMessages5NeverOs12StaticStringV_A2E4fileSu4lines6UInt32V5flagstFTfq4nnddn_n"

PR #275, opened 2018-04-20
Privileged instruction execution scenario

fractalwrench:
Adds a mazerunner scenario which attempts to execute an unauthorised privileged execution.

Note: this fails on the simulator, whereas the last CrashProbe results showed this passed in 5.0.0. @kattrali could this just be a consequence of running on a simulator?

```
"message": "Attempted to dereference null pointer.",
"errorClass": "SIGSEGV",
```

### DIFF
--- a/features/crashprobe.feature
+++ b/features/crashprobe.feature
@@ -10,7 +10,7 @@ Scenario: Executing privileged instruction
     And the "Bugsnag-API-Key" header equals "a35a2a72bd230ac0aa0f52715bbdc6aa"
     And the payload field "notifier.name" equals "iOS Bugsnag Notifier"
     And the payload field "events" is an array with 1 element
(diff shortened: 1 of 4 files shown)


features/crashprobe.feature
```diff
@@ -10,7 +10,7 @@ Scenario: Executing privileged instruction
     And the "Bugsnag-API-Key" header equals "a35a2a72bd230ac0aa0f52715bbdc6aa"
     And the payload field "notifier.name" equals "iOS Bugsnag Notifier"
     And the payload field "events" is an array with 1 element
-    And the exception "errorClass" equals "SIGILL"
+    And the exception "errorClass" equals "SIGSEGV"
```
fractalwrench: According to the crashprobe scenario, SIGILL is expected here: http://www.crashprobe.com/ios/08/

PR #279, opened 2018-05-15
fix: Adds fallback dictionary when serialising device info.

fractalwrench:
## Goal

Previously the device tab was serialised using the dictionary supplied from the report, which could be null if the keypath did not exist. This could result in the device tab being completely null. This fix uses an empty dictionary as a fallback so that at least some information (freeDisk, locale, simulator) will always be returned.

### Changed
Initialise empty dictionary and defensively add items from report to it.

## Tests

Added a failing unit test for the encountered scenario then passed it.

## Review

  - [x] Final review

### DIFF
--- a/Source/BugsnagKSCrashSysInfoParser.m
+++ b/Source/BugsnagKSCrashSysInfoParser.m
@@ -16,9 +16,10 @@
 #define PLATFORM_WORD_SIZE sizeof(void*)*8
 
 NSDictionary *BSGParseDevice(NSDictionary *report) {
-    NSMutableDictionary *device =
-    [[report valueForKeyPath:@"user.state.deviceState"] mutableCopy];
-    
+    NSMutableDictionary *device = [NSMutableDictionary new];
+    NSDictionary *state = [[report valueForKeyPath:@"user.state.deviceState"] mutableCopy];
+    [device addEntriesFromDictionary:state];
+
     [device addEntriesFromDictionary:BSGParseDeviceState(report[@"system"])];
     
     BSGDictSetSafeObject(device, [[NSLocale currentLocale] localeIdentifier],
--- a/Source/BugsnagKSCrashSysInfoParser.m
+++ b/Source/BugsnagKSCrashSysInfoParser.m
@@ -17,7 +17,7 @@
 
 NSDictionary *BSGParseDevice(NSDictionary *report) {
     NSMutableDictionary *device = [NSMutableDictionary new];
-    NSDictionary *state = [[report valueForKeyPath:@"user.state.deviceState"] mutableCopy];
+    NSDictionary *state = [report valueForKeyPath:@"user.state.deviceState"];
     [device addEntriesFromDictionary:state];
 
     [device addEntriesFromDictionary:BSGParseDeviceState(report[@"system"])];
--- a/iOS/BugsnagTests/BugsnagKSCrashSysInfoParserTest.m
+++ b/iOS/BugsnagTests/BugsnagKSCrashSysInfoParserTest.m
@@ -1,0 +1,27 @@
+//
+//  BugsnagKSCrashSysInfoParserTestTest.m
+//  Tests
+//
+//  Created by Jamie Lynch on 15/05/2018.
+//  Copyright © 2018 Bugsnag. All rights reserved.
+//
+
+#import <XCTest/XCTest.h>
+
+#import "BugsnagKSCrashSysInfoParser.h"
+
+@interface BugsnagKSCrashSysInfoParserTest : XCTestCase
+@end
+
+@implementation BugsnagKSCrashSysInfoParserTest
+
+- (void)testEmptyDictSerialisation {
+    // ensures that an empty dictionary parameter returns a fallback dictionary populated with at least some information
+    NSDictionary *device = BSGParseDevice(@{});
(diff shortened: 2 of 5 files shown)


kastiglione:
👍

Source/BugsnagKSCrashSysInfoParser.m
```diff
@@ -16,9 +16,10 @@
 #define PLATFORM_WORD_SIZE sizeof(void*)*8
 
 NSDictionary *BSGParseDevice(NSDictionary *report) {
-    NSMutableDictionary *device =
-    [[report valueForKeyPath:@"user.state.deviceState"] mutableCopy];
-    
+    NSMutableDictionary *device = [NSMutableDictionary new];
+    NSDictionary *state = [[report valueForKeyPath:@"user.state.deviceState"] mutableCopy];
```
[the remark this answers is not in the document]
kastiglione: For consideration: Since `valueForKey:` can throw exceptions for incorrect keys, and since keypaths can't be checked by the compiler, is it worth considering wrapping these in a `@try`/`@catch`? It's certainly very unlikely these will throw, but not impossible.

iOS/BugsnagTests/BugsnagKSCrashSysInfoParserTest.m
```diff
@@ -0,0 +1,27 @@
...
+- (void)testEmptyDictSerialisation {
+    // ensures that an empty dictionary parameter returns a fallback dictionary populated with at least some information
+    NSDictionary *device = BSGParseDevice(@{});
+    XCTAssertNotNil(device);
+    XCTAssertNotNil(device[@"locale"]);
+    XCTAssertNotNil(device[@"freeDisk"]);
+    XCTAssertNotNil(device[@"simulator"]);
+}
+
+@end
```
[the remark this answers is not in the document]
fractalwrench: I've tested this out and a dictionary is returned with the same contents as if an empty dictionary had been passed in. Adding this as a test would currently fail the suite with an error, presumably due to the fact we've enabled strict build settings and this method parameter is enabled as NonNull. Do you feel we need to change anything so that this can be unit tested?

Source/BugsnagKSCrashSysInfoParser.m
```diff
@@ -16,9 +16,10 @@
 #define PLATFORM_WORD_SIZE sizeof(void*)*8
 
 NSDictionary *BSGParseDevice(NSDictionary *report) {
-    NSMutableDictionary *device =
-    [[report valueForKeyPath:@"user.state.deviceState"] mutableCopy];
-    
+    NSMutableDictionary *device = [NSMutableDictionary new];
```
martin308: Would it be better to do something like this? I'm not sure how `initWithDictionary` handles nil

```
NSDictionary *state = [report valueForKeyPath:@"user.state.deviceState"];
NSMutableDictionary* device = [[NSMutableDictionary alloc] initWithDictionary: state];
```

iOS/BugsnagTests/BugsnagKSCrashSysInfoParserTest.m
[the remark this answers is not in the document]
martin308: Is it possible to loosen the restrictions on just the test files? It seems like it might be useful to have a test like that

Source/BugsnagKSCrashSysInfoParser.m
fractalwrench: `initWithDictionary:` requires a Nonnull parameter, whereas `addEntriesFromDictionary:` handles a nil param.

fractalwrench:
@martin308 I've added a test for a nil dictionary by using `#pragma` to loosen the restrictions.

iOS/BugsnagTests/BugsnagKSCrashSysInfoParserTest.m
```diff
@@ -0,0 +1,36 @@
...
+    XCTAssertNotNil(device[@"locale"]);
+    XCTAssertNotNil(device[@"freeDisk"]);
+    XCTAssertNotNil(device[@"simulator"]);
+}
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wnonnull"
+
+- (void)testNilDictSerialisation {
+    XCTAssertNotNil(BSGParseDevice(nil));
```
kastiglione: This could return an empty dictionary and still pass. I think this should check the contents too.

fractalwrench:
Thanks for the review comments - I've addressed the pragma + validation comments.

iOS/BugsnagTests/BugsnagKSCrashSysInfoParserTest.m
```diff
@@ -0,0 +1,41 @@
...
+
+- (void)validateDeviceDict:(NSDictionary *)device {
+    XCTAssertNotNil(device);
+    XCTAssertNotNil(device[@"locale"]);
+    XCTAssertNotNil(device[@"freeDisk"]);
+    XCTAssertNotNil(device[@"simulator"]);
+}
+
+
+@end
```
martin308: Do we need a test for a valid dictionary that contains the expected key?

fractalwrench: This is tested elsewhere I believe

PR #280, opened 2018-05-18
Enforce requiring API key to initialise notifier

fractalwrench:
## Goal

Avoid instantiating the Notifier in an invalid state (without an API key).

### Changed
- Nullability annotations on API
- Gate all required methods in `Bugsnag.m` with a check to see if the notifier has been initialised, and NOP if not.
- Avoid initialising notifier if the supplied API key is nil.
- Attempting to set a non-nil API key as nil is ignored

## Tests

Added a unit test for API key setter override.

## Discussion

Much of the discussion for this changeset took place here: https://github.com/bugsnag/bugsnag-react-native/pull/237

## Review

  - [x] Final review

### DIFF
--- a/Source/Bugsnag.m
+++ b/Source/Bugsnag.m
@@ -177,7 +177,7 @@ static BugsnagNotifier *bsg_g_bugsnag_notifier = NULL;
 }
 
 + (BOOL)bugsnagStarted {
-    if (self.notifier == nil) {
+    if (!self.notifier.started) {
         bsg_log_err(@"Ensure you have started Bugsnag with startWithApiKey: "
                     @"before calling any other Bugsnag functions.");
 
--- a/Source/Bugsnag.m
+++ b/Source/Bugsnag.m
@@ -51,9 +51,13 @@ static BugsnagNotifier *bsg_g_bugsnag_notifier = NULL;
 
 + (void)startBugsnagWithConfiguration:(BugsnagConfiguration *)configuration {
     @synchronized(self) {
-        bsg_g_bugsnag_notifier =
-        [[BugsnagNotifier alloc] initWithConfiguration:configuration];
-        [bsg_g_bugsnag_notifier start];
+        if (configuration != nil && configuration.apiKey != nil) {
+            bsg_g_bugsnag_notifier =
+                    [[BugsnagNotifier alloc] initWithConfiguration:configuration];
+            [bsg_g_bugsnag_notifier start];
+        } else {
+            bsg_log_err(@"Bugsnag not initialized - a valid API key must be supplied.");
+        }
     }
 }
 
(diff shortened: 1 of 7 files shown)


Source/Bugsnag.m
```diff
@@ -51,9 +51,13 @@ + (void)startBugsnagWithApiKey:(NSString *)apiKey {
 
 + (void)startBugsnagWithConfiguration:(BugsnagConfiguration *)configuration {
     @synchronized(self) {
-        bsg_g_bugsnag_notifier =
-        [[BugsnagNotifier alloc] initWithConfiguration:configuration];
-        [bsg_g_bugsnag_notifier start];
+        if (configuration != nil && configuration.apiKey != nil) {
```
snmaynard: Might be nicer to have a configuration based validation method. `[configuration isValid]` for example, rather than the Bugsnag class having to know what that means.

snmaynard:
By tests I mean, that `configuration isValid` works as expected and that you can't start bugsnag if that function returns false.

fractalwrench:
Added a test to `BugsnagConfiguration` to encapsulate the API key check.

Source/BugsnagConfiguration.m
```diff
@@ -231,4 +246,9 @@ - (NSDictionary *)sessionApiHeaders {
              kHeaderApiSentAt: [BSG_RFC3339DateTool stringFromDate:[NSDate new]]
              };
 }
+
+- (BOOL)hasValidApiKey {
+    return _apiKey != nil && ![@"" isEqualToString:_apiKey];
```
martin308: could we just use `return [_apiKey length] == 0;`?

Source/BugsnagConfiguration.m
```diff
@@ -199,6 +200,20 @@ - (void)setAppVersion:(NSString *)newVersion {
     }
 }
 
+@synthesize apiKey = _apiKey;
+
+- (NSString *)apiKey {
+    return _apiKey;
+}
```
martin308: could we remove this as it is just the default generated by `@synthesize`?

Source/BugsnagConfiguration.h
```diff
@@ -69,7 +69,7 @@ typedef NSDictionary *_Nullable (^BugsnagBeforeNotifyHook)(
 /**
  *  The API key of a Bugsnag project
  */
-@property(readwrite, retain, nullable) NSString *apiKey;
+@property(readwrite, retain, nonnull) NSString *apiKey;
```
martin308: is this a breaking change? Do we need to increment the major version number?

Source/BugsnagConfiguration.m
kattrali: I think it needs to be specified since `setApiKey` is overridden.

Source/BugsnagConfiguration.m
```diff
@@ -199,6 +200,20 @@ - (void)setAppVersion:(NSString *)newVersion {
...
+
+- (void)setApiKey:(NSString *)apiKey {
+    if (apiKey != nil) {
```
kattrali: `if (apiKey.length > 0)` would be more comprehensive here and catch both the nil and the empty string case.

Source/BugsnagConfiguration.m
kattrali: Flipped the boolean but yeah, `[_apiKey length] > 0` is the ticket here.

Tests/BugsnagConfigurationTests.m
```diff
@@ -99,4 +99,31 @@ - (void)testUser {
...
+
+- (void)testHasValidApiKey {
+    BugsnagConfiguration *config = nil;
+    XCTAssertFalse([config hasValidApiKey]);
+
+    config = [BugsnagConfiguration new];
+    XCTAssertFalse([config hasValidApiKey]);
+
+    config.apiKey = @"5adf89e0aaa";
+    XCTAssertTrue([config hasValidApiKey]);
```
kattrali: `config.apiKey = ""` should also be false.

Source/BugsnagConfiguration.h
kattrali: This is a breaking change in Swift at least, as anyone fetching the value would need to update their code with the appropriate number of `?` and `!`. Its also not enforced, as `BugsnagConfiguration().apiKey` could be nil. 

We should not make this change just yet and instead design/think more about the interface, such as automatic API key detection from `Info.plist` or having a recommended constructor which takes an non-null API key as an argument. This change is also not required for this fix.

fractalwrench:
@martin308 good catch with `bugsnagStarted` - I've altered this so that it checks a boolean property which is set to true when the notifier has been started, rather than checking for nil.

I have addressed all other inline feedback.

Source/Bugsnag.m
```diff
@@ -161,7 +177,7 @@ + (void)clearTabWithName:(NSString *)tabName {
 }
 
 + (BOOL)bugsnagStarted {
-    if (self.notifier == nil) {
+    if (!self.notifier.started) {
```
martin308: Does this handle the notifier being `nil`?

fractalwrench: Yes, as nil will be coerced to false.

PR #281, opened 2018-05-22
Handled/unhandled error state serialisation fix

fractalwrench:
## Goal
Handled error data serialisation is synchronised through the use of `NSLock`. However, this does not apply to unhandled errors, which suspend all threads and can therefore unintentionally serialise data belonging to a handled error. 

This fix alters serialisation of error data by ignoring some unnecessary fields for unhandled errors, and calculating `severity` and other items using existing data.

### Added
- New internal typedef for a report write callback, which passes information on the whether a report is handled/unhandled.

### Changed
Ignore serialised values on unhandled errors for:

- depth
- severity
- handledStateJSON
- metaDataJSON
- userOverridesJSON

Calculate severity/depth using existing information for unhandled errors in `BugsnagCrashReport`

## Tests

New unit tests for serialisation of report depth + severity, mazerunner run on CI, manual testing to ensure that one handled/unhandled error were received.

### Outstanding Questions

- We may want to make `BSGReportCallback` public and deprecated `onCrashHandler`. I'm not 100% convinced that the information available in context makes this worth doing as it can mostly be obtained by other means, so I've left as-is for now.
- I'm fairly sure that depth, severity, and handledStateJSON don't need to be serialised, but it would be good to confirm that this doesn't cause some unintended side effect or misses an edge-case.

## Review

  - [ ] Final review

### DIFF
--- a/Source/BugsnagNotifier.m
+++ b/Source/BugsnagNotifier.m
@@ -103,12 +103,12 @@ void BSSerializeDataCrashHandler(const BSG_KSCrashReportWriter *writer, int type
     if (bsg_g_bugsnag_data.stateJSON) {
         writer->addJSONElement(writer, "state", bsg_g_bugsnag_data.stateJSON);
     }
+    if (bsg_g_bugsnag_data.metaDataJSON) {
+        writer->addJSONElement(writer, "metaData", bsg_g_bugsnag_data.metaDataJSON);
+    }
 
     // write additional user-supplied metadata
     if (userReported) {
-        if (bsg_g_bugsnag_data.metaDataJSON) {
-            writer->addJSONElement(writer, "metaData", bsg_g_bugsnag_data.metaDataJSON);
-        }
         if (bsg_g_bugsnag_data.handledState) {
             writer->addJSONElement(writer, "handledState", bsg_g_bugsnag_data.handledState);
         }
--- a/Source/BugsnagNotifier.m
+++ b/Source/BugsnagNotifier.m
@@ -34,6 +34,7 @@
 #import "BugsnagSessionTracker.h"
 #import "BugsnagSessionTrackingApiClient.h"
 #import "BSG_RFC3339DateTool.h"
+#import "BSG_KSCrashType.h"
 
 #if TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
 #import <UIKit/UIKit.h>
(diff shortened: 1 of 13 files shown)


Source/BugsnagNotifier.m
```diff
@@ -81,40 +82,41 @@
...
         writer->addJSONElement(writer, "state", bsg_g_bugsnag_data.stateJSON);
     }
-    if (bsg_g_bugsnag_data.userOverridesJSON) {
-        writer->addJSONElement(writer, "overrides",
-                               bsg_g_bugsnag_data.userOverridesJSON);
+
+    // write additional user-supplied metadata
+    if (userReported) {
+        if (bsg_g_bugsnag_data.metaDataJSON) {
+            writer->addJSONElement(writer, "metaData", bsg_g_bugsnag_data.metaDataJSON);
```
kattrali: This value should be attached to all reports as it contains the values from [`+[Bugsnag addAttribute:withValue:toTabWithName:]`](https://docs.bugsnag.com/platforms/ios/customizing-error-reports/#adding-custom-diagnostics). We should add a test for this case.

fractalwrench:
I've updated this so that the metadata JSON is always included in reports. What would be the best way to test this? Mazerunner for Cocoa is currently out of action on my local machine.

Tests/BugsnagCrashReportTests.m
```diff
@@ -449,4 +449,35 @@ - (void)testEnhancedErrorMessageIgnoresUnknownAssertionTypes {
...
+    XCTAssertEqual(report.depth, 2);
+}
+
+- (void)testReportSeverity {
+    // unhandled reports should calculate their own severity
+    NSDictionary *dict = @{@"user.state.crash.severity": @"info"};
+    BugsnagCrashReport *report = [[BugsnagCrashReport alloc] initWithKSReport:dict];
+    XCTAssertEqual(report.severity, BSGSeverityError);
+
+    // handled reports should use the serialised depth
```
snmaynard: This comment needs updating. I would also split these tests into 2 if it was me.

snmaynard:
You could add a unit test for the metadata thing, similar to the severity tests you have. Whats wrong with mazerunner though?

fractalwrench:
Added a unit test for metadata JSON serialisation as requested, and split some of the newly added tests into two.

Note that I can't seem to dismiss @kattrali's old review, so am unable to merge.

PR #283, opened 2018-06-06
set _started before reporting

ghazel:
I'm using `[Bugsnag startBugsnagWithConfiguration:config];`

When starting Bugsnag, several sessions are not sent to the server, and instead this is logged many times:

```
Ensure you have started Bugsnag with startWithApiKey: before calling any other Bugsnag functions.
```

This seems to be due to `_started` not being set yet when `willEnterForeground` is called.

Moving the `_started` value change fixes the issue and correctly reports existing session data.

### DIFF
--- a/Source/BugsnagNotifier.m
+++ b/Source/BugsnagNotifier.m
@@ -356,9 +356,10 @@ NSString *const kAppWillTerminate = @"App Will Terminate";
                  object:nil];
 #endif
 
+    _started = YES;
+
     // notification not received in time on initial startup, so trigger manually
     [self willEnterForeground:self];
-    _started = YES;
 }
 
 - (void)watchLifecycleEvents:(NSNotificationCenter *)center {


fractalwrench:
Thanks for the detailed report and PR @ghazel - this changeset should address the immediate issue. We're planning on adding some black box tests to improve our coverage of session tracking functionality before making changes in this area, so we may end up adding them onto this PR.

samsiner:
Was this ever fixed?  I'm still getting this error message in console.  I have the latest version of Bugsnag (6.25.2) installed using SPM.

image

johnkiely1:
Hi @samsiner, would you mind writing in to us support@bugsnag.com with some more details about your project setup. In order to help us investigate, could also include some screenshots or code snippets of your Bugsnag configuration code along with which Bugsnag project this relates to that would be great. Thanks.

PR #284, opened 2018-06-06
handle nil contentValue

ghazel:
`contentValue` (`data[@"value"]`) can be `nil`. I have seen entries in `notable_addresses` which have only `address` and `type`. When this occurs, the code crashed trying to handle nil `contentValue`, and ignoring it worked.


fractalwrench:
Thanks for tracking this down @ghazel - I'm going to write a few unit tests for this to make sure that we're not missing any other cases here.

fractalwrench:
This is addressed in #289, so I'm closing the PR for now - thanks again for letting us know.

Issue #282, opened 2018-06-05
fatal error: no 'assign', 'retain', or 'copy' attribute is specified - 'assign' is assumed


ghazel:
### Expected behavior
Compiles cleanly when included.

### Observed behavior
```objc
BugsnagConfiguration.h:103:1: fatal error: no 'assign', 'retain', or 'copy' attribute is specified - 'assign' is assumed
BugsnagConfiguration.h:194:1: fatal error: no 'assign', 'retain', or 'copy' attribute is specified - 'assign' is assumed
BugsnagConfiguration.h:195:1: fatal error: no 'assign', 'retain', or 'copy' attribute is specified - 'assign' is assumed
```

### Steps to reproduce
Include in a project with `-Werror`

### Version
`89cbcb4118306e7dd0767df642d6d3c942e766b3`

### Additional information

Three lines in `BugsnagConfiguration.h ` have `@property(nullable)` when they *probably* should have `@property(retain, nullable)`.

ghazel:
https://github.com/bugsnag/bugsnag-cocoa/blob/89cbcb4118306e7dd0767df642d6d3c942e766b3/Source/BugsnagConfiguration.h#L103

https://github.com/bugsnag/bugsnag-cocoa/blob/89cbcb4118306e7dd0767df642d6d3c942e766b3/Source/BugsnagConfiguration.h#L194

https://github.com/bugsnag/bugsnag-cocoa/blob/89cbcb4118306e7dd0767df642d6d3c942e766b3/Source/BugsnagConfiguration.h#L195

PR #285, opened 2018-06-06
retain

ghazel:
#282

### DIFF
--- a/Source/BugsnagConfiguration.h
+++ b/Source/BugsnagConfiguration.h
@@ -100,7 +100,7 @@ typedef NSDictionary *_Nullable (^BugsnagBeforeNotifyHook)(
 /**
  * The current user
  */
-@property(nullable) BugsnagUser *currentUser;
+@property(retain, nullable) BugsnagUser *currentUser;
 
 /**
  *  Additional information about the state of the app or environment at the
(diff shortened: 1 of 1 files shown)


kattrali:
Thanks, @ghazel! Out of curiosity, what version of Xcode are you running?

ghazel:
Xcode Version 9.4 (9F1027a)

ghazel:
You should be able to trigger it with `-Wall -Werror` on just about any version, though.

kattrali:
Right - we have `-Werror` on already though, I'm not sure why that didn't trigger it.

ghazel:
Ah, it's triggered by `-Wextra` (which includes this specific warning, `-Wobjc-property-no-attribute`). The reason I see it and you don't is that my iOS app defines `-Wextra`, and by including `Bugsnag.h` in my project (which includes `BugsnagConfiguration.h`), the compiler uses my warning flags on your code.

I tried defining `-Wextra` in the Bugsnag project, and that raises a whole bunch of (easily fixed) errors. So that's up to y'all.

Thankfully, this PR seems to cover all the public-facing cases.

NSExceptional:
Unrelated to this project, but I can't seem to trigger this warning at all in Xcode itself, even with `-Weverything`

PR #286, opened 2018-06-06
On by default session tracking

fractalwrench:
## Goal

Enables automatic tracking of sessions by default, and adds a new API for setting the notify/session endpoints.

## Design

This should prevent the unintentional scenario where a user sets the notify endpoint with a sessions endpoint.

### Added

`setEndpointsForNotify:sessions:`

### Changed

- Updated docs for `sessionURL` and `notifyURL` to indicate they should be only used to read the value, not write.
- Auto capture flag is now true by default

### Outstanding Questions

- What should our approach be to deprecating the `notifyURL` and `sessionURL` properties? We may want to keep these properties to allow users to retrieve the value they have set. One option would be to deprecate them entirely, another would be to change to readonly rather than readwrite.

- What is our approach for writing mazerunner scenarios? We're missing scenarios for the session tracking feature entirely, so presumably we need to write these in addition to the new test cases in the design spec. Cocoa mazerunner scenarios also don't currently work on my machine, and I'm concerned that waiting 20-30 minutes for CI to run the scenarios is a very long feedback loop.

## Review

- [ ] Reviewed the test cases added for completeness and possible points for discussion
- [ ] Check the scope of the changeset - is everything in the diff required for the pull request?
  - [ ] Final review


fractalwrench:
I've now added some mazerunner scenarios which should add coverage for session tracking. I believe I've found a couple of issues, the first being that the session request sends `user.emailAddress` rather than `user.email`. I've corrected this in e12ff70.

The second issue is that the `AutoSessionScenario` scenario appears to receive 2 requests, rather than 1. I'm not 100% sure why this occurs, but it's possible that (willEnterForeground)[https://github.com/bugsnag/bugsnag-cocoa/blob/master/Source/BugsnagNotifier.m#L360] is called multiple times, resulting in a duplicate number of sessions. I've created a separate ticket to investigate this further.

Source/BugsnagConfiguration.h
```diff
@@ -142,16 +138,46 @@ BugsnagBreadcrumbs *breadcrumbs;
...
+ * Determines whether app sessions should be tracked automatically. By default this value is true.
  */
 @property BOOL shouldAutoCaptureSessions;
 
 /**
- * Set the endpoint to which tracked sessions reports are sent. This defaults to https://sessions.bugsnag.com,
- * but should be overridden if you are using Bugsnag On-premise, to point to your own Bugsnag endpoint.
+ * Retrieves the endpoint used to notify Bugsnag of errors
+ *
+ * NOTE: it is strongly recommended that you set this value via setEndpointsForNotify:sessions: instead.
```
martin308: is it worth using something like [this](https://stackoverflow.com/a/3908766/549238) to make this as deprecated?

Source/BugsnagConfiguration.h
```diff
@@ -142,16 +138,46 @@ BugsnagBreadcrumbs *breadcrumbs;
...
+ *
+ * @see setEndpointsForNotify:sessions:
+ */
+@property(readwrite, retain, nullable) NSURL *notifyURL;
+
+/**
+ * Retrieves the endpoint used to send tracked sessions to Bugsnag
+ *
+ * NOTE: it is strongly recommended that you set this value via setEndpointsForNotify:sessions: instead.
```
martin308: is it worth using something like [this](https://stackoverflow.com/a/3908766/549238) to make this as deprecated?

Source/BugsnagUser.m
```diff
@@ -14,7 +14,7 @@ @implementation BugsnagUser
 - (instancetype)initWithDictionary:(NSDictionary *)dict {
     if (self = [super init]) {
         _userId = dict[@"id"];
-        _emailAddress = dict[@"emailAddress"];
+        _emailAddress = dict[@"email"];
```
martin308: are the changes in this file related to this PR?

Tests/BugsnagConfigurationTests.m
```diff
@@ -87,6 +88,61 @@ - (void)testSessionEndpoints {
...
+    XCTAssertEqualObjects([NSURL URLWithString:@"http://notify.example.com"], config.notifyURL);
+    XCTAssertEqualObjects([NSURL URLWithString:@"http://sessions.example.com"], config.sessionURL);
+}
+
+- (void)testSetEmptyNotifyEndpoint {
+    @try {
+        BugsnagConfiguration *config = [BugsnagConfiguration new];
+        [config setEndpointsForNotify:@"" sessions:@"http://sessions.example.com"];
+        XCTFail();
+    } @catch(NSException * e) {
```
martin308: is it better to use something like `XCTAssertThrowsSpecific` here?

Tests/BugsnagConfigurationTests.m
```diff
@@ -87,6 +88,61 @@ - (void)testSessionEndpoints {
...
+    }
+}
+
+- (void)testSetMalformedNotifyEndpoint {
+    @try {
+        BugsnagConfiguration *config = [BugsnagConfiguration new];
+        [config setEndpointsForNotify:@"http://" sessions:@"http://sessions.example.com"];
+        XCTFail();
+    } @catch(NSException * e) {
```
martin308: is it better to use something like `XCTAssertThrowsSpecific` here?

Tests/BugsnagUserTest.m
```diff
@@ -19,7 +19,7 @@ - (void)testDictDeserialisation {
 
     NSDictionary *dict = @{
             @"id": @"test",
-            @"emailAddress": @"fake@example.com",
+            @"email": @"fake@example.com",
```
martin308: are the changes in this file related to this PR?

Source/BugsnagConfiguration.h
fractalwrench: That's one of the outstanding questions - what are your thoughts on this? If we add `setEndpointsForNotify:sessions:`, then developers are able to set the value, but won't be able to read it without `notifyURL` and `sessionsURL`.

The way I see it there are a couple of options:

- Leave `notifyURL` and `sessionsURL` as is and document they shouldn't be set
- Deprecate then remove the properties
- Make the properties readonly (breaking change)

I've currently leant towards the first option.

Source/BugsnagUser.m
fractalwrench: This addresses an issue in the session request serialisation which was found when adding mazerunner scenarios. Currently the session request sends `user.emailAddress` rather than `user.email`, which is corrected in e12ff70.

I can split this off into a separate PR if you think it makes more sense.

martin308: that makes sense, I think it would be good to have it in a separate PR to track it more easily

Source/BugsnagConfiguration.m
```diff
@@ -249,6 +251,22 @@ - (NSDictionary *)sessionApiHeaders {
              };
 }
 
+- (void)setEndpointsForNotify:(NSString *_Nonnull)notify sessions:(NSString *_Nonnull)sessions {
```
martin308: should we support providing a null sessions value? It seems on the other notifiers that is something that we support

Tests/BugsnagConfigurationTests.m
```diff
@@ -87,6 +88,55 @@ - (void)testSessionEndpoints {
...
+    BugsnagConfiguration *config = [BugsnagConfiguration new];
+    [config setEndpointsForNotify:@"http://notify.example.com" sessions:@"http://sessions.example.com"];
+    XCTAssertEqualObjects([NSURL URLWithString:@"http://notify.example.com"], config.notifyURL);
+    XCTAssertEqualObjects([NSURL URLWithString:@"http://sessions.example.com"], config.sessionURL);
+}
+
+- (void)testSetEmptyNotifyEndpoint {
+    BugsnagConfiguration *config = [BugsnagConfiguration new];
+    XCTAssertThrowsSpecificNamed([config setEndpointsForNotify:@"" sessions:@"http://sessions.example.com"],
+            NSException, NSInternalInconsistencyException);
```
martin308: 🎉

Source/BugsnagConfiguration.h
martin308: I see the dilemma. Is it possible to split the property up and only deprecate the write? I only jumped on this as I saw in the android notifier that we were deprecating bits like this as part of similar changes. If it doesn't make sense for `bugsnag-cocoa` then that is cool

Source/BugsnagConfiguration.m
fractalwrench: No - these annotations are part of the original design spec.

fractalwrench:
I believe the outstanding issues with this PR are the question of what to do with the `notifyURL` and `sessionURL` properties, which it'd be good to get @kattrali's opinion on. 

I found a few issues with the existing Cocoa session tracking implementation, which I've addressed in this PR and need review:

1. `BugsnagUser` incorrectly serialised the email address as `user.emailAddress`, whereas the Session Tracking API expects `user.email`. The fix has been to change the key in the relevant dictionary.

2. `[BugsnagNotifier start`] sets a boolean property `self.started` to indicate when the notifier has been fully initialised. Within this method, `[self willEnterForeground:]` is called to track an initial session, which resulted in a NOP. The fix here was to set the flag to true before starting the first session.

3. There appears to be an extra session recorded within `[[BugsnagNotifier alloc] init]`, which has been removed so that only 1 session is captured on initialisation by default.

4. There is a potential for `[BugsnagSessionTracker send]` to send the same session multiple times. Although the method uses `@synchronize`, the underlying `BugsnagApiClient` uses `NSURLSession`, which is asynchronous. The proposed fix is to use a semaphore to control access, and to signal within the completion block when the request has finished. I'm not convinced that this is a likely scenario for a real-world application as I've only seen it in mazerunner scenarios which have their own hooks to flush sessions (reducing the time spent waiting) - however, it would be nice to address it.

Source/BugsnagSessionTracker.m
```diff
@@ -87,34 +92,39 @@ - (void)incrementHandledError {
...
                             }
-                        }];
-        }
+                        } else {
+                            NSLog(@"Failed to send sessions to Bugsnag: %@", error);
+                        }
+                        dispatch_semaphore_signal(requestSemaphore);
+                    }];
     }
+    dispatch_semaphore_wait(requestSemaphore, DISPATCH_TIME_FOREVER);
```
kattrali: What happens here if `payload.sessions.count == 0` (and the previous block is never triggered)? Might need to be accounted for to prevent a deadlock.

Source/BugsnagConfiguration.h
kattrali: Might be worth trying:

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/AutoSessionScenario.m
```diff
@@ -0,0 +1,20 @@
...
+// Copyright (c) 2018 Bugsnag. All rights reserved.
+//
+
+#import "AutoSessionScenario.h"
+
+
+@implementation AutoSessionScenario
+
+- (void)startBugsnag {
+    self.config.shouldAutoCaptureSessions = YES;
```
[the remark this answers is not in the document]
fractalwrench: Ordinarily this would be true, but the `config` object passed to the scenario has `shouldAutoCaptureSessions` set to false. This avoids needing to update all the other scenarios to anticipate an additional request.

Source/BugsnagSessionTracker.m
fractalwrench: Addressed by sending a signal to the semaphore immediately in this case.

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/AutoSessionScenario.m
```diff
@@ -0,0 +1,20 @@
...
+    [super startBugsnag];
+}
+
+- (void)run {
+    [self flushAllSessions];
```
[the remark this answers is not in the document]
fractalwrench: My main concern with this approach is that if we apply it to all the session scenarios, it's going to lead to very long CI build times, as it can take 60s for this to occur. I've updated the scenarios to send sessions naturally for now, but think we'll have to revisit this approach if it gets to the point where we have >5 session scenarios.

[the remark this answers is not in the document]
bengourley: Can you shorten the interval in tests? It sounds similar to something that I did in the tests [bugsnag-node](https://github.com/bugsnag/bugsnag-node/blob/1b6ba6dbd3ba18b4ee0f4a284318306a44bc5e66/lib/sessions/tracker.js#L11).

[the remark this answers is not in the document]
bengourley: Also did some casual time travelling https://github.com/bugsnag/bugsnag-node/blob/7f4ee1c1219216b9f19ab7ce5ea75c33e555388d/test/sessions/tracker.js#L12

fractalwrench:
Addressed the inline feedback and added a changelog entry. I've decided to make `notifyURL` and `sessionURL` readonly, as the suggested approach in the comments wasn't successful.

kattrali:
Adding `- (void)setNotifyURL` etc wasn’t successful?

fractalwrench:
@kattrali nope - not in the example Swift project. I attempted adding:

```
- (void)setNotifyURL:(NSURL *)url {
    _notifyURL = url;
}
```

Then ran `pod install`, and attempted to compile `config.notifyURL = URL(string: "")`

kattrali:
What did the message say? Could be something that needs to be aliased for swift. If resolving that doesn’t work, it would make more sense to skip setting `readonly` to avoid the breaking change and keep the setters as deprecated.

fractalwrench:
@kattrali the exact error message is: `Error:(30, 42) cannot assign to property: 'notifyURL' is a get-only property`

For the following declaration: `- (void)setNotifyURL:(NSURL *)url NS_SWIFT_NAME(notifyURL);`

It seems to compile ok for Objective-C, although there is a lint warning.

kattrali:
I see. And if you don’t change the `readwrite` property but add the setter?

fractalwrench:
@kattrali that seems to work, although there's another warning:

`Writable atomic property 'notifyURL' cannot pair a synthesized getter with a user defined setter`

I can ignore that and proceed with the above approach if you're happy?

kattrali:
Haha my happiness is compiler happiness + making usage intent clear :)

Does defining the getter + making the setter deprecated resolve everything?

fractalwrench:
@kattrali yes, that seems to work - I will update the PR accordingly

fractalwrench:
@kattrali for Objective-C yes, that resolves the issues, but the deprecation notice isn't shown in Swift.

fractalwrench:
After discussing this with @snmaynard we decided that of the options below, (2) would be the best:

1. Keep the properties as readwrite and document not to use them
2. Make the properties readonly (breaking change)
3. Deprecate the properties entirely

I'm not aware of anything else outstanding on this PR.

Tests/BugsnagSinkTests.m
```diff
@@ -39,7 +39,7 @@ - (void)setUp {
     config.autoNotify = NO;
     config.apiKey = @"apiKeyHere";
     config.releaseStage = @"MagicalTestingTime";
-    config.notifyURL = nil;
+    [config setEndpointsForNotify:@"http://localhost:8000" sessions:@"http://localhost:8000"];
```
CodeHex: The notifyURL was cleared previous, but now this is being set to localhost:8000, is this correct? Also sessions endpoint is localhost:10000

PR #289, opened 2018-06-11
Handle nil content value

fractalwrench:
## Goal

The notifier currently assumes that the "value" key in notable_addresses data is non-null, whereas this is not necessarily the case, which can lead to an `NSInvalidArgumentException`.

This change adds unit test coverage for this method, and ensures that it can handle any of the items in the dictionary being null.

### Added

Unit Tests which cover behaviour of the class.

### Changed

Skip iteration if the contentValue is nil.

## Review

  - [x] Final review

### DIFF
--- a/Source/BugsnagCrashReport.m
+++ b/Source/BugsnagCrashReport.m
@@ -630,9 +630,13 @@ initWithErrorName:(NSString *_Nonnull)name
             }
             NSString *contentValue = data[@"value"];
 
+            if (!contentValue) {
+                continue;
+            }
+
             if ([self isReservedWord:contentValue]) {
                 reservedWord = contentValue;
-            } else if (!([[contentValue componentsSeparatedByString:@"/"] count] > 2)) {
+            } else if ([[contentValue componentsSeparatedByString:@"/"] count] <= 2) {
                 // must be a string that isn't a reserved word and isn't a filepath
                 [interestingValues addObject:contentValue];
             }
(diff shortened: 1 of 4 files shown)


Source/BugsnagCrashReport.m
```diff
@@ -630,9 +630,13 @@ + (instancetype)errorDataFromThreads:(NSArray *)threads {
...
             NSString *contentValue = data[@"value"];
 
+            if (!contentValue) {
+                continue;
+            }
+
             if ([self isReservedWord:contentValue]) {
                 reservedWord = contentValue;
-            } else if (!([[contentValue componentsSeparatedByString:@"/"] count] > 2)) {
+            } else if ([[contentValue componentsSeparatedByString:@"/"] count] <= 2) {
```
fractalwrench: This change just simplifies the condition to make it more readable

fractalwrench:
I've added both an implicit and explicit test case for where "value" can be nil. I've also added a type check as the original issue occurred if `[NSNull null]` was present, as this wouldn't evaluate to false but is nil.

Source/BugsnagCrashReport.m
```diff
@@ -630,9 +630,12 @@ + (instancetype)errorDataFromThreads:(NSArray *)threads {
             }
             NSString *contentValue = data[@"value"];
 
-            if (!contentValue) {
+#pragma clang diagnostic push
+#pragma ide diagnostic ignored "OCDFAInspection"
```
fractalwrench: suppresses warnings on a redundant nsstring type check, which actually isn't redundant when `contentValue` is `NSNull`

PR #294, opened 2018-07-12
Ensure sessions are sent in the background

snmaynard:
When starting the app we can send sessions on the main thread which causes a slowdown on boot. This changes the timer to fire immediately rather than calling the session tick function manually.


snmaynard:
I think that this implementation will still cause the timer to fire on the main thread so this needs a bigger rethink. I lack the context on the decision to use a timer rather than just a network threadpool that we push to here so I will hold off on making larger changes.

kattrali:
Replaced by #295

PR #295, opened 2018-07-13
Fix session delivery logic to always use the delivery queue

kattrali:
## Goal

Streamline session delivery, ensuring requests are sent on the API client delivery queue.

### `SessionTracker`

* Moved the event handling for entering/leaving the foreground here
* Separated starting a session manually from automatic session tracking logic
* Removed delivery logic

### `Notifier`

* Manually starts a new session in the `start()` method
* Calls new session tracker event hooks

### `SessionTrackingApiClient`

* Moved all of the logic for delivering sessions here

## Tests

* Added new unit test cases for starting sessions manually vs auto-capture to ensure sessions are only recorded when `shouldAutoCaptureSessions` is true.
* Refactored integration test cases for starting sessions and automatic session tracking to test the default behavior without setting `shouldAutoCaptureSessions`

## Discussion

* This changeset removes the logic in `BugsnagConfiguration` to start a session if no session has ever been started in the setter for `shouldAutoCaptureSessions` because of a possible race condition. The inline docs are updated accordingly, but is this behavior needed?

### Alternative Approaches

* I considered only doing the `SessionTrackingApiClient` change, however the larger changeset makes the flow easier to follow and removes the timer logic. This is still a possibility if reviewing a smaller changeset is preferable.

## Review

  - [x] Final review


Source/BugsnagConfiguration.m
```diff
@@ -218,23 +218,6 @@ - (void)setApiKey:(NSString *)apiKey {
...
-
-- (BOOL)shouldAutoCaptureSessions {
-    return _shouldAutoCaptureSessions;
-}
-
-- (void)setShouldAutoCaptureSessions:(BOOL)shouldAutoCaptureSessions {
-    @synchronized (self) {
-        _shouldAutoCaptureSessions = shouldAutoCaptureSessions;
-
-        if (shouldAutoCaptureSessions) { // track any existing sessions
```
fractalwrench: IIRC this logic (or similar) may have been added to enable automatic session tracking on React Native. Have we tested the effects of these changes on downstream notifiers?

Source/BugsnagSessionTracker.h
```diff
@@ -17,25 +17,48 @@ typedef void (^SessionTrackerCallback)(BugsnagSession *newSession);
...
+- (void)handleAppBackgroundEvent;
 
 /**
- * Called when a session is altered
+ Handle some variation of Bugsnag.notify() being called.
+ Increases the number of handled errors recorded for the current session, if
+ a session exists.
  */
-@property SessionTrackerCallback callback;
+- (void)handleHandledErrorEvent;
```
fractalwrench: Would it also make sense to have an unhandled equivalent of this method?

Source/BugsnagNotifier.m
```diff
@@ -383,34 +388,15 @@ - (void)watchLifecycleEvents:(NSNotificationCenter *)center {
...
-
-    if (!self.sessionTimer) {
-        _sessionTimer = [NSTimer scheduledTimerWithTimeInterval:sessionTickSeconds
-                                                         target:self
-                                                       selector:@selector(sessionTick:)
-                                                       userInfo:nil
-                                                        repeats:YES];
-        [self sessionTick:self];
-    }
+    [self.sessionTracker handleAppForegroundEvent];
```
fractalwrench: Will this capture 2 sessions on application startup rather than 1? `startNewSessionIfAutoCaptureEnabled` will also start a session, unless I'm mistaken.

Source/BugsnagSessionTracker.m
```diff
@@ -41,92 +50,61 @@ - (instancetype)initWithConfig:(BugsnagConfiguration *)config
...
-        self.currentSession.handledCount++;
-        if (self.callback && (self.config.shouldAutoCaptureSessions || !self.currentSession.autoCaptured)) {
-            self.callback(self.currentSession);
-        }
+- (void)handleAppForegroundEvent {
+    if (self.backgroundStartTime
+        && [[NSDate date] timeIntervalSinceDate:self.backgroundStartTime] >= BSGNewSessionBackgroundDuration) {
+        [self startNewSessionIfAutoCaptureEnabled];
     }
+    self.backgroundStartTime = nil;
```
fractalwrench: If this can be nil and we're expecting sessions to be started from multiple threads, would it make sense to make this property atomic/make any checks compound operations?

Source/BugsnagSessionTrackingApiClient.m
```diff
@@ -6,15 +6,59 @@
...
 @implementation BugsnagSessionTrackingApiClient
 
 - (NSOperation *)deliveryOperation {
     return [NSOperation new];
 }
 
+- (void)deliverSessionsInStore:(BugsnagSessionFileStore *)store {
+    [self.sendQueue addOperationWithBlock:^{
+        if (!self.config.apiKey)
+            return;
```
fractalwrench: Is it worth logging a warning if this is the case? I believe this state shouldn't really be allowed.

Tests/BugsnagSessionTrackerTest.m
```diff
@@ -25,73 +25,100 @@ - (void)setUp {
...
-    [self.sessionTracker startNewSession:[NSDate date]
-                                withUser:self.user
-                            autoCaptured:NO];
+    [self.sessionTracker startNewSession];
 
     BugsnagSession *secondSession = self.sessionTracker.currentSession;
     XCTAssertNotEqualObjects(firstSession.sessionId, secondSession.sessionId);
 }
 
 - (void)testIncrementCounts {
```
fractalwrench: Is it possible to test unhandled counts? Or is that already covered elsewhere?

features/fixtures/ios-swift-cocoapods/iOSTestApp/AppDelegate.swift
```diff
@@ -15,6 +15,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         let arguments = ProcessInfo.processInfo.arguments
         var delay: TimeInterval = 0
         var eventType = "none"
+        var eventMode = "regular"
```
fractalwrench: What's the difference between an event type and an event mode? Do they need to be separate concepts?

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/ReadGarbagePointerScenario.m
```diff
@@ -32,6 +32,11 @@
  */
 @implementation ReadGarbagePointerScenario
 
+- (void)startBugsnag {
+    self.config.shouldAutoCaptureSessions = NO;
```
fractalwrench: Is it worth making this the default in the mazerunner fixture's AppDelegate, seeing as most of the scenarios already assume this?

Source/BugsnagNotifier.m
kattrali: It only captures 1 event, since the foreground session only happens when coming from the background and having `backgroundTime` more than 60s before.

Source/BugsnagSessionTracker.h
kattrali: Not really; if an unhandled event occurs, the app is crashing. There can’t be more than 1 unhandled event in a single session.

Tests/BugsnagSessionTrackerTest.m
kattrali: It needs an integration test case. I’ll note to add one.

features/fixtures/ios-swift-cocoapods/iOSTestApp/AppDelegate.swift
kattrali: There needs to be separation between “run this scenario’s crashing code” and “configure Bugsnag for this scenario (and not run the crashy code”. Without this, an individual scenario can’t configure Bugsnag for relaunching the app.

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/ReadGarbagePointerScenario.m
kattrali: No, because having it there made it difficult to test the default behavior for session tracking. For example, the auto-capture scenarios should test the behavior when not setting the value at all to ensure the default behavior is to send a session request. It’s also easier to debug an individual case when only setting the minimal values for running the scenario by default. I lost a bit of time to confusion when I couldn’t figure out where a config value was being set between AppDelegate, Scenario, and the Scenario subclass.

Source/BugsnagSessionTrackingApiClient.m
kattrali: Logging is a good idea. It shouldn’t happen but this is a guard to prevent either crashing or sending an invalid payload if it hasn’t been set.

Source/BugsnagConfiguration.m
kattrali: bugsnag-react-native will need a slight tweak in the initialization logic. While configuring the JS options, when all of the following are true:

* Bugsnag has already been configured on the native side
* The native configuration has `shouldAutoCaptureSessions` set to `NO`
* The JS configuration has `autoCaptureSessions` set to `true`

Then `[BugsnagReactNative start]` should call `startSession`

martin308: Yeh there wouldn't be an issue with it the way it is setup right now. For the native notifiers we turn off automatic session tracking

kattrali:
I think the review feedback has been addressed:

* Added logging to the unlikely error cases
* Documented the main thread-based hooks
* Added tests for the handled/unhandled event counts in report payloads

PR #296, opened 2018-07-16
fix: Parse system info from root if no nested system dictionary exists

kattrali:
In reports, the base report has a “system” dict for system info.

Sessions rely on the raw info parser to gather app and device data, 
which is not nested in a separate “system” dict. 

## Goal

The changes in #292 fixed sending a custom version for reports but
broke app version parsing for sessions. This change refactors app
state parsing slightly to ensure both reports and sessions fetch the
correct app state (and custom versions, if provided).

## Changeset

Moved the top level version parsing out to error reports from `BSGParseAppState`. The preferred app version is now (optionally) specified up front.

```diff
- BSGParseAppState(errorReport)
+ BSGParseAppState(sysInfo, preferredVersion)
```

## Tests

Added new integration test cases for verifying app version with or without custom versions set.

## Review

  - [x] Final review


Source/BugsnagKSCrashSysInfoParser.m
```diff
@@ -85,17 +85,12 @@
...
 
-    NSString *version = [report valueForKeyPath:@"user.config.appVersion"];
-
-    if (!version) {
-        version = system[@"CFBundleShortVersionString"];
-    }
+    NSString *version = preferredVersion ?: report[@"CFBundleShortVersionString"];
 
-    BSGDictSetSafeObject(app, system[@"CFBundleVersion"], @"bundleVersion");
+    BSGDictSetSafeObject(app, report[@"CFBundleVersion"], @"bundleVersion");
```
martin308: should `bundleVersion` be a const like the other keys? `BSGKeyReleaseStage` etc.

kattrali: Maybe; I want to think about how we are doing key constants a bit in the meantime in case there's an easy way to both define and find the right ones.

PR #298, opened 2018-07-17
Add `osBuild` as a reported device value

sberrevoets:
This adds `osBuild` as a new value that's being reported for every
device. This can be useful to tell apart release versions from beta
versions of the OS.

The technical name used by `sysctl` is "os version", but since the
`osVersion` key was already in use for reporting the marketing version
(e.g. 11.4), I used `osBuild` as the key instead.

Tests are updated and pass. I'm not entirely sure if there's more
that needs to be done to make this visible in the web UI, but it seems
like this should do it.


sberrevoets:
Thanks! Would be great if we could get a release with this so we can start testing iOS 12 😃

PR #301, opened 2018-07-31
Deregister notification observers and listeners before application termination

fractalwrench:
## Goal

Deregister any listeners or observers when the app is about to terminate, to prevent allocations from occurring during `dealloc`.

## Changeset

- Update Podfiles for example apps to latest version
- Add an observer for termination notifications for each platform
- Remove all notification observers when a termination notification is received
- Stop watching connectivity changes when about to terminate
- Stop watching orientation/battery monitoring when about to terminate

## Tests

Ran the iOS and macOS example apps and confirmed the `stop` method was invoked, and that no crashes were detected on exiting the application

## Review

  - [x] Final review


Source/BugsnagNotifier.m
```diff
@@ -359,18 +364,38 @@ - (void)start {
...
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(stop:)
+                                                 name:name
+                                               object:nil];
+}
+
+/**
+ * Removes observers and listeners to prevent allocations when the app is terminated
+ */
+- (void)stop:(id)sender {
```
kattrali: I'd name this a bit more verbosely, like `unsubscribeFromNotifications`. There is an off chance that with a name like `stop`, it could be picked up erroneously in app review as a private API usage.

PR #303, opened 2018-08-02
Capture trace of error reporting thread and identify with boolean flag

fractalwrench:
## Goal

Captures the trace of the crashed thread, which was previously omitted, and to identify it with a boolean flag in the serialised JSON.

## Changeset

- Add the current thread trace during serialisation
- Conditionally write `errorReportingThread` boolean flag by checking `crashed` flag

Note: the thread name is not captured due to known limitations with how KSCrash captures this information.

## Tests

Added unit tests for general thread serialisation, and for checking that the current thread is now captured. Tested with dashboard.

## Review

  - [x] Final review


Source/BugsnagCrashReport.m
```diff
@@ -579,29 +580,38 @@ - (NSArray *)serializeThreadsWithException:(NSMutableDictionary *)exception {
...
+        [self serialiseThread:bugsnagThreads thread:thread backtrace:backtrace crashedThread:isCrashedThread];
+    }
+    return bugsnagThreads;
+}
 
-            for (NSDictionary *frame in backtrace) {
+- (void)serialiseThread:(NSMutableArray *)bugsnagThreads
+                 thread:(NSDictionary *)thread
+              backtrace:(NSArray *)backtrace
+          crashedThread:(BOOL)isCrashedThread {
```
snmaynard: I would be careful to call this the reportingThread or the currentThread to emphasize that its not the crashing thread in all cases.

fractalwrench:
Labelling as WIP as the mazerunner scenario appears to have failed on Travis CI, and the build queue is currently blocked which means I'm unlikely to investigate this further by the end of today :(

PR #308, opened 2018-08-16
Prevent potential crash during app teardown

fractalwrench:
## Goal

If the session API client is garbage collected while sessions are enqueued to be delivered, there
can be a crash when accessing self.config.apiKey from the operation queue. This changeset fixes that by copying the API key/session URL accessed from within the operation block.

## Changeset

Copy the `sessionURL` and `apiKey` values so that `self` is not accessed from the operation block.

## Tests

Existing unit tests/mazerunner.

## Review

  - [x] Final review

### DIFF
--- a/Source/BugsnagSessionTrackingApiClient.m
+++ b/Source/BugsnagSessionTrackingApiClient.m
@@ -19,8 +19,8 @@
 }
 
 - (void)deliverSessionsInStore:(BugsnagSessionFileStore *)store {
-    NSString *apiKey = self.config.apiKey;
-    NSURL *sessionURL = self.config.sessionURL;
+    NSString *apiKey = [self.config.apiKey copy];
+    NSURL *sessionURL = [self.config.sessionURL copy];
 
     [self.sendQueue addOperationWithBlock:^{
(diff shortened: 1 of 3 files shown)


snmaynard:
I feel like we might want to call `cancelAllOperations` in dealloc of the api client instead. Do you have thoughts @kattrali?

fractalwrench:
@kattrali is Simon's suggestion something we might want to do, in addition to or in place of the fix that was originally proposed?

Source/BugsnagSessionTrackingApiClient.m
```diff
@@ -19,8 +19,11 @@ - (NSOperation *)deliveryOperation {
 }
 
 - (void)deliverSessionsInStore:(BugsnagSessionFileStore *)store {
+    NSString *apiKey = self.config.apiKey;
+    NSURL *sessionURL = self.config.sessionURL;
```
kattrali: Both of these values should be [copied](https://developer.apple.com/documentation/objectivec/nsobject/1418807-copy?language=objc)

PR #312, opened 2018-09-07
Handle NSExceptions within C++ exception handler

fractalwrench:
## Goal

Capture uncaught NSExceptions when PLCrashReporter is initialised in a project after Bugsnag.

## Design

`NSException` crashes were previously discarded from within Bugsnag's C++ exception handler, as they were handled by `NSUncaughtExceptionHandler` (within `BSG_KSCrashSentry_NSException`) instead.

PLCrashReporter does not appear to call the previously registered `NSUncaughtExceptionHandler`, meaning that if PLCR was initialised after Bugsnag, no report could be captured. However, Bugsnag's C++ exception handler is triggered when PLCR is installed. This changeset adapts the C++ exception handler to serialise `NSException` instead of `BSG_KSCrashSentry_NSException`.

## Changeset

- Move the logic for capturing `NSException` into the C++ sentry
- Stop registering `BSG_KSCrashSentry_NSException`

## Tests

Raised an `NSException` in the objective-c-ios example app and verified that a report was received by the Bugsnag dashboard for the following matrix:

- Only bugsnag installed
- Bugsnag initialised first
- PLCrashReporter initialised first

## Review

  - [ ] Final review

### DIFF
--- a/Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
+++ b/Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
@@ -222,9 +222,8 @@ static void CPPExceptionTerminate(void) {
             @"Crash handling complete. Restoring original handlers.");
         bsg_kscrashsentry_uninstall((BSG_KSCrashType)BSG_KSCrashTypeAll);
         bsg_kscrashsentry_resumeThreads();
+        bsg_g_originalTerminateHandler();
     }
-
-    bsg_g_originalTerminateHandler();
 }
 
 // ============================================================================
--- a/Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
+++ b/Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
@@ -155,8 +155,7 @@ static void CPPExceptionTerminate(void) {
     try {
         throw;
     } catch (NSException *exception) {
-        BSG_KSLOG_DEBUG(@"Detected NSException. Letting the current "
-                        @"NSException handler deal with it.");
+        BSG_KSLOG_DEBUG(@"Detected NSException, beginning report serialisation.");
         isNSException = true;
         handleNSException(exception);
     } catch (std::exception &exc) {
--- a/Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
+++ b/Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
@@ -222,7 +222,10 @@ static void CPPExceptionTerminate(void) {
             @"Crash handling complete. Restoring original handlers.");
         bsg_kscrashsentry_uninstall((BSG_KSCrashType)BSG_KSCrashTypeAll);
         bsg_kscrashsentry_resumeThreads();
-        bsg_g_originalTerminateHandler();
+
+        if (bsg_g_originalTerminateHandler != NULL) {
+            bsg_g_originalTerminateHandler();
+        }
     }
 }
 
--- a/Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
+++ b/Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
@@ -130,6 +130,13 @@ static void handleNSException(NSException *exception) {
 
         BSG_KSLOG_DEBUG(@"Crash handling complete. Restoring original handlers.");
         bsg_kscrashsentry_uninstall((BSG_KSCrashType)BSG_KSCrashTypeAll);
+
+        NSUncaughtExceptionHandler* handler = NSGetUncaughtExceptionHandler();
+
+        if (handler != NULL) {
+            BSG_KSLOG_DEBUG(@"Calling original exception handler.");
+            handler(exception);
+        }
     }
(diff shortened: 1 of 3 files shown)


Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
```diff
@@ -92,6 +92,47 @@ void __cxa_throw(void *thrown_exception, std::type_info *tinfo,
...
+            bsg_kscrashsentry_uninstall((BSG_KSCrashType)BSG_KSCrashTypeAll);
+        }
+
+        BSG_KSLOG_DEBUG(@"Suspending all threads.");
+        bsg_kscrashsentry_suspendThreads();
+
+        BSG_KSLOG_DEBUG(@"Filling out context.");
+        NSArray *addresses = [exception callStackReturnAddresses];
+        NSUInteger numFrames = [addresses count];
+        uintptr_t *callstack = new uintptr_t[numFrames];
```
fractalwrench: adapted memory allocation as this compiles as C++, not C

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
```diff
@@ -92,6 +92,47 @@ void __cxa_throw(void *thrown_exception, std::type_info *tinfo,
 }
 }
 
+static void handleNSException(NSException *exception) {
```
fractalwrench: this mostly originates from the handler code in BSG_KSCrashSentry_NSException, with a few adaptations

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
```diff
@@ -92,6 +92,47 @@ void __cxa_throw(void *thrown_exception, std::type_info *tinfo,
...
+        bsg_g_context->NSException.name = strdup([[exception name] UTF8String]);
+        bsg_g_context->crashReason = strdup([[exception reason] UTF8String]);
+        bsg_g_context->stackTrace = callstack;
+        bsg_g_context->stackTraceLength = (int)numFrames;
+
+        BSG_KSLOG_DEBUG(@"Calling main crash handler.");
+        bsg_g_context->onCrash();
+
+        BSG_KSLOG_DEBUG(@"Crash handling complete. Restoring original handlers.");
+        bsg_kscrashsentry_uninstall((BSG_KSCrashType)BSG_KSCrashTypeAll);
```
fractalwrench: we do not call through to the previous handler here as if we were installed last it is redundant

kattrali: Instead of calling to the last C++ handler, this should use [NSGetUncaughtExceptionHandler](https://developer.apple.com/documentation/foundation/1416853-nsgetuncaughtexceptionhandler?language=objc) and call the NSException handler, as it may be set.

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
```diff
@@ -117,6 +158,7 @@ static void CPPExceptionTerminate(void) {
         BSG_KSLOG_DEBUG(@"Detected NSException. Letting the current "
```
kattrali: This comment should be updated somewhat to reflect what is happening here.

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
fractalwrench: I have updated this logic so that we call the `NSUncaughtExceptionHandler` if an `NSException` is triggered, and the original terminate handler if a C++ exception occurs.

fractalwrench:
I have updated this PR to address all the outstanding inline comments by:

- Calling the previous NSUncaughtExceptionHandler when an NSException occurs, and the previous terminate handler only when a C++ exception occurs (leaving the terminate handler as-is led to duplicate callbacks in PLCR for NSException)
- Updating the log statement

I verified this via manual testing in the example app for the two basic scenarios (configuring PLCR first, and configuring Bugsnag first)

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
```diff
@@ -174,9 +222,8 @@ static void CPPExceptionTerminate(void) {
             @"Crash handling complete. Restoring original handlers.");
         bsg_kscrashsentry_uninstall((BSG_KSCrashType)BSG_KSCrashTypeAll);
         bsg_kscrashsentry_resumeThreads();
+        bsg_g_originalTerminateHandler();
```
bengourley: According to [this documentation](http://www.cplusplus.com/reference/exception/set_terminate/) the returned pointer from `std::setTerminate(fn)` can be a `null` pointer.

bengourley:
I tested out the changes and it still works on my project.

fractalwrench:
I have added a null check for the previous terminate handler.

@kattrali did you have further details about the potential memory leak you mentioned? Is this still something that needs addressing?

fractalwrench:
@snmaynard calling the previous `NSUncaughtExceptionHandler` instead of the C++ terminate handler for NSExceptions was part of a previously addressed [review comment](https://github.com/bugsnag/bugsnag-cocoa/pull/312#discussion_r216187441).

PR #315, opened 2018-10-23
Add Device time of report capture to JSON payload

fractalwrench:
## Goal

Reports may be cached on a device if no network connection is available. In this case, the report may not be delivered for hours, or even days, after the error was originally captured by Bugsnag.

Sending the time at which the report was originally captured allows for timestamps to be normalised. This avoids confusing information similar to the following being shown:

screen shot 2018-10-23 at 16 19 51

## Design

The pipeline normalises timestamps based off the `device.time` field.

## Changeset

Serialises the time at which the report was generated into the `device.time` payload field.

## Tests

- Added mazerunner tests for handled/unhandled errors, payload serialisation unit test
- Checked payload manually for presence of `device.time` in handled/unhandled error
- Checked breadcrumb time is updated in the dashboard when sending in a payload that has been cached and had the timestamp manually edited

## Review

- [ ] Check the scope of the changeset - is everything in the diff required for the pull request?
  - [ ] Final review

### DIFF
--- a/features/handled_errors.feature
+++ b/features/handled_errors.feature
@@ -6,7 +6,7 @@ Scenario: Override errorClass and message from a notifyError() callback
     And the request is a valid for the error reporting API
     And the exception "errorClass" equals "Bar"
     And the exception "message" equals "Foo"
-    And the payload field "events.0.device.time" is not null
+    And the event "device.time" is a timestamp
 
 Scenario: Reporting an NSError
     When I run "HandledErrorScenario" with the defaults on "iPhone8-11.2"
--- a/features/handled_errors.feature
+++ b/features/handled_errors.feature
@@ -6,7 +6,7 @@ Scenario: Override errorClass and message from a notifyError() callback
     And the request is a valid for the error reporting API
     And the exception "errorClass" equals "Bar"
     And the exception "message" equals "Foo"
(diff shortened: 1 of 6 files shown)


features/handled_errors.feature
```diff
@@ -6,6 +6,7 @@ Scenario: Override errorClass and message from a notifyError() callback
     And the request is a valid for the error reporting API
     And the exception "errorClass" equals "Bar"
     And the exception "message" equals "Foo"
+    And the payload field "events.0.device.time" is not null
```
snmaynard: I would like to verify the format of this field and also that it is the current time.

fractalwrench:
I have updated the mazerunner scenario to check that the field is a timestamp. There's also some coverage in the unit tests for ensuring the timestamp is serialised from the KSCrash field, if that's sufficient?

Note: Cocoa mazerunner scenarios still do not work on my machine and this makes it very hard to write/debug scenarios.

fractalwrench:
I have updated the mazerunner scenario to check that the field is a timestamp within 30s of the current time .

features/steps/ios_steps.rb
```diff
@@ -1,3 +1,5 @@
+require 'date'
```
snmaynard: You dont use this require.

PR #320, opened 2018-11-29
Strip top bits from pointers before reading memory contents

### LINKED ISSUE #318: Messages from fatalError/etc are no longer propagated on Bugsnag reports
### Expected behavior
Forcing the app to crash by calling `fatalError("i am a message")` includes the "i am a message" text in the bugsnag report.

### Observed behavior
No message is included:
https://app.bugsnag.com/recharge-labs/ios-consumer/errors/5bbe93a371c2fc001a3e9ba8

### Version
Bugsnag 5.17.0
Xcode Version 10.0 (10A255)

### Additional information
Original feature request (#159) and PR to add the feature (#188).  I'm guessing that something in Swift 4.2 or Xcode 10 changed the way those notable addresses are attached to the crash report?  Or maybe broke the heuristics in some other way?

kattrali:
In the new Xcode 10 build system, Swift object register values have the
top bit used as a flag. This change strips the flag while not losing
anything relevant to us in our quest to see error messages for assertion
failures.

## Design

Added a new validation to `bsg_kscrw_i_writeMemoryContentsIfNotable` which strips the top from the address if it is not valid and rechecks the new value. If it is then valid, it is used as the object address.

## Tests

* Tested manually on physical devices and emulators for iOS 11 & 12
* Added a new integration test, which runs successfully on Xcode 10.1 + iOS 12.1

## Discussion

This technique no longer works at all for assertion messages less than 16 characters, as they aren't stored in the register values.

### Linked issues

Fixes #318

## Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ Changelog
   addresses
   [#319](https://github.com/bugsnag/bugsnag-cocoa/pull/319)
 
+* Add `fatalError` and other assertion failure messages in reports for
+  Swift 4.2 apps. Note that this only includes messages which are 16
+  characters or longer. See the linked pull request for more information.
+  [#320](https://github.com/bugsnag/bugsnag-cocoa/pull/320)
+
 ## 5.17.0 (2018-09-25)
 
 ### Enhancements
--- a/features/crashprobe.feature
+++ b/features/crashprobe.feature
@@ -166,8 +166,15 @@ Scenario: Crash within Swift code
     And the exception "message" equals "Unexpectedly found nil while unwrapping an Optional value"
     And the exception "errorClass" equals "Fatal error"
 
-    And the "method" of stack frame 0 starts with "_T0s18_fatalErrorMessages5NeverOs12StaticStringV_A2E4fileSu4lines6UInt32V5"
-    And the "method" of stack frame 1 starts with "_T010iOSTestApp10SwiftCrashC3run"
+Scenario: Assertion failure in Swift code
+    When I set environment variable "BUGSNAG_API_KEY" to "a35a2a72bd230ac0aa0f52715bbdc6aa"
+    And I configure the app to run on "iPhone8-11.2"
+    And I crash the app using "SwiftAssertion"
+    And I relaunch the app
+    Then I should receive a request
+    And the request is a valid for the error reporting API
+    And the exception "errorClass" equals "Fatal error"
+    And the exception "message" equals "several unfortunate things just happened"
 
 Scenario: Dereference a null pointer
     When I set environment variable "BUGSNAG_API_KEY" to "a35a2a72bd230ac0aa0f52715bbdc6aa"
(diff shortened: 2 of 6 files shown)


features/crashprobe.feature
```diff
@@ -166,8 +166,15 @@ Scenario: Crash within Swift code
     And the exception "message" equals "Unexpectedly found nil while unwrapping an Optional value"
     And the exception "errorClass" equals "Fatal error"
 
-    And the "method" of stack frame 0 starts with "_T0s18_fatalErrorMessages5NeverOs12StaticStringV_A2E4fileSu4lines6UInt32V5"
-    And the "method" of stack frame 1 starts with "_T010iOSTestApp10SwiftCrashC3run"
+Scenario: Assertion failure in Swift code
```
fractalwrench: If it's not possible to capture messages <16 characters on Swift 4.2, I think it'd be good to add a scenario to verify this behaviour.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ Changelog
   addresses
   [#319](https://github.com/bugsnag/bugsnag-cocoa/pull/319)
 
+* Add `fatalError` and other assertion failure messages in reports for
+  Swift 4.2 apps
```
fractalwrench: It'd be worth noting that we can't capture messages <16 characters

features/crashprobe.feature
kattrali: Its not the _intended_ behavior, just a known limitation. It also would be specific to Xcode 10 until whenever the storage style changes again.

CHANGELOG.md
kattrali: Updated, also updated the commit with some addition information about the limitations:

    In the new Xcode 10 build system, Swift object register values have the
    top bit used as a flag. This change strips the flag while not losing
    anything relevant to us in our quest to see error messages for assertion
    failures.

    This technique does not capture messages which are less than 16
    characters, as short strings are stored as raw char arrays on the stack
    rather than being allocated. (See WWDC 2018 #401 for more info on new
    string optimizations)

    While it is possible to check for char arrays as well as pointers when
    searching for notable address values, sweeping up local variables has a
    likely chance of capturing unintended data as well from the surrounding
    code, some of which may be sensitive. It is also not guaranteed that the
    value would still be on the stack after the message is logged, so it is
    possible to get only unrelated string values as the message.

    In the current Swift stdlib, the following messages passed to
    fatalError, preconditionFailure, and precondition (and their internal
    func counterparts) are less than 16 characters:

    * empty string
    * `unavailable`
    * `not implemented`
    * `abstract method`
    * `unknown value`
    * `invalid count` (where a dictionary contains < 0 items(?))
    * `invalid index` (where a dictionary ceases to be a dictionary)
    * `don't touch me` (from SpriteKit)
    * `close() failed` (from the private Subprocess implementation)

    The vast majority have more meaningful messages.

And a relevant quote from the linked wwdc session:

> If the string fits within 15 bytes then the actual string is represented directly in the string type without going to the heap to allocate a separate buffer to represent the string.

PR #321, opened 2018-12-05
fix: Check for null image name before comparing strings

kattrali:
If the number of loaded images changes between when _dyld_image_count()
is computed and when _dyld_get_image_name() is called, there can be less
images than the count suggests. In this case, _dyld_get_image_name()
returns NULL, and the lack of a check would crash the app (again?) 
because both arguments to strcmp are required to be non-null. The 
resulting exception backtrace looks somewhat like this:

```
_platform_strcmp
bsg_ksdlimageNamed (BSG_KSDynamicLinker.c:41:21)
bsg_ksdlimageUUID (BSG_KSDynamicLinker.c:56:31)
+[BSG_KSSystemInfo appUUID] (BSG_KSSystemInfo.m:158:13)
+[BSG_KSSystemInfo systemInfo] (BSG_KSSystemInfo.m:432:5)
(any access to systemInfo, from an API client, etc)
```

## Tests

This one is difficult to verify manually or to write a test for, (short of swapping out the implementation of `_dyld_image_count()` to immediately unload an image), but the intent should be easy to follow from the source exception.

### Outstanding Questions

  - [x] Final review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## TBD
+
+### Bug Fixes
+
+* Fix possible crash when fetching system info to append to a crash report
+  [#321](https://github.com/bugsnag/bugsnag-cocoa/pull/321)
+
 ## 5.17.2 (2018-12-05)
 
 * Add Device time of report capture to JSON payload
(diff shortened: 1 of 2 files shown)


kattrali:
> BSG_KSSystemInfo.m 364 (and various others

This one in particular is "safe", as its loading the first image from the array, and there's going to be at least one image at all times.

I went through a few other code paths in the dynamic linker though and while I can see how they eventually nope out if the image header or name is null, I'll add a few more explicit checks.

PR #325, opened 2019-02-15
Add stopSession() and resumeSession() to Bugsnag

fractalwrench:
## Goal
Adds the ability to stop and resume sessions. This prevents the handled/unhandled error count from incrementing when a session is in the stopped state, which may be desirable if a user manually tracks sessions and does not care about crashes that occur in the background.

## Changeset

Added the public interface for stopping/resuming a session, along with the initial implementation and unit test coverage.

__NOTE__: mazerunner scenarios and documentation will be added in separate PRs.


fractalwrench:
Thanks for the feedback @kattrali - I've addressed the inline comments and have updated the `BugsnagSessionTracker` implementation to use a `self.runningSession` method.

PR #326, opened 2019-02-19
test: add mazerunner scenarios for verifying stopSession and resumeSe…

fractalwrench:
## Goal

Adds mazerunner scenarios which verify the behaviour of #325

### DIFF
--- a/features/session_stopping.feature
+++ b/features/session_stopping.feature
@@ -1,0 +1,29 @@
+Feature: Stopping and resuming sessions
+
+Scenario: When a session is stopped the error has no session information
+    When I run "StoppedSessionScenario"
+    Then I should receive 2 requests
+    And the request 0 is valid for the session tracking API
+    And the request 1 is valid for the error reporting API
+    And the payload field "events.0.session" is not null for request 1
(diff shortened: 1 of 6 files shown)


features/session_stopping.feature
```diff
@@ -0,0 +1,29 @@
...
+
+Scenario: When a new session is started the error uses different session information
+    When I run "NewSessionScenario"
+    Then I should receive 3 requests
+    And the request 0 is valid for the session tracking API
+    And the request 1 is valid for the session tracking API
+    And the request 2 is valid for the error reporting API
+    And the payload field "events.0.session.events.handled" equals 1 for request 2
+    And the payload field "events.1.session.events.handled" equals 1 for request 2
+    And the payload field "events.0.session.id" of request 1 does not equal the payload field "events.1.session.id" of request 2
```
Cawllec: Should this be `And the payload field "events.0.session.id" of request 2 does not equal the payload field "events.1.session.id" of request 2` ? Rather than comparing request 1 against request 2?

fractalwrench: Good catch - I'll update this step.

PR #327, opened 2019-02-19
Add fallback metadata in the event of report file corruption

kattrali:
## Goal

Capture and report basic report metadata in the event that a crash report file is corrupted or otherwise could not be written.

## Design

The file path calculation logic is altered to include the severity, handled-ness, and first few characters of the error class, which are parsed and used in the event that the file cannot be read.

This change refactors the `sendReports`/`filterReports` parameter types to include the file path, which required some alteration to the `Sink` and `ApiClient` callback type signatures since an array was no longer being passed around (and we only used the count of the array anyhow).

## Tests

* Tested manually on iOS and Apple TV
* Added new automated tests for the file path calculation logic and scenario where the file is corrupted


Source/BugsnagCrashReport.m
```diff
@@ -478,7 +499,11 @@ - (NSDictionary *)toJson {
     BSGDictSetSafeObject(event, BSGFormatSeverity(self.severity), BSGKeySeverity);
     BSGDictSetSafeObject(event, [self breadcrumbs], BSGKeyBreadcrumbs);
     BSGDictSetSafeObject(event, metaData, BSGKeyMetaData);
-    
+
+    if ([self isIncomplete]) {
+        BSGDictSetSafeObject(event, @YES, BSGKeyIncomplete);
```
fractalwrench: Do we want to include the field if it's `false` also? This would allow filtering on the value either way

Source/BugsnagCrashReport.m
```diff
@@ -620,6 +645,42 @@ - (NSString *_Nullable)enhancedErrorMessageForThread:(NSDictionary *_Nullable)th
 
 @end
 
+@implementation FallbackReportData
+
+- (instancetype)initWithMetadata:(NSString *)metadata {
```
fractalwrench: It'd be nice to separate this out into individual methods to improve readability (e.g. one that returns the severity, another that returns the handled state, etc).

Source/BugsnagCrashReport.m
```diff
@@ -620,6 +645,42 @@ - (NSString *_Nullable)enhancedErrorMessageForThread:(NSDictionary *_Nullable)th
 
 @end
 
+@implementation FallbackReportData
```
fractalwrench: It'd be good to add some brief documentation on what 'h' and 'u' mean for future travellers, either here or in `BSG_KSCrashC.c`

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashC.c
```diff
@@ -63,11 +63,51 @@ static char *bsg_g_stateFilePath;
...
+    }
+    if (length > 5) // Remove initial .json from path
+        length -= 5;
+
+    // append contextual info
+    BSG_KSCrash_Context *context = crashContext();
+    filepath[length++] = '-';
+    filepath[length++] = severity;
+    filepath[length++] = '-';
+    filepath[length++] = context->crash.crashType == BSG_KSCrashTypeUserReported ? 'h' : 'u';
```
fractalwrench: AppCode warns that this is attempting to fit an `int` into a `char`, we could add an explicit cast to suppress this.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashC.c
```diff
@@ -63,11 +63,51 @@ static char *bsg_g_stateFilePath;
...
+            break;
+        }
+        filepath[length] = base[length];
+    }
+    if (length > 5) // Remove initial .json from path
+        length -= 5;
+
+    // append contextual info
+    BSG_KSCrash_Context *context = crashContext();
+    filepath[length++] = '-';
```
fractalwrench: It'd be nice to extract the separator to a constant in case it changes in future, and to be consistent with the implementation in `FallbackData`

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry.h
```diff
@@ -162,7 +162,7 @@ typedef struct BSG_KSCrash_SentryContext {
 BSG_KSCrashType
 bsg_kscrashsentry_installWithContext(BSG_KSCrash_SentryContext *context,
                                      BSG_KSCrashType crashTypes,
-                                     void (*onCrash)(void));
+                                     void (*onCrash)(char, char *));
```
fractalwrench: The `KSCrashSentry_Tests` setup hasn't been updated to use this new signature (although it doesn't seem to have much of an effect on the overall result of the tests)

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_Deadlock.m
```diff
@@ -114,7 +114,7 @@ - (void)handleDeadlock {
     bsg_g_context->registersAreValid = false;
 
     BSG_KSLOG_DEBUG(@"Calling main crash handler.");
-    bsg_g_context->onCrash();
+    bsg_g_context->onCrash('e', "");
```
fractalwrench: Do we want to return something that indicates this is a deadlock rather than an empty string?

Tests/BSGFilepathTests.m
```diff
@@ -0,0 +1,50 @@
...
+
+
+int bsg_create_filepath(char *base, char filepath[512], char severity, char error_class[21]);
+
+@interface BSGFilepathTests : XCTestCase
+
+@end
+
+@implementation BSGFilepathTests
+
```
fractalwrench: It'd be good to verify this method gracefully handles a empty/null errorClass (and possibly other params)

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_Signal.c
```diff
@@ -107,7 +107,16 @@ void bsg_kssighndl_i_handleSignal(int sigNum, siginfo_t *signalInfo,
         bsg_g_context->signal.signalInfo = signalInfo;
 
         BSG_KSLOG_DEBUG("Calling main crash handler.");
-        bsg_g_context->onCrash();
+        char errorClass[21];
+        const char *sigName = bsg_kssignal_signalName(sigNum);
```
fractalwrench: the return value of `sigName` can be NULL, could that cause problems here?

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_User.c
```diff
@@ -92,7 +93,9 @@ void bsg_kscrashsentry_reportUserException(const char *name, const char *reason,
         bsg_g_context->userException.state = appState;
 
         BSG_KSLOG_DEBUG("Calling main crash handler.");
-        bsg_g_context->onCrash();
+        char errorClass[21];
+        strncpy(errorClass, bsg_g_context->userException.name, sizeof(errorClass));
+        bsg_g_context->onCrash(severity[0], errorClass);
```
fractalwrench: Can we trust that `severity` will always be a valid value here?

Source/BugsnagFileStore.m
```diff
@@ -128,13 +128,17 @@ - (NSUInteger)fileCount {
...
     NSArray *fileIds = [self fileIds];
-    NSMutableArray *files =
-            [NSMutableArray arrayWithCapacity:[fileIds count]];
+    NSMutableDictionary *files =
+    [NSMutableDictionary dictionaryWithCapacity:[fileIds count]];
     for (NSString *fileId in fileIds) {
         NSDictionary *fileContents = [self fileWithId:fileId];
         if (fileContents != nil) {
-            [files addObject:fileContents];
+            [files setObject:fileContents forKey:fileId];
```
fractalwrench: AppCode warns that modern Obj-C syntax can be used here: `files[fileId] = fileContents`;

features/crashprobe.feature
```diff
@@ -218,3 +218,14 @@ Scenario: Access a non-object as an object
...
+Scenario: Crash report file corruption
+    When I set environment variable "BUGSNAG_API_KEY" to "a35a2a72bd230ac0aa0f52715bbdc6aa"
+    And I crash the app using "AccessNonObjectScenario"
+    And I corrupt all reports on disk
+    And I relaunch the app
+    Then I should receive a request
+    And the request is a valid for the error reporting API
+    And the exception "errorClass" equals "EXC_BAD_ACCESS"
+    And the event "unhandled" is true
+    And the event "severity" equals "error"
```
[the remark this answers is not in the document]
snmaynard: We should check the incomplete flag is set. I think that achieves what @fractalwrench wants without testing the test framework.

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_Signal.c
snmaynard: yes

Source/BugsnagApiClient.h
```diff
@@ -7,7 +7,7 @@
 
 @class BugsnagConfiguration;
 
-typedef void (^RequestCompletion)(id data, BOOL success, NSError *error);
+typedef void (^RequestCompletion)(NSUInteger reportCount, BOOL success, NSError *error);
```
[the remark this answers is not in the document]
kattrali: `RequestCompletion` is an alias for `BSG_KSCrashReportFilterCompletion`, which did need to change. `RequestCompletion` only exists to avoid importing KSCrash internals headers into global scope.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashC.c
kattrali: Not sure why it would feel that way, considering that `'h'` and `'u'` are chars.

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_Deadlock.m
kattrali: This sentry is unused.

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_User.c
kattrali: Yes, but I added a check for an uncertain future.

PR #332, opened 2019-03-11
fix: BSGConnectivity hostName

Jekiwijaya:
## Goal
Fixing `https://notify.bugsnag.com` is being unresolved, since it's include the protocol, the accepted one is only hostname.
Group

https://developer.apple.com/documentation/systemconfiguration/1514904-scnetworkreachabilitycreatewithn
> nodename
The node name of the desired **host**. This name is the same as that passed to the gethostbyname or getaddrinfo functions. The value of this parameter is copied into the new object.

## Changeset
Changes
```diff
- NSString *hostName = [url absoluteString];
+ NSString *hostName = [url host];
```

## Tests
It should be resolved
Group2

### Outstanding Questions

  - [x] Final review

- [x] The correct target branch has been selected (`master` for fixes, `next` for
  features)


kattrali:
Thank you, @Jekiwijaya!

PR #333, opened 2019-03-12
fix: Add helper for syncing a unity session to native reports

kattrali:
## Goal

Allow registering a session with the `SessionTracker` to be propagated to crash reports. This allows the session to be tracked separately via bugsnag-unity while preserving the stability score. The change is internally facing only.

Related to bugsnag/bugsnag-unity#138

## Tests

Tested manually as a part of bugsnag/bugsnag-unity#138

### Outstanding Questions

  - [x] Final review

### DIFF
--- a/Source/BugsnagNotifier.h
+++ b/Source/BugsnagNotifier.h
@@ -29,7 +29,7 @@
 #import "BugsnagConfiguration.h"
 #import "BugsnagMetaData.h"
 
-@class BSGConnectivity;
+@class BSGConnectivity, BugsnagSessionTracker;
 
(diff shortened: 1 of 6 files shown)


fractalwrench:
This will need a changelog entry when the changes are complete.

kattrali:
@fractalwrench Will it? Its entirely internally facing and documented in bugsnag/bugsnag-unity#138.

fractalwrench:
@kattrali this PR alters source code so we should document the change, even though it isn't anticipated to affect Cocoa users. At least, that's my interpretation of the documented Platform Change Process.

PR #334, opened 2019-03-21
fix: Use exception stack when available

kattrali:
## Goal

Improve the debugging experience for caught `NSException`s by providing the exception stack rather than a stack generated at the time `notify()` is called.

## Design

If there is a stacktrace attached to an exception, uses the exception stack rather than generating a new one by passing `-[NSException callStackReturnAddresses]` as a `uintptr_t` array to `reportUserException`. If the exception has not been thrown or the call stack addresses array is null, a stacktrace is generated as before.

## Changeset

* Adding the new stack generation implementation
* Scoping `depth` argument to only affect generated stacktraces

## Tests

* Tested manually on a Mac app on macOS 10.13 and on an iOS app running iOS 12.1
* Added new integration test validating the stacktrace method contents

### Outstanding Questions

  - [x] Final review

### DIFF
--- a/Source/BugsnagCrashReport.h
+++ b/Source/BugsnagCrashReport.h
@@ -181,7 +181,8 @@ __deprecated_msg("Use toJson: instead.");
  */
 @property(readonly, copy, nonnull) NSDictionary *overrides;
 /**
- *  Number of frames to discard at the top of the stacktrace
+ *  Number of frames to discard at the top of the generated stacktrace.
+ *  Stacktraces from raised exceptions are unaffected.
  */
 @property(readwrite) NSUInteger depth;
 /**
--- a/Source/BugsnagNotifier.m
+++ b/Source/BugsnagNotifier.m
@@ -443,8 +443,10 @@ NSString *const kAppWillTerminate = @"App Will Terminate";
         [BugsnagHandledState handledStateWithSeverityReason:HandledError
                                                    severity:BSGSeverityWarning
                                                   attrValue:error.domain];
-    [self notify:NSStringFromClass([error class])
-             message:error.localizedDescription
+    NSException *wrapper = [NSException exceptionWithName:NSStringFromClass([error class])
+                                                   reason:error.localizedDescription
+                                                 userInfo:error.userInfo];
+    [self notify:wrapper
         handledState:state
                block:^(BugsnagCrashReport *_Nonnull report) {
                  NSMutableDictionary *metadata = [report.metaData mutableCopy];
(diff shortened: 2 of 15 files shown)


Source/BugsnagCrashReport.h
```diff
@@ -181,7 +181,8 @@ __deprecated_msg("Use toJson: instead.");
  */
 @property(readonly, copy, nonnull) NSDictionary *overrides;
 /**
- *  Number of frames to discard at the top of the stacktrace
+ *  Number of frames to discard at the top of the generated stacktrace.
+ *  Stacktraces from raised exceptions are unaffected.
```
fractalwrench: We should also update our docs so they state this.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
```diff
@@ -356,7 +357,17 @@ - (void)reportUserException:(NSString *)name
            terminateProgram:(BOOL)terminateProgram {
     const char *cName = [name cStringUsingEncoding:NSUTF8StringEncoding];
     const char *cReason = [reason cStringUsingEncoding:NSUTF8StringEncoding];
+    NSArray *addresses = [exception callStackReturnAddresses];
+    NSUInteger numFrames = [addresses count];
+    uintptr_t *callstack = malloc(numFrames * sizeof(*callstack));
+    for (NSUInteger i = 0; i < numFrames; i++) {
```
fractalwrench: How does this perform with very large stacktraces e.g. recursion gone wrong? The previous implementation has a limit of 100 frames.

Source/BugsnagNotifier.m
```diff
@@ -506,20 +502,14 @@ - (void)internalClientNotify:(NSException *_Nonnull)exception
...
 
-    [self notify:exception.name ?: NSStringFromClass([exception class])
-             message:exception.reason
-        handledState:state
-               block:^(BugsnagCrashReport *_Nonnull report) {
-                 if (block) {
-                     block(report);
-                 }
-               }];
+    [self notify:exception handledState:state block:block];
```
fractalwrench: Would this alter the depth of the stack, as there's no longer an extra block?

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
kattrali: Unlimited stack errors can’t be caught and notified. This implementation is more in line with uncaught NSException handling.

Source/BugsnagNotifier.m
kattrali: Yes, technically, however for how we use it, no. This function is only used in cocoa apps in react native, where a custom stacktrace is attached, disregarding depth altogether. And as it isn't a public interface, there should be no other effects.

PR #336, opened 2019-04-01
feat: Persist breadcrumbs on disk

kattrali:
## Goal

Persist breadcrumbs on disk to allow reading upon next boot in the event of an uncatchable app termination.

## Design

When adding breadcrumbs, this change serializes the collection to disk on the crumb read/write queue.

## Changeset

Added `cachePath` to `BugsnagBreadcrumbs`. This is the location used for serialization.

## Tests

Added new unit tests for the contents of crumbs on disk.

## Review

  - [x] Final review

### DIFF
--- a/Source/BugsnagBreadcrumb.m
+++ b/Source/BugsnagBreadcrumb.m
@@ -181,11 +181,18 @@ NSString *BSGBreadcrumbTypeValue(BSGBreadcrumbType type) {
 NSUInteger BreadcrumbsDefaultCapacity = 20;
 
 - (instancetype)init {
+    static NSString *const BSGBreadcrumbCacheFileName = @"bugsnag_breadcrumbs.json";
     if (self = [super init]) {
         _breadcrumbs = [NSMutableArray new];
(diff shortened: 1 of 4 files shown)


Source/BugsnagBreadcrumb.m
```diff
@@ -205,7 +212,19 @@ - (void)addBreadcrumbWithBlock:
...
+            [self.breadcrumbs addObject:crumb];
+            // Serialize crumbs to disk inside barrier to avoid simultaneous
+            // access to the file
+            if (self.cachePath != nil) {
+                static NSString *const arrayKeyPath = @"objectValue";
+                NSArray *items = [self.breadcrumbs valueForKeyPath:arrayKeyPath];
+                if ([NSJSONSerialization isValidJSONObject:items]) {
+                    NSData *data = [NSJSONSerialization dataWithJSONObject:items
+                                                                   options:0
+                                                                     error:nil];
```
fractalwrench: It would be good to check if an error occurs here and log if this is the case.

PR #348, opened 2019-05-17
Fix handled stacktrace generation

snmaynard:
We have a variable scope issue with stacktrace generation. This pulls the stack variables into the outer else block so they persist until the onCrash call.


snmaynard:
Tests for this case are being added on another branch - we can add them in here when they are done.

PR #349, opened 2019-05-21
fix(oom): Report short version as app.version

kattrali:
Out-of-memory reports are incorrectly reporting the build number as the version number. This change fixes that error and adds a few assertions for it in OOM report tests.

### DIFF
--- a/Source/BSGOutOfMemoryWatchdog.m
+++ b/Source/BSGOutOfMemoryWatchdog.m
@@ -146,6 +146,8 @@
         NSDictionary *lastBootInfo = [self readSentinelFile];
         if (lastBootInfo != nil) {
             self.lastBootCachedFileInfo = lastBootInfo;
+            NSString *lastBootBundleVersion =
+                [lastBootInfo valueForKeyPath:@"app.bundleVersion"];
             NSString *lastBootAppVersion =
                 [lastBootInfo valueForKeyPath:@"app.version"];
             NSString *lastBootOSVersion =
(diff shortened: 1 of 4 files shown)


Source/BSGOutOfMemoryWatchdog.m
```diff
@@ -214,7 +214,7 @@ - (NSMutableDictionary *)generateCacheInfoWithConfig:(BugsnagConfiguration *)con
     app[@"id"] = systemInfo[@BSG_KSSystemField_BundleID] ?: @"";
     app[@"name"] = systemInfo[@BSG_KSSystemField_BundleName] ?: @"";
     app[@"releaseStage"] = config.releaseStage;
-    app[@"version"] = systemInfo[@BSG_KSSystemField_BundleVersion] ?: @"";
+    app[@"version"] = systemInfo[@BSG_KSSystemField_BundleShortVersion] ?: @"";
```
CodeHex: Do we want to add the bundle version in as well? Think it would something like
`app[@"bundleVersion"] = systemInfo[@BSG_KSSystemField_BundleVersion] ?: @"";`

Source/BSGOutOfMemoryWatchdog.m
```diff
@@ -155,7 +155,7 @@ - (BOOL)computeDidOOMLastLaunchWithConfig:(BugsnagConfiguration *)config {
             NSString *osVersion = [BSG_KSSystemInfo osBuildVersion];
             NSDictionary *appInfo = [[NSBundle mainBundle] infoDictionary];
             NSString *appVersion =
-                [appInfo valueForKey:(__bridge NSString *)kCFBundleVersionKey];
+                [appInfo valueForKey:@BSG_KSSystemField_BundleShortVersion];
```
CodeHex: If this is being used to detect if the app has the same version and hence report the OOM, should this be on app version and bundle version rather than just app version?

Source/BSGOutOfMemoryWatchdog.m
kattrali: Yeah - having both makes sense 👍

Source/BSGOutOfMemoryWatchdog.m
kattrali: That's a good question. The safest thing is to check both, so I'll make it do that.

PR #361, opened 2019-06-10
ci: Update end-to-end tests environment

kattrali:
* Bumps the SDK version on CI
* Updates the wait time for crash reports
* Printing failing request payloads on CI (if any)
* Support Xcode 10.2 `simctl create` format

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,8 @@ matrix:
             - make test
             - make test OS=10.2
             - make test OS=9.2
-        - osx_image: xcode9.2
-          env: MAZE_SDK=11.2 SDK=iphonesimulator11.2
+        - osx_image: xcode10.1
+          env: MAZE_SDK=12.1 SDK=iphonesimulator12.1
           script: bundle exec maze-runner -c
 install: make bootstrap
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ matrix:
             - make test OS=9.2
         - osx_image: xcode10.1
           env: MAZE_SDK=12.1 SDK=iphonesimulator12.1
-          script: bundle exec maze-runner -c
+          script: bundle exec maze-runner -c || (cat maze_output/* && exit 1)
 install: make bootstrap
 
 before_deploy: headerdoc2html -o docs Source -j; gatherheaderdoc docs; mv docs/masterTOC.html docs/index.html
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -14,9 +14,9 @@ matrix:
         - osx_image: xcode9.4
           env: SDK=appletvsimulator11.4 BUILD_TV=1
           script: make test
-        - osx_image: xcode9.2
-          env: MAZE_SDK=11.2 SDK=iphonesimulator11.2
-          script: bundle exec maze-runner -c
+        - osx_image: xcode10.1
+          env: MAZE_SDK=12.1
+          script: bundle exec maze-runner -c || (cat maze_output/* && exit 1)
```
paulz: great idea to show maze output when build fails, if I understand correctly

features/scripts/launch_ios_simulators.sh
```diff
@@ -1,10 +1,18 @@
 #!/usr/bin/env bash
 
 INSTALL_PATH=build/Build/Products/Debug-iphonesimulator/iOSTestApp.app
-OS_VERSION=${MAZE_SDK:="11.2"}
+XCODE_VERSION=$(xcodebuild -version)
```
paulz: if I may propose a change to this file: https://github.com/bugsnag/bugsnag-cocoa/pull/362
this is PR to this branch
Runtime selection using full path `com.apple.CoreSimulator.SimRuntime.iOS-12_1` works for both Xcode 10.1 and 10.2.1, so we do not need to check for Xcode version. And selecting simulator by short name such as "iPhone 8" works too.

features/steps/crash_assertion_steps.rb
```diff
@@ -2,16 +2,27 @@
...
   stacktrace = exception["stacktrace"]
   assert_true(stacktrace.length > 0, "The stacktrace must have more than 0 elements")
 
   case stacktrace.first["method"]
-  when "__pthread_kill"
+  when "__pthread_kill" # Any
     assert_equal(exception["errorClass"], "SIGABRT")
     assert_equal(stacktrace[1]["method"], "abort")
-  when "_nc_table_find_64"
+  when "nanov2_allocate_from_block" # iOS 12.1
```
paulz: there are seems two cases for iOS 12.1, is it correct?
Can we case switch depending on the simulator runtime instead of first method?
Like `case MAZE_SDK` or `case iOS_version`

kattrali: Yeah, that sounds good 👍

PR #364, opened 2019-06-11
Fix acceptance tests running on Mojave

paulz:
## Goal

Verify stack trace contains list of methods (in order)

Fixes failure on when running on Mojave when simulator runtime contains an extra method
`_platform_strlcpy` in addition to expected `pthread_getname_np` and `-[AsyncSafeThreadScenario run]`

## Tests

Locally on Mojave and on High Sierra on travis

### Outstanding Questions

Can new step be reused somewhere else?

  - [X] Structural review of the classes, functions, and properties modified
  - [X] Final review


snmaynard:
CI has failed so we should address that - I havent looked into it yet.

PR #365, opened 2019-06-12
fix(codec): Parse numbers in arrays correctly

kattrali:
## Goal

Fix parsing logic for number values within arrays

## Design

Previously, parsing numbers from JSON arrays exited prematurely if the "name" value was empty. However, name is always empty when parsing arrays, and is only used when parsing JSON objects. 

So instead of exiting early to prevent parsing an invalid name, pass nil along as the name value.

## Tests

* Added a new test for adding metadata array
* Enabled some unused (??) tests for JSON parsing

## Review

  - [x] Final review

### DIFF
--- a/Tests/KSCrash/KSJSONCodec_Tests.m
+++ b/Tests/KSCrash/KSJSONCodec_Tests.m
@@ -210,24 +210,24 @@ static NSString* toString(NSData* data)
     XCTAssertEqualObjects(result, original, @"");
 }
 
-//- (void)testSerializeDeserializeArrayMultipleEntries
-//{
-//    NSError* error = (NSError*)self;
-//    NSString* expected = @"[\"One\",1000,true]";
-//    id original = @[@"One",
-//            @1000,
-//            @YES];
-//    NSString* jsonString = toString([BSG_KSJSONCodec encode:original
-//                                                options:BSG_KSJSONEncodeOptionSorted
-//                                                  error:&error]);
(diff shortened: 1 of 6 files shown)


Tests/KSCrash/KSJSONCodec_Tests.m
```diff
@@ -210,24 +210,24 @@ - (void)testSerializeDeserializeArrayStringIntl
...
-//                                                  error:&error]);
-//    XCTAssertNotNil(jsonString, @"");
-//    XCTAssertNil(error, @"");
-//    XCTAssertEqualObjects(jsonString, expected, @"");
-//    id result = [BSG_KSJSONCodec decode:toData(jsonString) options:0 error:&error];
-//    XCTAssertNotNil(result, @"");
-//    XCTAssertNil(error, @"");
-//    XCTAssertEqualObjects(result, original, @"");
-//}
+- (void)testSerializeDeserializeArrayMultipleEntries
```
paulz: Awesome job restoring commented out tests!

features/handled_errors.feature
```diff
@@ -1,17 +1,20 @@
 Feature: Handled Errors and Exceptions
 
-Scenario: Override errorClass and message from a notifyError() callback and discard lines from stack
+Scenario: Override errorClass and message from a notifyError() callback, customize report
```
paulz: proposing adding the comment to the scenario about metadata, please adjust wording as necessary.

PR #367, opened 2019-06-17
Fix reporting disk free space

paulz:
## Goal

Skip reporting free space when getting file system attributes fails

## Changeset

Also disable debugger in test as it prevents tests from running in Xcode.

## Tests

Unit test for iOS and tvOS using failure and macOS using success paths.

### Outstanding Questions

  - [X] Final review

  features)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,9 @@ matrix:
         - osx_image: xcode9.4
           env: PLATFORM=iOS
           script:
-            - fold_start(); make test; fold_end()
-            - fold_start(); make test OS=10.3.1; fold_end()
-            - fold_start(); make test OS=9.3; fold_end()
+            - fold_start; make test; fold_end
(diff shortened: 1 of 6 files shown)


kattrali:
@paulz Can you explain a bit more about what this change fixes, and why?

paulz:
> @paulz Can you explain a bit more about what this change fixes, and why?

Yes, here https://github.com/bugsnag/bugsnag-cocoa/pull/367/files#diff-511ffb2809f48997203cfba9e0630b9dR47 we no longer setting freeDisk space when error is not nil

paulz:
Using the UserHome directory is more reliable as Documents are not accessible sometimes, see test failures on travis:  https://travis-ci.com/bugsnag/bugsnag-cocoa/jobs/208117433#L1456
as Documents directory is was available on iOS 10. To makes test pass more reliably for all versions I chose another base folder UserHome.

PR #370, opened 2019-06-21
Fix internal notify() unhandled count

kattrali:
## Goal

Fix incrementing unhandled counts when using internal notify() API. This resolves discrepancies in stability scores for users of bugsnag-react-native after receiving unhandled JavaScript events.

## Changeset

* Added a helper function to `BugsnagSessionTracker` for incrementing unhandled counts
* Cached the session `unhandledCount` in `BugsnagNotifier` as the value can be more than 1 now
* Read unhandled value passed through `internalClientNotify` to determine event handledness
* Updated `notify:handledState:block:` to respect handledness and increment either the handled or unhandled event counts as needed

## Tests

* Added new end-to-end tests using the `internalClientNotify()` function to send handled and unhandled events after a session and verifying the session event counts

This pull request is ready for final review.

### DIFF
--- a/Source/BugsnagSessionTracker.m
+++ b/Source/BugsnagSessionTracker.m
@@ -165,7 +165,23 @@ NSString *const BSGSessionUpdateNotification = @"BugsnagSessionChanged";
 
     @synchronized (session) {
         session.handledCount++;
-        if (self.callback && (self.config.shouldAutoCaptureSessions || !session.autoCaptured)) {
+        if (self.callback) {
+            self.callback(session);
+        }
+        [self postUpdateNotice];
+    }
+}
+
+- (void)handleUnhandledErrorEvent {
+    BugsnagSession *session = [self runningSession];
+
+    if (session == nil) {
+        return;
+    }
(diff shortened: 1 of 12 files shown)


Source/BugsnagSessionTracker.m
```diff
@@ -172,4 +172,20 @@ - (void)handleHandledErrorEvent {
...
+- (void)handleUnhandledErrorEvent {
+    BugsnagSession *session = [self runningSession];
+
+    if (session == nil) {
+        return;
+    }
+
+    @synchronized (session) {
+        session.unhandledCount++;
+        if (self.callback && (self.config.shouldAutoCaptureSessions || !session.autoCaptured)) {
```
paulz: is this logic correct? can this be extracted into a method? was hoping this could be simple:
```
if (self.callback) {
    self.callback(session);
}
```

kattrali: yeah, though it converts into a no-op. I simplified the logic instead!

PR #374, opened 2019-06-27
Fix reporting disk free space

paulz:
## Goal

Skip reporting free space when getting file system attributes fails.

Also, add Bugsnag tests to the tvOS test suite.

## Tests

Unit test for success and failure cases.

### Outstanding Questions

  - [ ] Final review


iOS/BugsnagTests/BugsnagKSCrashSysInfoParserTest.m
```diff
@@ -6,9 +6,10 @@
 //  Copyright © 2018 Bugsnag. All rights reserved.
 //
 
-#import <XCTest/XCTest.h>
+@import XCTest;
```
fractalwrench: Nice to see some modern syntax creeping in :)

PR #380, opened 2019-07-11
fix(oom): Disable out-of-memory reporting in debug mode

kattrali:
## Goal

Remove erroneous out-of-memory reports during development if reporting is enabled in debug mode. Since running the app via Xcode/simctl terminates the existing app launch session with a kill signal, no cleanup can be performed and the app state would indicate that a possible OOM may have occurred.

## Changeset

* Disables `reportOOMs` if in `DEBUG` mode.

## Tests

* Tested manually and by updating the end-to-end tests

### DIFF
--- a/Tests/BugsnagConfigurationTests.m
+++ b/Tests/BugsnagConfigurationTests.m
@@ -55,6 +55,15 @@
     XCTAssertFalse([config shouldSendReports]);
 }
 
+- (void)testDefaultReleaseStage {
+    BugsnagConfiguration *config = [BugsnagConfiguration new];
+#if DEBUG
+    XCTAssertEqualObjects(@"development", config.releaseStage);
+#else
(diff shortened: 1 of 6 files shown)


Tests/BugsnagConfigurationTests.m
```diff
@@ -113,6 +126,7 @@ - (void)testSetEndpoints {
     XCTAssertEqualObjects([NSURL URLWithString:@"http://sessions.example.com"], config.sessionURL);
 }
 
+#if DEBUG
```
tomlongridge: Is this only applicable in debug mode? Having trouble seeing why.

kattrali: Yeah - I'll add a bit more context to the block but `NSAssert()`-based behavior is optimized out in release mode and no exception would be thrown.

PR #381, opened 2019-07-11
refactor: Remove unused zombie detection

kattrali:
## Goal

Remove unused code, reducing the library maintenance surface

## Changeset

Removes implementation and tests relating to zombie object detection, which until this point was entirely disabled.

## Tests

* Tested manually on a device that reporting behaved the same


kattrali:
Yeah - it was a global rename. I think I'll look at pruning it down altogether.

PR #382, opened 2019-07-11
fix: Improve parsing for crash reports by using builtin JSON impl

kattrali:
## Goal

While testing report parsing, I found a few cases where the implementation of JSON parsing differed from [RFC 8259](https://tools.ietf.org/html/rfc8259), specifically around handling escaping for control characters, exponents in number values (and large numbers generally), and string character values in the form `'u' hex hex hex hex`. This change adds a few new tests for these cases and replaces the JSON decoder logic with the builtin `NSJSONSerialization`.

## Changeset

* Changed `BSG_KSCrashState` from C to ObjC to use the ObjC interface to JSON parsing. This change uses a more manual (and less clever) parsing of the state object rather than loading the struct from the JSON file directly.
* Removed the JSON parsing logic in `BSG_KSJSONCodec`
* Changed the implementation of `+[BSG_KSJSONCodecObjC decode:options:error:]` to use `NSJSONSerialization`, making the tests pass. This change also ignores the customization `options` parameters, which now permits `NSNull` values in payloads.

## Tests

* Tested report parsing and state detection continues to work as expected, particularly `didCrashLastLaunch`


Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashState.m
```diff
@@ -196,16 +197,29 @@ bool bsg_kscrashstate_i_loadState(BSG_KSCrash_State *const context,
...
-        BSG_KSLOG_ERROR("%s, offset %d: %s", path, errorOffset,
-                        bsg_ksjsonstringForError(result));
+    if (error != nil) {
+        BSG_KSLOG_ERROR(@"%s, %@", path, error);
         return false;
     }
+
+    context->activeDurationSinceLastCrash = [objectContext[@"activeDurationSinceLastCrash"] doubleValue];
+    context->activeDurationSinceLaunch = [objectContext[@"activeDurationSinceLaunch"] doubleValue];
+    context->appLaunchTime = [objectContext[@"activeDurationSinceLaunch"] unsignedLongLongValue];
```
tomlongridge: Incorrect field assignment - "appLaunchTime"?

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSJSONCodecObjC.m
```diff
@@ -504,28 +504,14 @@ + (NSData *)encode:(id)object
...
-
-    if (result != BSG_KSJSON_OK &&
-        !(decodeOptions & BSG_KSJSONDecodeOptionKeepPartialObject)) {
-        return nil;
+    *error = nil;
+    id result = nil;
+    @try {
+        result = [NSJSONSerialization JSONObjectWithData:JSONData options:0 error:error];
+    } @catch (NSException *exception) {
+        result = @{};
```
tomlongridge: No error handling required here?

Tests/KSCrash/KSJSONCodec_Tests.m
```diff
@@ -1370,37 +1385,13 @@ - (void) testDeserializeObjectWithNullIgnoreNullInArray
...
-}
-
 - (void) testFloatParsingDoesntOverflow
 {
     NSError *error = (NSError*)self;
 
     char * buffer = malloc(0x1000000);
     for (int i = 0; i < 0x1000000; i++) {
-        buffer[i] = ';';
+        buffer[i] = ' ';
```
tomlongridge: Is this related to the code changes? Just intrigued...!

kattrali: yeah - this wasn't a reasonable test before since having endless semicolons at the end of the file wasn't valid. Using some whitespace keeps the intent (ensuring a float value doesn't overflow into later characters) without causing other parsing errors.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSJSONCodecObjC.m
kattrali: Nope! If the `try` succeeds, it returns a dict. In the event of failure, this `catch` falls back to an empty map. It also fixes the edge case where JSON parsing failed and no report was generated.

PR #383, opened 2019-07-15
feat: Support loading config options from Info.plist

kattrali:
## Goal

Read config options from Info.plist values, supporting workflows where options are specified directly or added to script build phases.

## Design

* `BugsnagConfiguration` adds a new utility constructor, `loadConfig`, which reads the `infoDictionary` values under `bugsnag` within the main bundle.
* Options are then passed to the config builder which constructs a configuration if the options are all valid. Invalid configurations are discarded to avoid cases where endpoints or other critical options are mistyped and reports sent to an incorrect server.


Source/BSGConfigurationBuilder.m
```diff
@@ -0,0 +1,145 @@
...
+ * type or name, to avoid data leakage to the hosted version from on-premise.
+ */
+BugsnagConfiguration *BSGValidateAndSetEndpoints(BugsnagConfiguration *config, NSMutableDictionary *options);
+
+@implementation BSGConfigurationBuilder
+
++ (BugsnagConfiguration *)configurationFromOptions:(NSDictionary *)options {
+    BugsnagConfiguration *config = [BugsnagConfiguration new];
+    NSMutableDictionary *properties = [options mutableCopy];
+    config = BSGValidateAndSetStringOption(config, properties, NSStringFromSelector(@selector(apiKey)));
```
fractalwrench: Why is `NSStringFromSelector(@selector(apiKey))` required rather than `@"apiKey"`?

Source/BSGConfigurationBuilder.m
```diff
@@ -0,0 +1,145 @@
...
+    }
+    config = BSGValidateAndSetBooleanOption(config, properties, NSStringFromSelector(@selector(autoNotify)));
+    config = BSGValidateAndSetBooleanOption(config, properties, BSGAutoCollectBreadcrumbsKey);
+    config = BSGValidateAndSetBooleanOption(config, properties, BSGAutoCaptureSessionsKey);
+    config = BSGValidateAndSetBooleanOption(config, properties, NSStringFromSelector(@selector(reportOOMs)));
+    config = BSGValidateAndSetBooleanOption(config, properties, NSStringFromSelector(@selector(reportBackgroundOOMs)));
+    config = BSGValidateAndSetStringOption(config, properties, NSStringFromSelector(@selector(releaseStage)));
+    config = BSGValidateAndSetNotifyReleaseStages(config, properties);
+    config = BSGValidateAndSetEndpoints(config, properties);
+    if (properties.count > 0) {
```
fractalwrench: Just ignoring these seems preferable to failing to initialise entirely. This also means that properties can be an `NSDictionary`, and none of the calls would need to mutate the dict's contents.

Source/BSGConfigurationBuilder.m
```diff
@@ -0,0 +1,145 @@
...
+}
+
+@end
+
+static BOOL BSGValueIsBoolean(id object) {
+    return [object isKindOfClass:[NSNumber class]]
+    && CFGetTypeID((__bridge CFTypeRef)object) == CFBooleanGetTypeID();
+}
+
+NSString *BSGTransformOptionToPropertyName(NSString *option) {
```
fractalwrench: I'm not clear on the purpose of this method

Tests/BSGConfigurationBuilderTests.m
```diff
@@ -0,0 +1,494 @@
...
+    XCTAssertNil(config);
+}
+
+- (void)testDecodeApiKey {
+    BugsnagConfiguration *config = [BSGConfigurationBuilder
+        configurationFromOptions:@{@"apiKey" : @"foo123"}];
+    XCTAssertEqualObjects(@"foo123", config.apiKey);
+}
+
+- (void)testDecodeDefaultAutoNotify {
```
fractalwrench: These tests could be probably be condensed into one method as they're all testing the default values - or alternatively, the check for an API key could be removed from all but one of the `testDecodeDefault*` methods.

Tests/BSGConfigurationBuilderTests.m
```diff
@@ -0,0 +1,494 @@
...
+    }];
+    XCTAssertNil(config);
+}
+
+- (void)testDecodeEndpointsUnknownKeys {
+    BugsnagConfiguration *config = [BSGConfigurationBuilder configurationFromOptions:@{
+        @"endpoints" : @{
+            @"notify" : @"https://notify.example.com",
+            @"sessions" : @"https://sessions.example.com",
+            @"florps" : @"https://florps.example.com",
```
fractalwrench: We should just ignore invalid keys, as long as any required information (i.e. notify + sessions) is there

Tests/BSGConfigurationBuilderTests.m
```diff
@@ -0,0 +1,494 @@
...
+    BugsnagConfiguration *config = [BSGConfigurationBuilder configurationFromOptions:@{
+        @"endpoints" : @{
+            @"notify" : @"https://notify.example.com",
+            @"sessions" : @"https://sessions.example.com"
+        },
+    }];
+    XCTAssertNil(config);
+}
+
+- (void)testDecodeEndpointsOnlyNotifySet {
```
fractalwrench: This is missing an API key from the supplied dict

Tests/BSGConfigurationBuilderTests.m
```diff
@@ -0,0 +1,494 @@
...
+    BugsnagConfiguration *config = [BSGConfigurationBuilder configurationFromOptions:@{
+        @"endpoints" : @{
+            @"notify" : @"https://notify.example.com",
+        },
+    }];
+    XCTAssertNil(config);
+}
+
+- (void)testDecodeEndpointsOnlySessionsSet {
```
fractalwrench: This is missing an API key from the supplied dict

Tests/BSGConfigurationBuilderTests.m
```diff
@@ -0,0 +1,494 @@
...
+            @"apiKey" : @"b128acce"
+        }];
+    XCTAssertNotNil(config);
+    XCTAssertEqualObjects(@"b128acce", config.apiKey);
+    XCTAssertEqual(2, config.notifyReleaseStages.count);
+    XCTAssertTrue([config.notifyReleaseStages containsObject:@"beta"]);
+    XCTAssertTrue([config.notifyReleaseStages containsObject:@"prod"]);
+}
+
+- (void)testDecodeReportOOMsWithoutApiKey {
```
fractalwrench: It seems like there only needs to be 1 test case in this suite to check what happens if an API key isn't present

Source/BugsnagConfiguration.m
```diff
@@ -53,6 +54,15 @@ @interface BugsnagConfiguration ()
 
 @implementation BugsnagConfiguration
 
++ (instancetype _Nullable)loadConfig {
```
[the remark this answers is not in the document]
kattrali: Throwing exceptions are not a standard control flow mechanism in Cocoa - the program can continue running without this method succeeding. However, it could return more info about what failed, so I'll update it to do that.

> In many environments, use of exceptions is fairly commonplace. For example, you might throw an exception to signal that a routine could not execute normally—such as when a file is missing or data could not be parsed correctly. Exceptions are resource-intensive in Objective-C. You should not use exceptions for general flow-control, or simply to signify errors.

[Documentation Archive - Exception Handling](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjectiveC/Chapters/ocExceptionHandling.html)

Tests/BSGConfigurationBuilderTests.m
kattrali: Keeping the individual tests as narrow as possible makes it easy to see which part of the implementation is failing. In lieu of a complete suite of every possible permutation, these cover the critical paths of setting 1-2 fields (including API key) and ensuring correct behavior.

Smaller tests can also produce better results when enabling test execution randomization.

robinmacharg:
This has been implemented in #582.  Closing.

PR #394, opened 2019-07-22
fix(oom): Improve background app state detection

kattrali:
## Goal

Correctly determine foreground/active state when the app is launched via a background fetch or media playback event or extension, or is interrupted by Siri or a phone call. These cases should be regarded as in the background or inactive respectively.

## Changeset

* Added detection for app extension state
* Added initial app state (foreground and active) checking

## Tests

* Added additional end-to-end tests for when the app is interrupted and will terminate normally (for completeness)

### DIFF
--- a/Source/BSGOutOfMemoryWatchdog.m
+++ b/Source/BSGOutOfMemoryWatchdog.m
@@ -64,6 +64,14 @@
                    name:UIApplicationWillEnterForegroundNotification
                  object:nil];
     [center addObserver:self
+               selector:@selector(handleTransitionToActive:)
+                   name:UIApplicationDidBecomeActiveNotification
+                 object:nil];
+    [center addObserver:self
+               selector:@selector(handleTransitionToInactive:)
+                   name:UIApplicationWillResignActiveNotification
+                 object:nil];
+    [center addObserver:self
                selector:@selector(handleLowMemoryChange:)
                    name:UIApplicationDidReceiveMemoryWarningNotification
                  object:nil];
@@ -114,6 +122,17 @@
     self.cachedFileInfo[@"app"][@"releaseStage"] = change[NSKeyValueChangeNewKey];
     [self writeSentinelFile];
 }
+
+- (void)handleTransitionToActive:(NSNotification *)note {
+    self.cachedFileInfo[@"app"][@"isActive"] = @YES;
+    [self writeSentinelFile];
+}
+
+- (void)handleTransitionToInactive:(NSNotification *)note {
+    self.cachedFileInfo[@"app"][@"isActive"] = @NO;
+    [self writeSentinelFile];
+}
+
 - (void)handleTransitionToForeground:(NSNotification *)note {
     self.cachedFileInfo[@"app"][@"inForeground"] = @YES;
     [self writeSentinelFile];
@@ -154,6 +173,8 @@
                 [lastBootInfo valueForKeyPath:@"device.osBuild"];
             BOOL lastBootInForeground =
                 [[lastBootInfo valueForKeyPath:@"app.inForeground"] boolValue];
+            BOOL lastBootWasActive =
+                [[lastBootInfo valueForKeyPath:@"app.isActive"] boolValue];
             NSString *osVersion = [BSG_KSSystemInfo osBuildVersion];
             NSDictionary *appInfo = [[NSBundle mainBundle] infoDictionary];
             NSString *bundleVersion =
@@ -163,7 +184,8 @@
             BOOL sameVersions = [lastBootOSVersion isEqualToString:osVersion] &&
                                 [lastBootBundleVersion isEqualToString:bundleVersion] &&
                                 [lastBootAppVersion isEqualToString:appVersion];
-            BOOL shouldReport = config.reportOOMs && (config.reportBackgroundOOMs || lastBootInForeground);
+            BOOL shouldReport = config.reportOOMs
+                && (config.reportBackgroundOOMs || (lastBootInForeground && lastBootWasActive));
             [self deleteSentinelFile];
             return sameVersions && shouldReport;
         }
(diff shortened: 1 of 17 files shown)


features/out_of_memory.feature
```diff
@@ -20,6 +20,29 @@ Feature: Reporting out of memory events
...
+        Then I should receive 0 requests
+
+    Scenario: The app is terminated normally
+        The application can be gracefully terminated by the OS if more
+        memory is needed for other applications or directly by calling
+        exit(0)
+
+        When I run "OOMWillTerminateScenario"
+        And I wait for 4 seconds
+        And I relaunch the app
```
fractalwrench: I might be reading this wrong, but it appears as though the `relaunch the app` step in mazerunner launches the fixture with the same `EVENT_TYPE`. If that's true, it would mean that in this scenario the app always terminates and zero requests would be expected regardless of OOM detection.

features/out_of_memory.feature
```diff
@@ -20,6 +20,29 @@ Feature: Reporting out of memory events
         And the event "app.bundleVersion" equals "5"
         And the event breadcrumbs contain "Crumb left before crash"
 
+    @ci_skip
+    Scenario: The OS kills the application while inactive
+        The application is in the foreground but inactive when interrupted 
+        by a phone call or Siri
+
+        When I run "OOMScenario"
+        And the app is interrupted by Siri
```
fractalwrench: This step doesn't work on my machine, and the scenario fails with 1 request being received.

Source/BSGOutOfMemoryWatchdog.m
```diff
@@ -114,6 +122,17 @@ - (void)observeValueForKeyPath:(NSString *)keyPath
     self.cachedFileInfo[@"app"][@"releaseStage"] = change[NSKeyValueChangeNewKey];
     [self writeSentinelFile];
 }
+
+- (void)handleTransitionToActive:(NSNotification *)note {
+    self.cachedFileInfo[@"app"][@"isActive"] = @YES;
+    [self writeSentinelFile];
```
fractalwrench: Can this IO be moved to a background thread?

Source/BSGOutOfMemoryWatchdog.m
```diff
@@ -246,4 +285,17 @@ - (NSMutableDictionary *)generateCacheInfoWithConfig:(BugsnagConfiguration *)con
     return cache;
 }
 
+// Only available on iOS/tvOS
+#if BSGOOMAvailable
+- (UIApplicationState)currentAppState {
+    // Only checked outside of app extensions since sharedApplication is
+    // unavailable to extension UIKit APIs
+    if ([BSG_KSSystemInfo isRunningInAppExtension]) {
+        return YES;
```
fractalwrench: This should return `UIApplicationState` rather than a `BOOL`

Source/BSGOutOfMemoryWatchdog.m
```diff
@@ -246,4 +285,17 @@ - (NSMutableDictionary *)generateCacheInfoWithConfig:(BugsnagConfiguration *)con
...
+    }
+    UIApplication *app = [UIApplication performSelector:@selector(sharedApplication)];
```
fractalwrench: Is there a chance this call could fail if the selector needs to be specified like this?

Source/BSGOutOfMemoryWatchdog.m
```diff
@@ -221,7 +243,24 @@ - (NSMutableDictionary *)generateCacheInfoWithConfig:(BugsnagConfiguration *)con
...
+    // "inactive". From the UIApplicationState docs:
+    // > UIApplicationStateActive
+    // >   The app is running in the foreground and currently receiving events.
+    // > UIApplicationStateInactive
+    // >   The app is running in the foreground but is not receiving events.
+    // >   This might happen as a result of an interruption or because the app
+    // >   is transitioning to or from the background.
+    // > UIApplicationStateBackground
+    // >   The app is running in the background.
+    app[@"inForeground"] = @(state == UIApplicationStateInactive
```
fractalwrench: Is `app.inForeground` reported differently for OOM events vs regular error reports? It'd be good to try and address any discrepancy in how the value is reported if feasible, as otherwise this could lead to confusion.

Source/BugsnagBreadcrumb.h
```diff
@@ -132,6 +132,6 @@ typedef void (^BSGBreadcrumbConfiguration)(BugsnagBreadcrumb *_Nonnull);
 /**
  * Reads and return breadcrumb data currently stored on disk
  */
-- (NSDictionary *_Nullable)cachedBreadcrumbs;
+- (NSArray *_Nullable)cachedBreadcrumbs;
```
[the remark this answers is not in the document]
kattrali: It’s not breaking because it never returned a dict in the first place.

features/out_of_memory.feature
kattrali: Took a couple of tries but figured it out - this scenario only works if the simulator is not running headless. I can tweak it  ¯\_(ツ)_/¯

Source/BSGOutOfMemoryWatchdog.m
kattrali: It’s already on a background thread to actually write the file, the thread is managed within the file writing function.

Source/BSGOutOfMemoryWatchdog.m
kattrali: No. This is to avoid a check by the compiler that applicationState is not called from app extensions, which is already accounted for. There is no compile-time means of insuring that is the case.

Source/BSGOutOfMemoryWatchdog.m
kattrali: Good idea. Raised a ticket to check if other events are reporting foreground state correctly at launch. This one specifically is isolated since it manages its own state.

Source/BugsnagBreadcrumb.h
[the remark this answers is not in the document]
fractalwrench: Ok, missed that the first time round. Should this property be public at all?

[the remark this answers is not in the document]
kattrali: Nope, that wasn't the intent at all.

PR #395, opened 2019-07-22
Fixing malloc without free causing memory leak

caroaguilar:
Fix memory leak happening every time a user exception is reported. We found this while using the `bugsnag-react-native` SDK and profiling the app with Xcode Instruments. From there we were able to track this bug down to `bugsnag-cocoa` SDK. 

Screen Shot 2019-07-22 at 3 30 08 PM

By adding the missing `free()` to the SDK locally we see the leak fixed and not showing in Instruments.

The `malloc` in question is found here https://github.com/bugsnag/bugsnag-cocoa/blob/master/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m#L317

Since `bsg_kscrash_reportUserException` from above file could or could not return, I thought it was better to free the resource in `BSG_KSCrashSentry_User.c` which is final in call hierarchy. 

## Changeset

Adding the respectively `free()` for the `malloc`

## Review

  - [x] Final review

### DIFF
--- a/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
+++ b/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
@@ -370,6 +370,8 @@ IMPLEMENT_EXCLUSIVE_SHARED_INSTANCE(BSG_KSCrash)
                                     depth,
                                     terminateProgram);
 
+    free(callstack);
+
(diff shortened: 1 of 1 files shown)


caroaguilar:
Hi @kattrali, thank you for your quick response and review!

Makes total sense, already applied the changes requested. 

Thanks :)

PR #406, opened 2019-08-23
Move end-to-end tests to Buildkite

Cawllec:
## Goal

Move the end-to-end tests to buildkite allowing us to test against real devices, while retaining the existing tests for running locally against local emulators.  There is additional work such as combining scenario sets into a shared library that can be done as a later task.

## Changes

- Remove e2e tests from `.travis.yml`
- Copied current tests into `end-to-end-tests` directory to separate from existing maze features
- Added buildkite pipeline definition for running tests
- Updated features to match Maze-runner v2 changes
- Added OOM tests
- Added shared target for both test sets
- Move testing steps from CONTRIBUTING.md to TESTING.md
- Add easy way to run tests locally using Makefile

## TODO
- Add symbolication and re-enable crashprobe tests


fractalwrench:
Just a quick note that this seems to duplicate all the scenarios on the filesystem, with one copy in `features/fixtures` and another in `end-to-end-tests`.

Duplication of the scenario code on Android has been painful when updating/adding new tests, as it becomes quite easy to forget to make changes in one file, leading to inconsistency when running locally vs on browserstack.

Unless it's absolutely necessary, duplication of scenario code should be avoided. It should be possible to edit the `xcodeproj` of both fixtures to share the same source files, happy to go into more detail if that'd help

Cawllec:
I'll need to look into the shared library, it shouldn't be too big a task as I've made relatively minimal changes to the actual scenario files.  I'll add it as a todo and take a look.

kattrali:
Can you make a list of which scenarios actually changed? We might not need even a shared library, just one project that runs different schemes or scenarios or something for devices vs simulators.

twometresteve:
Overtaken by numerous other PRs.

PR #424, opened 2019-10-11
Deliver each event in a separate request to avoid exceeding payload size limit

fractalwrench:
## Goal

The payload size limit can be exceeded if multiple events are batched together into one request, meaning events are dropped by our API. This changeset removes the batching functionality and sends each event in a separate request to reduce the likelihood of this occurring.

## Changeset

- Sent each event in a separate request within `sendAllReportsWithCompletion`.
- Logged the filename when a request succeeds/fails
- Passed through 1 event at a time to `BugsnagApiClient` for errors (sessions are still batched together in multiple payloads)
- Changed completion callback to send the filename rather than the count of reports sent

## Tests

- Stored multiple events on disk, then launched the notifier configured to point at a local endpoint and confirmed that each request was received individually.
- Updated mazerunner scenario assertions where they assumed requests were batched


fractalwrench:
Closing as stale.

PR #433, opened 2019-11-05
Rename breadcrumb 'name' to 'message' removing length limit

kattrali:
## Goal

1. Rename 'name' to 'message'
2. Remove 4kb length limit on individual breadcrumbs


fractalwrench:
This should be closed as the notifier spec no longer mandates that breadcrumbs should be trimmed on size alone, as users have found this confusing.

PR #435, opened 2019-11-07
Lay out structure for property/method renames for v6.0

kattrali:
## Goal

This changeset lays out an initial structure for the updates necessary to get the library ready for v6.0.0. The primary change in this major revision will be to rename some properties and methods to better align with other Bugsnag notifier libraries. These changes have been added in a backwards-compatible way to allow for incremental releases leading up to v6, when all deprecated properties will be removed. 

Every change should be listed in the changelog and upgrade migration guide, and the new properties should replace the existing ones in the integration guide on docs.bugsnag.com.

### DIFF
--- a/Source/Bugsnag.h
+++ b/Source/Bugsnag.h
@@ -212,7 +212,7 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
  * By default, sessions are automatically started when the application enters the foreground.
  * If you wish to manually call startSession at
  * the appropriate time in your application instead, the default behaviour can be disabled via
- * shouldAutoCaptureSessions.
+ * autoTrackSessions.
  *
  * Any errors which occur in an active session count towards your application's
  * stability score. You can prevent errors from counting towards your stability
@@ -230,7 +230,7 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
  * When a session is stopped, errors will not count towards your application's
  * stability score. This can be advantageous if you do not wish these calculations to
  * include a certain type of error, for example, a crash in a background service.
- * You should disable automatic session tracking via shouldAutoCaptureSessions if you call this method.
+ * You should disable automatic session tracking via autoTrackSessions if you call this method.
  *
  * A stopped session can be resumed by calling resumeSession,
  * which will make any subsequent errors count towards your application's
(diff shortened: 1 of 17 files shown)


Source/BugsnagSessionTracker.m
```diff
@@ -91,7 +91,7 @@ - (BugsnagSession *)runningSession {
 }
 
 - (void)startNewSessionIfAutoCaptureEnabled {
-    if (self.config.shouldAutoCaptureSessions  && [self.config shouldSendReports]) {
+    if (self.config.autoDetectSessions && [self.config shouldSendReports]) {
```
fractalwrench: I can't find the exact section in the spec but believe one of the points of discussion is that options like `shouldSendReports` should no longer affect sessions. This assumes that `shouldSendReports` has similar semantics to `autoDetectErrors` - is this the case? We may just want to remove the option entirely.

Source/Bugsnag.h
```diff
@@ -200,7 +200,8 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
  *
  * @param capacity max number of breadcrumb log messages to send
  */
-+ (void)setBreadcrumbCapacity:(NSUInteger)capacity;
++ (void)setMaxBreadcrumbs:(NSUInteger)capacity;
```
fractalwrench: This is not on the Bugsnag interface any more according to the spec, and should be on Configuration instead.

Source/Bugsnag.m
```diff
@@ -215,12 +215,16 @@ + (void)leaveBreadcrumbForNotificationName:
     }
 }
 
-+ (void)setBreadcrumbCapacity:(NSUInteger)capacity {
++ (void)setMaxBreadcrumbs:(NSUInteger)capacity {
     if ([self bugsnagStarted]) {
         self.notifier.configuration.breadcrumbs.capacity = capacity;
```
fractalwrench: The default capacity is now 25 according to the spec, with bounds of 0-100 - this should be enforced.

UPGRADING.md
```diff
@@ -0,0 +1,29 @@
+# Upgrading
+
+Guide to ease migrations between significant changes
+
+## v5 -> v6
+
+Version 6 introduces a number of property and method renames:
```
tomlongridge: Does it matter if this is included in master before the v6 release?

kattrali: Not particularly, though we could cut some beta releases.

PR #438, opened 2019-12-03
feat: (internal) Add internal plugin interface for custom behavi…

kattrali:
Adds a private capability to configure library behavior prior to capturing and delivering reports.

### DIFF
--- a/Source/Bugsnag.h
+++ b/Source/Bugsnag.h
@@ -27,6 +27,7 @@
 
 #import "BugsnagConfiguration.h"
 #import "BugsnagMetaData.h"
+#import "BugsnagPlugin.h"
 
 static NSString *_Nonnull const BugsnagSeverityError = @"error";
 static NSString *_Nonnull const BugsnagSeverityWarning = @"warning";
--- a/Source/Bugsnag.m
+++ b/Source/Bugsnag.m
@@ -29,12 +29,15 @@
 #import "BugsnagLogger.h"
 #import "BugsnagNotifier.h"
 #import "BugsnagKeys.h"
+#import "BugsnagPlugin.h"
 
 static BugsnagNotifier *bsg_g_bugsnag_notifier = NULL;
+static NSMutableArray <id<BugsnagPlugin>> *registeredPlugins;
 
 @interface Bugsnag ()
 + (BugsnagNotifier *)notifier;
 + (BOOL)bugsnagStarted;
++ (void)registerPlugin:(id<BugsnagPlugin>)plugin;
 @end
 
 @interface NSDictionary (BSGKSMerge)
@@ -54,6 +57,7 @@ static BugsnagNotifier *bsg_g_bugsnag_notifier = NULL;
         if ([configuration hasValidApiKey]) {
             bsg_g_bugsnag_notifier =
                     [[BugsnagNotifier alloc] initWithConfiguration:configuration];
+            [self startPlugins];
             [bsg_g_bugsnag_notifier start];
         } else {
             bsg_log_err(@"Bugsnag not initialized - a valid API key must be supplied.");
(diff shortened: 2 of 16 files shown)


Source/Bugsnag.h
```diff
@@ -27,6 +27,7 @@
 
 #import "BugsnagConfiguration.h"
 #import "BugsnagMetaData.h"
+#import "BugsnagPlugin.h"
```
fractalwrench: Is this import required?

Source/Bugsnag.m
```diff
@@ -54,6 +57,7 @@ + (void)startBugsnagWithConfiguration:(BugsnagConfiguration *)configuration {
         if ([configuration hasValidApiKey]) {
             bsg_g_bugsnag_notifier =
                     [[BugsnagNotifier alloc] initWithConfiguration:configuration];
+            [self startPlugins];
```
fractalwrench: Does it make sense to initialise the plugins before the notifier has been initialised? My first thought is that it would make more sense to start the plugins after `[bsg_g_bugsnag_notifier start]` has been invoked, as this could lead to unintended behaviour if the plugin believes the notifier is already initialised. I'm unsure _why_ we need a plugin interface so it'd be good to understand more about that too.

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/CustomPluginNotifierDescriptionScenario.m
```diff
@@ -0,0 +1,46 @@
...
+
+@implementation DescriptionPlugin
+
+- (instancetype)init {
+    self = [super init];
+    return self;
+}
+
+- (void)start {
+    id notifier = [Bugsnag notifier];
```
fractalwrench: Could we consider passing `notifier` as a parameter to `start`? This would avoid each plugin implementation requiring knowledge about `Bugsnag.h`

kattrali: The intent was to keep the interface internal for now. If we want to open it up later as a feature, then it would make sense to add it to a public header instead. As it is, access to use is gated by internal APIs, and the library is vendored, so there's no chance of the API breaking without also breaking the build process.

Source/Bugsnag.m
kattrali: If its initialized _after_ start, then reports could be sent or breadcrumbs saved, etc, before the plugin has a chance to modify the configuration or client instance.

Source/Bugsnag.h
kattrali: Yes. The header is a part of the public module export, so it has to be included in the umbrella header (Bugsnag.h).

Source/BugsnagNotifier.m
```diff
@@ -218,12 +218,23 @@ - (id)initWithConfiguration:(BugsnagConfiguration *)initConfiguration {
...
+#elif TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
+            @"iOS Bugsnag Notifier";
+#elif TARGET_OS_MAC
+            @"OSX Bugsnag Notifier";
+#else
+            @"Bugsnag Objective-C";
+#endif
         self.details = [@{
-            BSGKeyName : @"Bugsnag Objective-C",
+            BSGKeyName : notifierName,
```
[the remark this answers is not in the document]
kattrali: Previously, the name was set in `start`, this change moves it into the initializer so it can be overwritten if needed and the changes aren't lost when start is called.

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/CustomPluginNotifierDescriptionScenario.m
fractalwrench: Regardless, I think it'd be simpler for our own internal plugin implementations if we could just pass the notifier in as a parameter. Not too fussed on this point so I'll leave it up to you.

kattrali: The notifier is a private class, and `[Bugsnag notifier]` is private (not in any header). Even if it was passed in, the interface for notifier would have to come from somewhere, so passing it in wouldn't fix this issue.

fractalwrench: Ah got it - I had assumed `BugsnagNotifier` was visible. This approach makes sense.

PR #441, opened 2020-01-02
refactor: Remove unused file util helpers

kattrali:
Removes a few unused functions


robinmacharg:
Looks good.  Couple of questions.
What’s the methodology for determining unused functions?
How confident are we that there's no dynamic dispatch that could invoke the removed methods?

kattrali:
> What’s the methodology for determining unused functions?

A grep of the codebase. The functions names were not present outside of the source file and unit tests for that file.

> How confident are we that there's no dynamic dispatch that could invoke the removed methods?

Confident. There's no usages of `objc_msgSend`, manually constructing selectors, etc outside of crash test cases.

PR #401, opened 2019-08-09
Fix possible crash or deadlock arising from calling notify() from multiple queues concurrently

kattrali:
## Goal

Its possible that when `notify()` is called concurrently from multiple queues, that a crash occurs when caching/resetting crash state. This changeset solves the underlying issues by not sharing state between calls to `notify()` and isolating calls to suspend threads (for backtrace collection).

### Reproduction case

```objc
dispatch_queue_t queue1 = dispatch_queue_create("Queue 1", DISPATCH_QUEUE_CONCURRENT);
dispatch_queue_t queue2 = dispatch_queue_create("Queue 2", DISPATCH_QUEUE_CONCURRENT);

for (int i = 0; i < 4; i++) {
    NSString *message = [NSString stringWithFormat:@"Err %ld", (long)i];
    NSError *error = [FooError errorWithDomain:@"com.example"
                                          code:340
                                      userInfo:nil];
    dispatch_async(queue1, ^{
        [Bugsnag notifyError:error];
    });
    dispatch_async(queue2, ^{
        [Bugsnag notifyError:error];
    });
}
```

Fixes #399

## Changeset

The changes can be divided into four parts:

1. Adding a (failing on master) test which calls `notify()` several times from different concurrent queues
2. Refactoring crash reporting internals to allow providing separate crash contexts for each invocation of `notify()`
3. Implementing separate crash contexts for `notify()` calls
4. Adding locking around suspending threads and writing crash state.

The easiest way to view the changeset is one commit at a time, since each chunk can be evaluated independently.

## Tests

* Tested manually on a couple different iOS/tvOS devices running iOS 11 & 12
* Added automated tests replicating the failing scenario

PR #442, opened 2020-01-02
fix(utils): Remove global buffer used when writing files to disk

kattrali:
Fixes a case where shared memory buffer is used for writing to disk in a code path which can be accessed concurrently from multiple threads. Triggering the error on write requires calling notify from multiple threads at once while `suspendThreadsForUserReported` is disabled.

This change removes the buffer altogether, opting for a simpler solution where the system determines how/when to buffer when writing to disk. The buffer was added in kstenerud/KSCrash@be9bedb46b7980e9051e6ab445ea37bf0146733a so the original intent is somewhat lost to the mists.

## Tests

Tests for this are a bit fiddly since the issue is timing-based, though its possible to see the effect when using the reproduction case in #401 while setting `suspendThreadsForUserReported = NO`. Added an automated test for this case which would previously fail intermittently, as well as testing manually on iOS 12 / 13.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+## TBD
+
+## Bug fixes
+
+* Fix possible report corruption when using `notify()` from multiple threads
+  when configured to skip capturing/reporting background thread contents
+  (generally only Unity games).
+
 ## 5.23.0 (2019-12-10)
 
 This release removes support for reporting 'partial' or 'minimal' crash reports
--- a/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
+++ b/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
@@ -1658,9 +1658,6 @@ void bsg_kscrashreport_writeStandardReport(
 
     bsg_ksjsonendEncode(bsg_getJsonContext(writer));
 
-    if (!bsg_ksfuflushWriteBuffer(fd)) {
-        BSG_KSLOG_ERROR("Failed to flush write buffer");
-    }
     close(fd);
 }
 
--- a/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashState.m
+++ b/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashState.m
@@ -263,9 +263,6 @@ bool bsg_kscrashstate_i_saveState(const BSG_KSCrash_State *const state,
         goto done;
     }
     result = bsg_ksjsonendEncode(&JSONContext);
-    if (!bsg_ksfuflushWriteBuffer(fd)) {
-        BSG_KSLOG_ERROR(@"Failed to flush write buffer");
-    }
 
 done:
     close(fd);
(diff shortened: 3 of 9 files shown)


robinmacharg:
A few (maybe newbie) questions:

- Is this a good opportunity to slightly expand the relevant `KSFileUtils.h` function header documentation to highlight the potential failure mode. On a (minor) related note: `bsg_ksfureadBytesFromFD` has an erroneous log message: `BSG_KSLOG_ERROR("Could not write to fd...`
- `KSFileUtils_Tests.m` has a commented out test: testWriteBytesToFDBig.  Should this be reinstated, or removed at the same time as this change?
- There's an implicit (via the now-removed buffer declaration) 64K limit on the size of the passed in data that's been removed with this change.  Is it worth stating/checking the expected size limits of passed in data?
- What alternative approaches (to removing the buffer) have been considered?
- Concurrency, but related to the previous point: is there any possibility that an error in one thread's `write()` halfway through another's could affect operation? Are we safe avoiding semaphore's etc. in this case?

kattrali:
> Is this a good opportunity to slightly expand the relevant `KSFileUtils.h` function header documentation to highlight the potential failure mode? On a (minor) related note: `bsg_ksfureadBytesFromFD` has an erroneous log message:  `BSG_KSLOG_ERROR("Could not write to fd...`

Rebasing to apply the changes from #441, removing the erroneously documented functions.

> KSFileUtils_Tests.m has a commented out test: testWriteBytesToFDBig. Should this be reinstated, or removed at the same time as this change?

This actually tested one of the unused functions, so its going away too.

> There's an implicit (via the now-removed buffer declaration) 64K limit on the size of the passed in data that's been removed with this change. Is it worth stating/checking the expected size limits of passed in data?

I'd say no, because write(2) is smart enough to handle this case, and if any bytes remain after calling `write()`, then the `bsg_ksfuwriteBytesToFD()` function loops until all of the remaining data is written.

> What alternative approaches (to removing the buffer) have been considered?

* Add the new writing function *only* for `notify()` reports so the codepath did not need to change for crash reports. It seemed more difficult or maybe not worth the effort since the use of `bsg_ksfuwriteBytesToFD()` is deep in shared functions.
* Locks/semaphores weren't really an option since this codepath is hit at crash time and needs to be async signal-safe.

> Concurrency, but related to the previous point: is there any possibility that an error in one thread's `write()` halfway through another's could affect operation? Are we safe avoiding semaphore's etc. in this case?

Hopefully no:

* In the `notify()` case, [a new file path is generated for every report](https://github.com/bugsnag/bugsnag-cocoa/blob/b7cf1344238d47deb4d6d9e8182cf07e446b0236/Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_User.c#L61-L65).
* In the crash report case, all other threads are stopped (using `bsg_kscrashsentry_suspendThreads()`) as soon as the crash is detected, which should negate simulateous reports to the same file path.
* If `write()` fails, it sets `errno` rather than raising or otherwise terminating ungracefully. `errno` is thread-local ([1](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/sys/errno.h#L82), [2](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/libsyscall/custom/errno.c#L52), [3](https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/libsyscall/os/tsd.h#L100)).

PR #444, opened 2020-01-17
fix (oom): added missing event fields

robinmacharg:
## Goal

To add fields that appear in non-OOM events to OOM events (codeBundleId, locale, osName and modelNumber). 

## Design

The fields are trivial additions that add values to the info cached by the OOM watchdog.

## Changeset

New fields added to the `generateCacheInfoWithConfig()` method of `BSGOutOfMemoryWatchdog`.

## Tests

Manual testing to ensure values were being sent through.  An additional unit test was added to ensure that expected keys/values are present in the cache.

### Outstanding Questions

There is some commonality with non-OOM fields.  Is there scope for generalising the event field population?

Expected fields are tested for.  There are no tests for unexpected fields; should there be?

  - [X] Final review

- [X] The correct target branch has been selected (`master` for fixes, `next` for
  features)

### DIFF
--- a/Source/BSGOutOfMemoryWatchdog.m
+++ b/Source/BSGOutOfMemoryWatchdog.m
@@ -243,6 +243,8 @@
     app[@"releaseStage"] = config.releaseStage;
     app[@"version"] = systemInfo[@BSG_KSSystemField_BundleShortVersion] ?: @"";
     app[@"bundleVersion"] = systemInfo[@BSG_KSSystemField_BundleVersion] ?: @"";
+    // 'codeBundleId' only (optionally) exists for Unity clients and defaults otherwise to nil
+    app[@"codeBundleId"] = [config codeBundleId];
 #if BSGOOMAvailable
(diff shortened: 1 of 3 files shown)


Source/BSGOutOfMemoryWatchdog.m
```diff
@@ -243,6 +243,8 @@ - (NSMutableDictionary *)generateCacheInfoWithConfig:(BugsnagConfiguration *)con
     app[@"releaseStage"] = config.releaseStage;
     app[@"version"] = systemInfo[@BSG_KSSystemField_BundleShortVersion] ?: @"";
     app[@"bundleVersion"] = systemInfo[@BSG_KSSystemField_BundleVersion] ?: @"";
+    // 'codeBundleId' only (optionally) exists for Unity clients and defaults otherwise to nil
```
kattrali:
```suggestion
    // 'codeBundleId' only (optionally) exists for React Native clients and defaults otherwise to nil
```

PR #445, opened 2020-01-22
feat: Add a breadcrumb when the library starts

kattrali:
From Breadcrumbs > Behavior > Automatic breadcrumb types:

> #### Bugsnag loaded
>
> - Bugsnag loaded `Breadcrumbs` shall be implemented on every platform.
> - When the notifier has been initialized and starts tracking errors a `Breadcrumb` shall be collected.
> - The `Breadcrumb` shall have the `type` set to 'state'.
> - The `message` shall be set to 'Bugsnag loaded'.

### DIFF
--- a/Source/BugsnagNotifier.m
+++ b/Source/BugsnagNotifier.m
@@ -280,6 +280,7 @@ NSString *const kUndoOperation = @"Undo Operation";
 NSString *const kRedoOperation = @"Redo Operation";
 NSString *const kTableViewSelectionChange = @"TableView Select Change";
 NSString *const kAppWillTerminate = @"App Will Terminate";
+NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
 
 - (void)initializeNotificationNameMap {
     notificationNameMap = @{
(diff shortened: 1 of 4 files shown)


Source/BugsnagNotifier.m
```diff
@@ -411,6 +411,11 @@ - (void)start {
 
     [self.sessionTracker startNewSessionIfAutoCaptureEnabled];
 
+    [self addBreadcrumbWithBlock:^(BugsnagBreadcrumb *_Nonnull breadcrumb) {
+         breadcrumb.type = BSGBreadcrumbTypeState;
+         breadcrumb.name = @"Bugsnag loaded";
```
robinmacharg: Should this be a constant? Breadcrumb names seem to have a fair amount of indirection already.  Is there any value in concentrating possible string values (for automatically captured breadcrumbs) in a central location?

kattrali: I thought something similar, though I think automatic breadcrumbs overall needs a better home than this class and a place for all of the constants and conversions. In the meantime, I've moved the name into a constant and added a backlog refactor task to figure out a proper place for these.

PR #446, opened 2020-01-22
feat(spec): Added designated initializer to BugsnagConfiguration

robinmacharg:
## Goal

The notifier spec. states that an `apiKey` must be provided when initializing a `BugsnagConfiguration`.  To this end removing the default constructor and providing a new apiKey-accepting designated initializer is the solution.

## Design

An initializer has been added to `BugsnagConfiguration` and its designated status indicated with the `NS_DESIGNATED_INITIALIZER` macro.  `init()` is required (calls to designated initializers MUST call super) but has been made to fail if called.  It was desired to provide `apiKey` as an unnamed parameter but the `NS_SWIFT_NAME` macro doesn't appear to offer this functionality.

## Changeset

A new initializer was added.  The default `init()` was made to fail.

## Tests

Unit tests were added to ensure that the new initializer performed as expected.  All tests that depended on the default init() have been modified to use the new initializer.  A Swift test was added to ensure that the method is exposed correctly.

### Outstanding Questions

  - [x] Final review

  features)

### DIFF
--- a/Source/BugsnagConfiguration.m
+++ b/Source/BugsnagConfiguration.m
@@ -36,6 +36,10 @@
 static NSString *const kHeaderApiPayloadVersion = @"Bugsnag-Payload-Version";
 static NSString *const kHeaderApiKey = @"Bugsnag-Api-Key";
 static NSString *const kHeaderApiSentAt = @"Bugsnag-Sent-At";
+static NSString *const BSGApiKeyError = @"apiKey must be a 32-digit hexadecimal value.";
+static NSString *const BSGInitError = @"Init is unavailable.  Use [[BugsnagConfiguration alloc] initWithApiKey:] instead.";
+static const int BSGApiKeyLength = 32;
+NSString * const BSGConfigurationErrorDomain = @"com.Bugsnag.CocoaNotifier.Configuration";
 
 @interface Bugsnag ()
 + (BugsnagNotifier *)notifier;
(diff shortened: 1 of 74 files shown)


Source/BugsnagConfiguration.m
```diff
@@ -46,43 +49,60 @@ @interface BugsnagNotifier ()
...
+- (instancetype)init {
+    @throw BSGInitError;
+}
+
+/**
+ * The designated initializer.
+ */
+-(instancetype)initWithApiKey:(NSString *)apiKey {
+    if (! [self isValidApiKey:apiKey]) {
+        @throw BSGApiKeyError;
```
kattrali: I played around with this a bit, and noticed its not catchable in any way from Swift. The primary case where this might be missed in development is that sometimes people only initialize  bugsnag for their beta/release builds, so this could be an unexpected crash in a release.

I thought of an alternative here, where `initWithApiKey` takes an error parameter and converts into a `try` statement in Swift. Something like:

```objc
// BugsnagConfiguration.h
- (instancetype _Nullable)initWithApiKey:(NSString *_Nonnull) apiKey error:(NSError *_Nullable *_Nullable)error
    NS_DESIGNATED_INITIALIZER NS_SWIFT_NAME(init(_:)) __attribute__((swift_error(nonnull_error)));
```

which then makes the Swift initializer testable for errors using

```swift
let config = try BugsnagConfiguration("{a totally valid API key here}")
```

Does this make sense and seem like a workable way forward?

robinmacharg: Yes, good spot.  At the acknowledged (idiomatic) expense of a slightly longer invocation.  These changes - with the additional Swift tests they make possible - are now checked in.

Tests/BugsnagSwiftConfigurationTests.swift
```diff
@@ -0,0 +1,36 @@
...
+     * (read: longer) ObjC invocation).
+     */
+    func testDesignatedInitializerHasCorrectNS_SWIFT_NAME() {
+        
+        do {
+            let _ = try BugsnagConfiguration(DUMMY_APIKEY_16CHAR)
+        }
+        catch let e as NSError {
+            XCTAssertEqual(e.domain, BSGConfigurationErrorDomain)
+            XCTAssertEqual(e.code, BSGConfigurationErrorCode.invalidApiKey.rawValue)
```
kattrali: In the case that the expected values for `BSGConfigurationErrorDomain` and `BSGConfigurationErrorCode` are inadvertently changed, this test would still pass. Testing for the expected literal values here would catch that.

```suggestion
            XCTAssertEqual(e.domain, "com.Bugsnag.CocoaNotifier.Configuration")
            XCTAssertEqual(e.code, 0)
```

Tests/BugsnagSwiftConfigurationTests.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * Objective C trailing-NSError* initializers are translated into throwing
+     * Swift methods, allowing us to fail gracefully (at the expense of a more-explicit
+     * (read: longer) ObjC invocation).
+     */
+    func testDesignatedInitializerHasCorrectNS_SWIFT_NAME() {
+        
+        do {
+            let _ = try BugsnagConfiguration(DUMMY_APIKEY_16CHAR)
```
kattrali: If this line "succeeds" then the test still passes though the behavior is broken. An `XCTFail()` assertion after this line would catch that.

Tests/BugsnagSwiftConfigurationTests.swift
robinmacharg: Good comment.  The thinking was that a user can incorporate the constants in their own use of the library, in which case equality checks (regardless of the actual values) would still work for them.  If we're testing that values of constants are set in stone I'd prefer to put that in a separate test to highlight the specific thing that may have changed.

kattrali: Ah I see. Exposing the constants is great for letting people determine what went wrong, and its nice to have them exposed. So maybe there should be tests for both things (that the constant names are correct and the constant values are what we expect)?

porque no los dos

robinmacharg: Win-win.  At the risk of overkill we now have both tests in.

PR #448, opened 2020-01-24
Add a breadcrumb automatically for changes in network connectivity

kattrali:
Adds a breadcrumb each time the connection status noticeably changes, such as between no connectivity, cellular, and  wifi, using the Bugsnag server URL as a proxy for overall connectivity.

Refactors the connectivity checker as a part of this change to report connectivity in a way suitable to append to reports, reduce object references and increase documentation and testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@ Bugsnag Notifiers on other platforms.
   `Bugsnag.notify()` callback.
   [#458](https://github.com/bugsnag/bugsnag-cocoa/pull/458)
 
+* Add a breadcrumb when network connectivity changes
+
 ## Bug fixes
 
 * Fix possible report corruption when using `notify()` from multiple threads
--- a/Source/BSGConnectivity.m
+++ b/Source/BSGConnectivity.m
@@ -25,95 +25,114 @@
 //
 
 #import "BSGConnectivity.h"
+#import "Bugsnag.h"
 
-typedef void (^CallbackBlock)(SCNetworkReachabilityFlags flags);
+static SCNetworkReachabilityRef bsg_reachability_ref;
+BSGConnectivityChangeBlock bsg_reachability_change_block;
+SCNetworkReachabilityFlags bsg_current_reachability_state;
+
+NSString *const BSGConnectivityCellular = @"cellular";
+NSString *const BSGConnectivityWiFi = @"wifi";
+NSString *const BSGConnectivityNone = @"none";
+
+/**
+ * Check whether the connectivity change should be noted or ignored.
+ *
+ * @return YES if the connectivity change should be reported
+ */
+BOOL BSGConnectivityShouldReportChange(SCNetworkReachabilityFlags flags) {
+    #if TARGET_OS_TV || TARGET_OS_IPHONE
+        // kSCNetworkReachabilityFlagsIsWWAN does not exist on macOS
+        const SCNetworkReachabilityFlags importantFlags = kSCNetworkReachabilityFlagsIsWWAN | kSCNetworkReachabilityFlagsReachable;
+    #else
+        const SCNetworkReachabilityFlags importantFlags = kSCNetworkReachabilityFlagsReachable;
+    #endif
+    __block BOOL shouldReport = YES;
+    // Check if the reported state is different from the last known state (if any)
+    if ((flags & importantFlags) != (bsg_current_reachability_state & importantFlags)) {
+        // When first subscribing to be notified of changes, the callback is
+        // invoked immmediately even if nothing has changed. So this block
+        // ignores the very first check, reporting all others.
+        if (bsg_current_reachability_state == 0) {
+            shouldReport = NO;
+        }
+        // Cache the reachability state to report the previous value representation
+        bsg_current_reachability_state = flags;
+    } else {
+        shouldReport = NO;
+    }
+    return shouldReport;
+}
+
+/**
+ * Textual representation of a connection type
+ */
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Changelog
 
 * Add a breadcrumb when Bugsnag first starts with the message "Bugsnag loaded"
   [#445](https://github.com/bugsnag/bugsnag-cocoa/pull/445)
+* Add a breadcrumb when network connectivity changes
```
robinmacharg:
```suggestion
* Add a breadcrumb when network connectivity changes
  [#448](https://github.com/bugsnag/bugsnag-cocoa/pull/448)
```

Source/BSGConnectivity.m
```diff
@@ -35,85 +84,51 @@
...
 
-- (instancetype)initWithURL:(NSURL *)url
-                changeBlock:(ConnectivityChange)changeBlock {
-    if (self = [super init]) {
-        NSString *hostName = [url host];
-        _reachabilityRef = SCNetworkReachabilityCreateWithName(NULL, [hostName UTF8String]);
-
-        _connectivityChangeBlock = [changeBlock copy];
-        self.serialQueue = dispatch_queue_create("com.bugsnag.cocoa", NULL);
++ (void)monitorURL:(NSURL *)URL usingCallback:(BSGConnectivityChangeBlock)block {
```
robinmacharg: Would a different name better indicate the purpose of this method (which is not to monitor a URL but to send a breadcrumb *to* a URL).  It's not snappy (but then ObjC seldom is), but maybe something like: `sendConnectivityChangeEventTo:usingCallback:`?

Source/BSGConnectivity.m
```diff
@@ -35,85 +84,51 @@
...
-    };
-
-    if (self.reachabilityRef) {
-        SCNetworkReachabilitySetCallback(self.reachabilityRef,
-                                         BSGConnectivityCallback, &context);
-        SCNetworkReachabilitySetDispatchQueue(self.reachabilityRef,
-                                              self.serialQueue);
-    }
++ (BOOL)isValidHostname:(NSString *)host {
+    return host.length > 0 && ![host isEqualToString:@"localhost"];
```
robinmacharg: This doesn't catch `127.0.0.1`.  Since there's a (potential) loss of network at this point is an explicit check for IP address better than a more general host to IP resolution?  On desktop platforms this will also fail to catch `/etc/hosts` host aliasing but since this (along with endpoint reconfiguration) is an intentional user-action it feels safe to ignore the edge case.

Source/BugsnagNotifier.m
```diff
@@ -521,13 +521,13 @@ - (void)flushPendingReports {
...
-    self.networkReachable =
-        [[BSGConnectivity alloc] initWithURL:url
-                                 changeBlock:^(BSGConnectivity *connectivity) {
-                                   [weakSelf flushPendingReports];
-                                 }];
-    [self.networkReachable startWatchingConnectivity];
+    __weak typeof(self) weakSelf = self;
+    [BSGConnectivity monitorURL:url
+                  usingCallback:^(BOOL connected, NSString *connectionType) {
+        __strong typeof(weakSelf) strongSelf = weakSelf;
```
robinmacharg: A note (optionally with references, although Apple's relevant docs seem to have disappeared) might help explain the *why* of the `weakSelf/strongSelf` dance to developers not familiar with Objective C/ARC/Reference counting.  [This blog](https://dhoerl.wordpress.com/2013/04/23/i-finally-figured-out-weakself-and-strongself/), for instance, offers *an* explanation.

Tests/BSGConnectivityTest.m
```diff
@@ -0,0 +1,60 @@
...
+- (void)testValidHost {
+    XCTAssertTrue([BSGConnectivity isValidHostname:@"example.com"]);
+    // Could be an internal network hostname
+    XCTAssertTrue([BSGConnectivity isValidHostname:@"foo"]);
+
+    // Definitely will not work as expected
+    XCTAssertFalse([BSGConnectivity isValidHostname:@""]);
+    XCTAssertFalse([BSGConnectivity isValidHostname:nil]);
+    XCTAssertFalse([BSGConnectivity isValidHostname:@"localhost"]);
+}
```
robinmacharg: This will also need a test for `127.0.0.1` if the above comment is accepted.

Source/BSGConnectivity.m
```diff
@@ -25,8 +25,57 @@
...
+NSString *BSGConnectivityFlagRepresentation(SCNetworkReachabilityFlags flags) {
+    BOOL connected = (flags & kSCNetworkReachabilityFlagsReachable);
+    #if TARGET_OS_TV || TARGET_OS_IPHONE
+        return connected
+            ? ((flags & kSCNetworkReachabilityFlagsIsWWAN) ? @"cellular" : @"wifi")
+            : @"none";
+    #else
+        return connected ? @"wifi" : @"none";
+    #endif
+}
```
robinmacharg: Should the values be represented by constants?  e.g. "wifi" is repeated and so potentially corruptible.  Might also help clarity in tests.

kattrali:
> And it would be useful to know the anticipated impact on normal event reporting. For instance would rapid repeated connectivity changes cause other event types to be dropped? If connectivity is unstable there may not be time to send existing stored events before the next connectivity change (which may in turn may cause an app failure).

This question might have been a misunderstanding - this changeset just leaves breadcrumbs for connectivity changes, no events are being generated by the event.

PR #456, opened 2020-02-05
feat: add method to add metadata section values as a dictionary

robinmacharg:
## Goal

Adds a `BugsnagMetadata.addMetadataToSection:values:` method to allow merging arbitrary dictionaries with preexisting Metadata.

## Design

Simple fast enumeration loop over provided values with appropriate checks for:

- Validity of Key (String subclass)
- Validity of Value (can be sanitised: a recursive check that all (nested) values are of type Number, String, Array, Dictionary or Null

Conditional addition/deletion depending on whether value is null or not.  There is a check to ensure that empty sections are not created for (failing) invalid values.  Invalid values are logged and ignored.  Delegate callback is called regardless of whether the method actually made a material change to the metadata.

## Changeset

- Method added to `BugsnagMetadata`.
- Unit Tests added.

## Tests

- Added Unit tests only.  This change affects the framework API and doesn't change any external aspect.
- CI passes.

### Outstanding Questions

  - [ ] Final review

  features)
- [X] Performance and complexity - are there any cases of unexpected O(n^3) when iterating, recursing, flat mapping, etc?


Tests/BugsnagMetadataTest.m
```diff
@@ -0,0 +1,184 @@
...
+
+@end
+
+// MARK: - Tests
+
+@implementation BugsnagMetadataTest
+
+@synthesize delegateCalled;
+
+- (void)test_addAttribute_withName {
```
fractalwrench: The contents of this test looks good - if it's straightforward would it be possible to separate these test cases into individual functions? E.g., one for creation, one for checking an arbitrary tab name creates itself, etc.

Tests/BugsnagMetadataTest.m
```diff
@@ -0,0 +1,184 @@
...
+    tab2 = [metadata getTab:@"FirstTab"];
+    XCTAssertEqual(tab2.count, 0);
+    XCTAssertTrue(delegateCalled, "Expected the delegate's metadataChanged: method to be called.");
+
+    // Adding invalid values should fail silently (and not add the value)
+    delegateCalled = NO;
+    [metadata addAttribute:@"bar" withValue:[DummyClass new] toTabWithName:@"FirstTab"];
+    tab2 = [metadata getTab:@"FirstTab"];
+    XCTAssertEqual(tab2.count, 0);
+    XCTAssertTrue(delegateCalled, "Expected the delegate's metadataChanged: method to be called.");
```
fractalwrench: If _all_ the values are invalid and nothing has changed in the metadata, then we could probably avoid calling the delegate here - particularly if it does a lot of work. I'm unsure if this is realistically going to happen often, so would be good to get your judgement on whether you think it's worth making that change.

Tests/BugsnagMetadataTest.m
robinmacharg: NP, extended setUp() slightly, split tests up and adjusted any cumulative values being tested to account for this.

Tests/BugsnagMetadataTest.m
robinmacharg: Agreed.  I'd noted that.  It's small so I've made the change, ensuring that any delegate calls occur at most once per operation, and outside @synchronised sections.  Two tests had to be adjusted to check that the delegate is NOT called.

PR #458, opened 2020-02-06
Add apiKey as a mutable property of BugsnagEvent

robinmacharg:
## Goal

Add an `apiKey` property to `BugsnagEvent` (nee `BugsnagCrashReport`).  The behaviour is:

- Library consumers should be able to change the API key of event from OnError callbacks
- The property defaults to the value from `BugsnagConfiguration.apiKey`. 
- Changing this value should change the API key used in the payload body and header but not the global Configuration value.

## Design

A simple `@property` with custom getter/setter logic and validation.  The `BugsnagConfiguration.isValidApiKey` method was promoted to class-level since it's now used in two places.

Failing validation logs the failure but doesn't throw an error.  The thinking is that there will always be a fallback valid (32 digit hex string) apiKey and so event capture is the priority.

## Changeset

Just `BugsnagEvent.m/h` and associated unit tests.

## Tests

Tested that:
- The value retrieved the Configuration value when unset.
- A new value could be set in the event callback and this did not change the global value.
- The validation failure behaviour is correct.

Did not add any testing that the overridden `apiKey` value is correctly being sent to the backend via HTTP headers although this was observed in informal testing.  A solution to this involving swizzling the `ApiClient` class was examined but deemed unnecessarily complex.

### Outstanding Questions

  - [X] Final review

  features)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ Bugsnag Notifiers on other platforms.
 * Add a breadcrumb when Bugsnag first starts with the message "Bugsnag loaded"
   [#445](https://github.com/bugsnag/bugsnag-cocoa/pull/445)
 
+* Add a per-Event `apiKey` property.  This defaults to the global 
+  `BugsnagConfiguration` value but can be overridden in event passed to the 
+  `Bugsnag.notify()` callback.
+  [#458](https://github.com/bugsnag/bugsnag-cocoa/pull/458)
+
 ## Bug fixes
 
 * Fix possible report corruption when using `notify()` from multiple threads
--- a/Source/BugsnagConfiguration.h
+++ b/Source/BugsnagConfiguration.h
@@ -84,6 +84,7 @@ typedef NSDictionary *_Nullable (^BugsnagBeforeNotifyHook)(
     NSArray *_Nonnull rawEventReports, NSDictionary *_Nonnull report);
 
 @interface BugsnagConfiguration : NSObject
+
 /**
  *  The API key of a Bugsnag project
  */
--- a/Tests/BugsnagEventTests.m
+++ b/Tests/BugsnagEventTests.m
@@ -26,14 +26,14 @@
     config.releaseStage = @"foo";
     BugsnagHandledState *state =
         [BugsnagHandledState handledStateWithSeverityReason:HandledException];
-    BugsnagEvent *report =
+    BugsnagEvent *event =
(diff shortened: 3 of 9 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,11 @@ Bugsnag Notifiers on other platforms.
 * Add a breadcrumb when Bugsnag first starts with the message "Bugsnag loaded"
   [#445](https://github.com/bugsnag/bugsnag-cocoa/pull/445)
 
+* Add a per-Event `apiKey` property.  This defaults to the global 
+  `BugsnagConfiguration` value but can be overridden in event pased in to the
```
fractalwrench: s/in event pased in/in events passed to

Tests/BugsnagEventTests.m
```diff
@@ -412,4 +412,58 @@ - (void)testReportAddMetadata {
...
+ */
+- (void)testApiKey {
+    NSError *error;
+    BugsnagConfiguration *config = [[BugsnagConfiguration alloc] initWithApiKey:DUMMY_APIKEY_32CHAR_1 error:&error];
+    [Bugsnag startBugsnagWithConfiguration:config];
+
+    NSException *ex = [[NSException alloc] initWithName:@"myName" reason:@"myReason1" userInfo:nil];
+    
+    // Check that the event is passed the apiKey
+    [Bugsnag notify:ex block:^(BugsnagEvent * _Nonnull report) {
```
fractalwrench: nit: would be good to rename variables named `report` in this file to `event`

Source/BugsnagConfiguration.h
```diff
@@ -84,6 +84,13 @@ typedef NSDictionary *_Nullable (^BugsnagBeforeNotifyHook)(
     NSArray *_Nonnull rawEventReports, NSDictionary *_Nonnull report);
 
 @interface BugsnagConfiguration : NSObject
+
+/**
+ * Determine the apiKey-validity of a passed-in string:
+ * Exactly 32 hexadecimal digits.
+ */
++ (BOOL)isValidApiKey:(NSString *_Nullable)apiKey;
```
fractalwrench: Would this method be publicly accessible? If so, I think we should attempt to hide this method as otherwise we'd have to keep the method signature public until the next major version.

robinmacharg: Good catch, thanks.  Moved to an `@interface` section in `BugsnagEvent.m` to limit the visibility.

PR #459, opened 2020-02-10
feat: Add Bugsnag.getMetadata()

robinmacharg:
## Goal

Provide metadata retrieval from the global `Bugsnag` class.

## Design

A class method was added to the `Bugsnag` class, exposed appropriately to Swift.  Equivalent lower-level methods were renamed to align them with the implemented method.  The functionality was changed from "create if not available" to "return nil if not available".

## Changeset

`Bugsnag` and `BugsnagMetadata`, along with appropriate unit test  files.

## Tests

Unit tests were provided in Objective C (and Swift to confirm name exposure).

### Outstanding Questions

  - [X] Final review

  features)
- [X] Consistencyxbetween the changeset and the goal stated above


Source/Bugsnag.h
```diff
@@ -264,6 +266,16 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
...
+/**
+* Return the metadata for a specific named section
+*
+* @param section The name of the section
+* @returns The mutable dictionary representing the metaadata section, if it
+*          exists, or nil if not.
+*/
++ (NSMutableDictionary *_Nullable)getMetadata:(NSString *_Nonnull)section
+    NS_SWIFT_NAME(getMetadata(_:));
+
```
fractalwrench: The `Bugsnag` interface is missing a couple of the method variants that are defined in the notifier spec. For instance, 

- `getMetadata:section:key` could return an object of type `id`
- `clearMetadata:section:key` could clear a specific value
- `addMetadata:section` could take an `NSDictionary`, acting as syntactic sugar for `addMetadata:section:key`

Source/Bugsnag.h
```diff
@@ -264,6 +266,16 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
  */
 + (BOOL)resumeSession;
 
+/**
+* Return the metadata for a specific named section
+*
+* @param section The name of the section
+* @returns The mutable dictionary representing the metaadata section, if it
```
fractalwrench: s/metaadata/metadata

Source/BugsnagMetadata.m
```diff
@@ -86,15 +81,22 @@ - (void)addAttribute:(NSString *)attributeName
...
                 Class klass = [value class];
                 bsg_log_err(@"Failed to add metadata: Value of class %@ is not "
                             @"JSON serializable",
                             klass);
             }
-        } else {
-            [[self getTab:tabName] removeObjectForKey:attributeName];
+        }
+        else {
+            [[self getMetadata:tabName] removeObjectForKey:attributeName];
```
fractalwrench: nit: it'd be good to rename `tabName` to `section` for consistency across this file

Source/BugsnagMetadata.m
```diff
@@ -86,15 +81,22 @@ - (void)addAttribute:(NSString *)attributeName
         if (value) {
             id cleanedValue = BSGSanitizeObject(value);
             if (cleanedValue) {
-                [self getTab:tabName][attributeName] = cleanedValue;
-            } else {
+                NSDictionary *section = [self getMetadata:tabName];
+                if (!section) {
+                    [[self dictionary] setObject:[NSMutableDictionary new] forKey:tabName];
+                    section = [self getMetadata:tabName];
```
fractalwrench: Would it make sense to hold onto the `[NSMutableDictionary new]` here rather than call into `[self getMetadata]`? It's been a while since I looked at Objective-C in detail but I believe that would avoid checking a lock via the `@synchronized` keyword. Are there any potential downsides that mean using `[self getMetadata]` would be preferable?

iOS/BugsnagTests/Swift Tests/BugsnagMetadataTests.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+
+import XCTest
+
+class BugsnagMetadataTests: XCTestCase {
+
+    /**
+     * Test that getMetadata() is exposed to Swift correctly
+     */
+    func testgetMetadataIsExposedCorrectly() {
```
fractalwrench: My interpretation of the spec is that the `BugsnagMetadata` class itself doesn't need to be exposed, and that it's sufficient to simply specify the `addMetadata/clearMetadata/getMetadata` methods instead. Perhaps that's something we could check with the team before thinking of changing anything here?

iOS/BugsnagTests/Swift Tests/BugsnagSwiftTests.swift
```diff
@@ -0,0 +1,48 @@
...
+        do {
+            if let configuration = try BugsnagConfiguration(DUMMY_APIKEY_32CHAR_1) {
+                Bugsnag.start(with: configuration)
+                Bugsnag.addMetadata("mySection1", key: "myKey1", value: "myValue1")
+                
+                let exception1 = NSException(name: NSExceptionName(rawValue: "exception1"), reason: "reason1", userInfo: nil)
+                
+                Bugsnag.notify(exception1) { (event) in
+                    // Arbitrary test, replicating the ObjC one
+                    let value = (event.metadata["mySection1"] as? [String : Any])?["myKey1"] as? String
```
fractalwrench: `event.metadata` should be replaced by `[event getMetadata]` and all the other associated methods - for the purposes of this review I'm assuming that this will be implemented in another PR.

Tests/BugsnagTests.m
```diff
@@ -0,0 +1,80 @@
...
+
+@interface BugsnagTests : XCTestCase
+
+@end
+
+@implementation BugsnagTests
+
+/**
+ * Test that global metadata is added correctly, applied to each event, and
+ * deleted appropriately.
```
fractalwrench: When testing events we should probably also check that any alteration to metadata on the event doesn't affect the metadata on the configuration object either. I'm assuming we have existing test coverage for this or can add some in when the changes on `BugsnagEvent` are implemented.

robinmacharg:
Thanks for the detailed feedback - dealt with individually inline.

> One thing to note is that the objective-c-ios example app still seems to be using `event.metaData` and therefore fails to compile

This is an easy one to miss (it's not the first time).  There's a larger piece of work to unify the three Apple platforms in a single project which will simplify the burden going forwards.  The original reason for the separation (Carthage issues) has been resolved.  

For now I'll update the PR template, adding an explicit check for the example app (as well as fixing the example apps).

Source/BugsnagMetadata.m
robinmacharg: Agreed, I'd go further and say uses of "tab" should change across all files, if the vocabulary is changing across all notifiers.

Source/Bugsnag.h
robinmacharg: The task only calls for the single method to be implemented.  Required lower-level methods (in BugsnagMetadata for instance) have been implemented in separate branches as part of other tasks and it makes more sense to implement the missing methods once those have been merged.  I agree with you though, and think that we should realign the existing tasks to the spec, one class/section per task.

fractalwrench: Happy with this approach as long as the task is tracked somewhere.

Source/BugsnagMetadata.m
robinmacharg: No downsides I can see.  Good spot.

Tests/BugsnagTests.m
robinmacharg: Agreed.  Test expanded to cover this case.

iOS/BugsnagTests/Swift Tests/BugsnagMetadataTests.swift
robinmacharg: This - that the `BugsnagMetadata` class is only intended for internal (and therefore ObjC-only) use - has been confirmed in internal discussion.  I'll remove the `NS_SWIFT_NAME`s and associated tests.

iOS/BugsnagTests/Swift Tests/BugsnagSwiftTests.swift
robinmacharg: 👍

PR #466, opened 2020-02-18
feat: Added a mutable Bugsnag.context property

robinmacharg:
##  Goal

Add a mutable `context` property to `Bugsnag`.

## Design

Simple passthrough to the equivalent `BugsnagConfiguration` property.

## Changeset

The `Bugsnag` class.

## Tests

Unit tests that top-level manipulation affected the Configuration value as expected.

### Outstanding Questions

Concurrency concerns are deferred to the underlying Configuration property.

  - [X] Final review

  features)

### DIFF
--- a/Source/BugsnagEvent.m
+++ b/Source/BugsnagEvent.m
@@ -109,6 +109,15 @@ id BSGLoadConfigValue(NSDictionary *report, NSString *valueName) {
     ?: [report valueForKeyPath:fallbackKeypath]; // some custom values are nested
 }
 
+/**
+ * Attempt to find a context (within which the event is being reported)
+ * This can be found in user-set metadata of varying specificity or the global
+ * configuration.  Returns nil if no context can be found.
+ *
+ * @param report A dictionary of report data
+ * @param metadata Additional relevant data
+ * @returns A string context if found, or nil
+ */
 NSString *BSGParseContext(NSDictionary *report, NSDictionary *metadata) {
     id context = [report valueForKeyPath:@"user.overrides.context"];
     if ([context isKindOfClass:[NSString class]])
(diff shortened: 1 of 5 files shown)


Tests/BugsnagTests.m
```diff
@@ -107,4 +107,40 @@ - (void)testGetMetadata {
...
+        XCTAssertEqual([[Bugsnag configuration] context], @"secondContext");
+        [expectation fulfill];
+    }];
+
+    // Test that the context (changed inside the notify block) remains changed
+    [self waitForExpectationsWithTimeout:1.0 handler:^(NSError * _Nullable error) {
+        XCTAssertEqual([[Bugsnag configuration] context], @"secondContext");
+        
+        [Bugsnag notify:exception1 block:^(BugsnagEvent * _Nonnull report) {
+            XCTAssertEqual([[Bugsnag configuration] context], @"secondContext");
```
kattrali: shouldn't this be checking the report context value?

robinmacharg: Good catch; this highlights a situation where `context` was not being set correctly.  Fixed (I think!) in the latest commit.

Source/BugsnagEvent.m
```diff
@@ -109,17 +109,23 @@ id BSGLoadConfigValue(NSDictionary *report, NSString *valueName) {
...
-        return context;
-    context = BSGLoadConfigValue(report, @"context");
-    if ([context isKindOfClass:[NSString class]])
-        return context;
-    return nil;
+    // Select the first non-nil value, in order of specificity
+    id context = [report valueForKeyPath:@"user.overrides.context"]
+        ?: metadata[BSGKeyContext]
+        ?: BSGLoadConfigValue(report, @"context")
+        ?: [[Bugsnag configuration] context];
```
kattrali: This change is a bit tricky since there are two different ways an Event can be constructed.

1. `notify()` was just called and the Event instance has no persisted file. In this case, loading `context` from Configuration makes sense.
2. A persisted report was read from disk and constructed an Event object. In this case, the event may have happened minutes or days ago, and the `context` present in Configuration _now_ is likely not related.

To correctly handle both, I'd recommend setting context for Case 1 within the Event constructor used by `notify()` and leaving this function as it was, as it parses the serialized event file.

robinmacharg: Understood, and thanks for the explanation.  I've reverted `BSGParseContext()` and set `context` in the Event constructor, based on metadata (if present) or Configuration (if present).

PR #467, opened 2020-02-19
refactor(bugsnag): Remove public config accessor

kattrali:
* Add setUser to Bugsnag as a replacement

## Goal

Remove access to the internal configuration object, which should be publicly immutable after calling `start()`. 

## Changeset

* Moved `Bugsnag.configuration()` from the public API to the internal Private header
* Added `Bugsnag.setUser()` to replace setting user through the configuration object after initialization. There are further user changes in a later changeset, and this change already has associated tests (which have been updated to use the new method).

## Tests

* Existing tests cover this change, requiring calls to `Bugsnag.configuration()?.setUser()` to change to `Bugsnag.setUser()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,9 @@ Bugsnag Notifiers on other platforms.
   - Invalid values are logged and ignored.
   [#470](https://github.com/bugsnag/bugsnag-cocoa/pull/470)
 
+* Remove `Bugsnag.configuration()?`. All access to the configuration object
+  should be performed prior to calling `Bugsnag.start()`.
+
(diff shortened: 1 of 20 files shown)


kattrali:
> Private.h isn't exposed in any of the XCode projects that I could see. Is this by design?

No 🤔  I'm not sure what happened there, but I fixed it in #473. Its also not a new file. Didn't bother with the OSX/tvOS projects since those are going away in the release.

> For consistency of naming should it be `BugsnagPrivate.h`?

Not particularly? Its entirely private and contains no public structure named `BugsnagPrivate` or anything.

PR #468, opened 2020-02-19
Remove debugging functions from public API

kattrali:
Short and sweet. removing without replacement, as these weren't intended for public use. 

`setSuspendThreadsForUserReported` will be resurrected in a more useful form as a configuration option in a later changeset.

### DIFF
--- a/Source/Bugsnag.h
+++ b/Source/Bugsnag.h
@@ -200,12 +200,6 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
 
 + (NSDateFormatter *_Nonnull)payloadDateFormatter;
 
-+ (void)setSuspendThreadsForUserReported:(BOOL)suspendThreadsForUserReported;
-+ (void)setReportWhenDebuggerIsAttached:(BOOL)reportWhenDebuggerIsAttached;
-+ (void)setThreadTracingEnabled:(BOOL)threadTracingEnabled;
-+ (void)setWriteBinaryImagesForUserReported:
-    (BOOL)writeBinaryImagesForUserReported;
-
 /**
  * Starts tracking a new session.
  *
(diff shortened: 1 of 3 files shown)


zhaoze1991:
hi @kattrali @robinmacharg  , I was working to upgrade bugsnag from v5 to v6...  
our app has huge perf issue when reporting user errors so I used `setSuspendThreadsForUserReported:NO` to solve the issue.

with this pull request I can't call this API any more. what's the alternative way to disable suspendThreadsForUserReported? 

Thanks!

twometresteve:
> hi @kattrali @robinmacharg , I was working on upgrade bugsnag from v5 to v6...
> our app has huge perf issue when reporting user errors so I used `setSuspendThreadsForUserReported:NO` to solve the issue.
> 
> with this pull request I can't call this API any more. what's the alternative way to disable suspendThreadsForUserReported?
> 
> Thanks!

Hi @zhaoze1991 ,

I believe you want the sendThreads configuration option:
https://docs.bugsnag.com/platforms/ios/configuration-options/#sendthreads

Thanks,
Steve.

zhaoze1991:
hi, @twometresteve thanks for this, set sendThreads can workaround the issue I have.

But shouldn't `suspendThreadsForUserReported` be disabled? it's scary reporting user error could suspend all the threads.

zhaoze1991:
hi, @twometresteve after playaround with `sendThreads`, this is still a perf issue for us if we have multiple errors happened simultaneously .. I can see a noticeable lag in release & debug build.

when set this property to unhandled only, [Bugsna](https://github.com/bugsnag/bugsnag-cocoa/blob/master/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c#L977) still write the offending thread(the thread create the error)


my old settings:
```
Bugsnag.setThreadTracingEnabled(false)
Bugsnag.setWriteBinaryImagesForUserReported(false)
Bugsnag.setSuspendThreadsForUserReported(false)
```

twometresteve:
HI @zhaoze1991 - I think it would be best if you could email our support team on support@bugsnag.com and they will be very happy to help you.  Best regards, Steve.

PR #470, opened 2020-02-25
feat: add method to add metadata section values as a dictionary (restarted)

robinmacharg:
[Note: this PR replaces #456 ([Branch here](1c0b988f53d907085e9fb6d383c426ec3c13624b)) which was proving difficult to merge.  The PR text is largely unchanged.]

## Goal

Adds a `BugsnagMetadata.addMetadataToSection:values:` method to allow merging arbitrary dictionaries with preexisting Metadata.

## Design

Simple fast enumeration loop over provided values with appropriate checks for:

- Validity of Key (String subclass)
- Validity of Value (can be sanitised: a recursive check that all (nested) values are of type Number, String, Array, Dictionary or Null

Conditional addition/deletion depending on whether value is null or not.  There is a check to ensure that empty sections are not created for (failing) invalid values.  Invalid values are logged and ignored.  Delegate callback is called only where there is a material - CRUD - change to the metadata.

## Changeset

- Method added to `BugsnagMetadata`.
- Unit Tests added.

## Tests

- Added Unit tests only.  This change affects the framework API and doesn't change any external aspect.
- CI passes.

### Outstanding Questions

  - [ ] Final review

  features)
- [X] Performance and complexity - are there any cases of unexpected O(n^3) when iterating, recursing, flat mapping, etc?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,9 +49,8 @@ Bugsnag Notifiers on other platforms.
   [#459](https://github.com/bugsnag/bugsnag-cocoa/pull/459)
   
 * Added `Bugsnag.getMetadata(_ section: key:)`
-[#463](https://github.com/bugsnag/bugsnag-cocoa/pull/463)
+  [#463](https://github.com/bugsnag/bugsnag-cocoa/pull/463)
   
-
 * Add a per-Event `apiKey` property.  This defaults to the global 
   `BugsnagConfiguration` value but can be overridden in event passed to the 
   `Bugsnag.notify()` callback.
@@ -90,6 +89,17 @@ Bugsnag Notifiers on other platforms.
   the `BugsnagLogger.h` header.
   [#472](https://github.com/bugsnag/bugsnag-cocoa/pull/472)
   
+* Added a method to allow merging supplied and existing Event metadata.
+  `BugsnagMetadata.addMetadataToSection:values:` allows Event 
+  callbacks to modify Event metadata en-mass.  Supplied metadata should 
+  be a JSON-serializable dictionary.  The resulting Event metadata is the 
+  result of applying the following rules to the existing metadata for each supplied
+  value:
+  - Non-null values replace any existing key/value pair. 
+  - Null values remove a key/value pair.  
+  - Invalid values are logged and ignored.
+  [#470](https://github.com/bugsnag/bugsnag-cocoa/pull/470)
+
 ## Bug fixes
 
 * Fix possible report corruption when using `notify()` from multiple threads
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,6 @@
 Changelog
 =========
 
-## TBD
-
```
kattrali: No need to remove this. In fact, its [automatically replaced during the release process with the new version number and date](https://github.com/bugsnag/bugsnag-cocoa/blob/e95c825c64cb359ba87dfec32dfa66a11ce47dc9/Makefile#L67).

CHANGELOG.md
```diff
@@ -78,6 +76,10 @@ Bugsnag Notifiers on other platforms.
 * Add metadata accessor methods to `BugsnagEvent`
   [#465](https://github.com/bugsnag/bugsnag-cocoa/pull/465)
 
+* Added a method allowing merging arbitrary (JSON-encodable) dictionaries with 
+  an existing Event Metadata instance (`BugsnagMetadata.addMetadataToSection:values:`)
```
kattrali: What does this mean for users exactly? Saying that it "merges" dicts with Event Metadata could be unclear about what will happen when this method is called. Can you provide a code example here? The headerdoc for the function is fairly clear.

Source/BugsnagMetadata.m
```diff
@@ -94,33 +94,107 @@ - (NSDictionary *)toDictionary {
...
             id cleanedValue = BSGSanitizeObject(value);
             if (cleanedValue) {
-                NSDictionary *section = [self getMetadata:sectionName];
+                // Value is OK, try and set it
+                NSMutableDictionary *section = [self getMetadata:tabName];
                 if (!section) {
-                    section = [NSMutableDictionary new];
-                    [[self dictionary] setObject:section forKey:sectionName];
+                    section = [@{} mutableCopy];
+                    [[self dictionary] setObject:section forKey:tabName];
```
kattrali: Not necessary bad, but why the change from making a dictionary using the constructor to using a copy?

CHANGELOG.md
robinmacharg: TIL, thanks.  Added back in.

CHANGELOG.md
robinmacharg: I've expanded the message to have what I hope is enough clarity.  If you still think it would benefit from a code sample let me know.

Source/BugsnagMetadata.m
robinmacharg: I'm not actually sure why I made the change.  I agree, though, why jump through an unnecessary hoop?  Changed back.

PR #472, opened 2020-02-27
feat: Consolidate internal logging configuration

robinmacharg:
## Goal

Consolidate internal logging configuration to only use the `BSG_LOG_LEVEL` preprocessor macro.

## Design

The smallest change possible: an import and a variable change.  Additional first-pass documentation.

## Changeset

```
BSG_KSLogger.h
BugsnagLogger.h
```

## Tests

Manual testing with the example apps, repeated running of `pod install`.  How to automate testing in an appropriate amount of time was not obvious.

### Outstanding Questions

- Platform Documentation could do with being improved (but this can be done as a separate task; raised internally)
- Additional docs (e.g. Carthage, Manual) are probably required for completeness (ditto)
- While this has been tested to a reasonable extent, the #ifdef macros are a little soupy; are there any logging edge cases I've missed?

  - [X] Final review

  features)
- [X] Usage friction - is the proposed change in usage cumbersome or complicated?


Source/BugsnagLogger.h
```diff
@@ -1,12 +1,38 @@
+/**
+ * Higher-level user-accessible BugsnaglLogging configuration.  Controls how
```
fractalwrench: BugsnaglLogging -> Bugsnag Logging

Source/BugsnagLogger.h
```diff
@@ -1,12 +1,38 @@
...
+ *         rep.pods_project.targets.each do |target|
+ *             if target.name == "Bugsnag"
+ *                 target.build_configurations.each do |config|
+ *                     config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] ||= ['$(inherited)', 'BSG_LOG_LEVEL=BSG_LOGLEVEL_INFO']
+ *                 end
+ *             end
+ *         end
+ *     end
+ *
+ * Change the value of `BSG_LOG_LEVEL` to one of the levels given below and run `pod install`.
```
fractalwrench: Is it possible to alter the log level outside of using Cocoapods? If so it'd be good to document how to achieve this in a package-manager agnostic way.

robinmacharg: Agreed, as noted in the PR. An internal ticket has been raised to address this. While the mechanism (`BSG_LOG_LEVEL`) __is__ applicable across installation methods (Cocoapods, Carthage and Manually), documenting the specifics for all methods of library integration is left for a later date.

fractalwrench: As long as a separate task has been raised and it will be addressed as part of the Notifier Spec project I'm happy to leave this for now.

PR #477, opened 2020-02-28
(feat) Implement configurable enabledErrorTypes

robinmacharg:
## Goal

To allow users to configure the types of event that are reported.

## Design

This is implemented as a bitfield on `BugsnagConfiguration`.  `BugsnagSentry` translates this to equivalent KSCrash values. The OOM value is examined in `BugsnagNotifier`. The bitfield mapping has been broken out to make it testable.

## Changeset

As above, `BugsnagConfiguration`, `BugsnagSentry` and `BugsnagNotifier`.

## Tests

This is difficult to test at unit-level.  What can be has been.  E2E tests better test the behaviour and this is where the effort has been concentrated.

### Outstanding Questions

* E2E Testing is extensive but I have lingering doubts over whether it's __comprehensive__.  There's ambiguous behaviour around e.g. Mach errors being converted to Signals.

  - [X] Final review

  features)
- [X] Usage friction - is the proposed change in usage cumbersome or complicated?


robinmacharg:
Note to self: I need to check e2e on Mac vs iOS.

Source/BugsnagConfiguration.m
```diff
@@ -336,4 +342,6 @@ - (void)setMaxBreadcrumbs:(NSUInteger)capacity {
     self.breadcrumbs.capacity = capacity;
 }
 
+@synthesize enabledErrorTypes = _enabledErrorTypes;
```
kattrali: This should not be necessary since there's no overridden setter and getter.

Source/BugsnagNotifier.m
```diff
@@ -399,14 +399,19 @@ - (void)start {
 
     _started = YES;
     // autoDetectErrors disables all unhandled event reporting
-    BOOL configuredToReportOOMs = self.configuration.reportOOMs && self.configuration.autoDetectErrors;
+    BOOL configuredToReportOOMs = self.configuration.reportOOMs
```
kattrali: The new error types should replace `reportOOMs`, right?

UPGRADING.md
```diff
@@ -45,6 +45,8 @@ Swift:
 - config.add(beforeSession:)
 + config.onSessionBlocks
 + config.add(onSession:)
+
++ config.enabledErrorTypes
```
kattrali:
```suggestion
- config.reportOOMs
+ config.enabledErrorTypes
```

Source/BugsnagConfiguration.m
```diff
@@ -106,6 +106,12 @@ -(instancetype)initWithApiKey:(NSString *)apiKey
     _breadcrumbs = [BugsnagBreadcrumbs new];
     _automaticallyCollectBreadcrumbs = YES;
     _autoTrackSessions = YES;
+    // Default to recording all error types
+    _enabledErrorTypes = BSGErrorTypesOOMs
```
kattrali: Enabling OOM detection by default should be guarded to only happen in release builds, to avoid triggering the heuristic when killing/restarting an app in Xcode or similar.

Source/BugsnagNotifier.m
robinmacharg: I wasn't sure.  Do we want to deprecate this in the first instance rather than excise it completely?

kattrali: That would be fine too, though the setter/getter would need to proxy to enabled error types rather than leaving both in place. In the current configuration, there isn't a way to determine precedence since the property can't be unset and is off by default in some cases.

robinmacharg: Excised as an exercise in simplicity.

PR #482, opened 2020-03-10
(feat)Add callback-free methods to leave breadcrumbs

robinmacharg:
## Goal

Add additional methods to leave complex breadcrumbs without requiring a callback to be provided.

## Design

Two top-level `Bugsnag` methods largely following the "if started..." pattern established previously.

## Changeset

Primarily the `Bugsnag` class and header, along with unit tests in ObjC anbd Swift.  Several other minor typographical changes are included but can be ignored.

## Tests

It was only required to check whether breadcrumbs were set correctly and so unit testing sufficed.  A simple test for Swift exposure was also included.

### Outstanding Questions

  - [X] Final review

  features)

### DIFF
--- a/Source/Bugsnag.h
+++ b/Source/Bugsnag.h
@@ -162,7 +162,7 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
  * Leave a "breadcrumb" log message, representing an action that occurred
  * in your app, to aid with debugging.
  *
- * @param message  the log message to leave (max 140 chars)
+ * @param message  the log message to leave
  */
 + (void)leaveBreadcrumbWithMessage:(NSString *_Nonnull)message;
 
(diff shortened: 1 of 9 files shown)


Source/Bugsnag.h
```diff
@@ -183,6 +183,33 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
...
+/**
+ * Leave a "breadcrumb" log message, representing an action that occurred
+ * in your app.
+ *
+ * @param message The log message to leave
+ * @param metadata Additional metadata included with the breadcrumb
+ *
+ * , to aid with debugging, along with additional metadata.
+ */
++ (void)leaveBreadcrumbWithMessage:(NSString *_Nonnull)message
```
fractalwrench: The notifier spec states that `metadata` and `type` are optional parameters. I would interpret that by implementing the following two methods:

`leaveBreadcrumb:message`
`leaveBreadcrumb:message:metadata:type`

So I think it'd be ok to remove this particular method variant, unless there are strong reasons for including it.

Source/Bugsnag.h
```diff
@@ -183,6 +183,33 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
  */
 + (void)leaveBreadcrumbForNotificationName:(NSString *_Nonnull)notificationName;
```
fractalwrench: It may be outside the scope of this PR, but it'd be good to remove any non-compliant methods for leaving breadcrumbs such as these from our public interface. This could be done either by removing them entirely, or just removing them from the public header.

robinmacharg: This will be handled as a separate piece of work.

Source/Bugsnag.h
robinmacharg: As discussed this will be removed.

PR #484, opened 2020-03-12
Implement new plugin interface

fractalwrench:
## Goal

Implements the new plugin interface from the notifier spec. This allows the addition of user-defined plugins up-front before Bugsnag is configured.

This changeset has been adapted from the Android implementation: https://github.com/bugsnag/bugsnag-android/pull/815

## Changeset

- Added ability to add plugins to `BugsnagConfiguration`, which are deep-copied on Bugsnag initialisation
- Extracted the majority of the plugin loading logic to the `BugsnagPluginClient` class to reduce the responsibilities of `Bugsnag`.
- Initialised `BugsnagPluginClient` within `BugsnagClient` and loaded plugins when `[client start]` is invoked
- Removed legacy non-compliant methods on `BugsnagPlugin`

## Tests

- Added unit tests to verify the `PluginClient` and `Configuration` changes
- Updated existing E2E coverage for custom plugin creation

### DIFF
--- a/Source/BugsnagClient.m
+++ b/Source/BugsnagClient.m
@@ -275,7 +275,7 @@ void BSGWriteSessionCrashData(BugsnagSession *session) {
         [self metadataChanged:self.configuration.metadata];
         [self metadataChanged:self.configuration.config];
         [self metadataChanged:self.state];
-        _pluginClient = [[BugsnagPluginClient alloc] initWithPlugins:self.configuration.plugins];
+        self.pluginClient = [[BugsnagPluginClient alloc] initWithPlugins:self.configuration.plugins];
     }
     return self;
 }
--- a/Source/BugsnagClient.m
+++ b/Source/BugsnagClient.m
@@ -33,6 +33,7 @@
 #import "BugsnagLogger.h"
 #import "BugsnagKeys.h"
 #import "BugsnagSessionTracker.h"
+#import "BugsnagPluginClient.h"
 #import "BSGOutOfMemoryWatchdog.h"
 #import "BSG_RFC3339DateTool.h"
 #import "BSG_KSCrashC.h"
@@ -206,7 +207,12 @@ void BSGWriteSessionCrashData(BugsnagSession *session) {
 @property(nonatomic, strong) BugsnagErrorReportApiClient *errorReportApiClient;
 @property(nonatomic, strong, readwrite) BugsnagSessionTracker *sessionTracker;
 @property (nonatomic, strong) BSGOutOfMemoryWatchdog *oomWatchdog;
+@property (nonatomic, strong) BugsnagPluginClient *pluginClient;
 @property (nonatomic) BOOL appCrashedLastLaunch;
+@end
+
+@interface BugsnagConfiguration ()
+@property(nonatomic, readwrite, strong) NSMutableSet *plugins;
 @end
 
 @implementation BugsnagClient
@@ -269,6 +275,7 @@ void BSGWriteSessionCrashData(BugsnagSession *session) {
         [self metadataChanged:self.configuration.metadata];
         [self metadataChanged:self.configuration.config];
         [self metadataChanged:self.state];
+        _pluginClient = [[BugsnagPluginClient alloc] initWithPlugins:self.configuration.plugins];
     }
     return self;
 }
@@ -342,6 +349,7 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
 }
 
 - (void)start {
+    [self.pluginClient loadPlugins];
     [self.crashSentry install:self.configuration
                     apiClient:self.errorReportApiClient
                       onCrash:&BSSerializeDataCrashHandler];
--- a/Source/BugsnagClient.m
+++ b/Source/BugsnagClient.m
@@ -349,7 +349,6 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
 }
 
 - (void)start {
-    [self.pluginClient loadPlugins];
     [self.crashSentry install:self.configuration
                     apiClient:self.errorReportApiClient
                       onCrash:&BSSerializeDataCrashHandler];
@@ -431,6 +430,7 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
 
     // notification not received in time on initial startup, so trigger manually
     [self willEnterForeground:self];
+    [self.pluginClient loadPlugins];
 }
 
 - (void)addTerminationObserver:(NSString *)name {
--- a/Source/BugsnagPlugin.h
+++ b/Source/BugsnagPlugin.h
@@ -10,7 +10,15 @@
 
 @required
 
+/**
+ * Loads a plugin with the given Client. When this method is invoked the plugin should
+ * activate its behaviour - for example, by capturing an additional source of errors.
+*/
 - (void)load;
+/**
+ * Unloads a plugin. When this is invoked the plugin should cease all custom behaviour and
+ * restore the application to its unloaded state.
+ */
 - (void)unload;
 
 @end
--- a/Source/BugsnagPlugin.h
+++ b/Source/BugsnagPlugin.h
@@ -1,6 +1,8 @@
 #ifndef BugsnagPlugin_h
 #define BugsnagPlugin_h
 
+@class Bugsnag;
+
 /**
  * Internal interface for adding custom behavior
  */
@@ -8,9 +10,8 @@
 
 @required
 
-- (BOOL)isStarted;
-
-- (void)start;
+- (void)load;
+- (void)unload;
 
 @end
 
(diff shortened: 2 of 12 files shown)


Source/BugsnagClient.m
```diff
@@ -206,9 +207,14 @@ @interface BugsnagClient ()
...
 @property(nonatomic, strong, readwrite) BugsnagSessionTracker *sessionTracker;
 @property (nonatomic, strong) BSGOutOfMemoryWatchdog *oomWatchdog;
+@property (nonatomic, strong) BugsnagPluginClient *pluginClient;
 @property (nonatomic) BOOL appCrashedLastLaunch;
 @end
 
+@interface BugsnagConfiguration ()
+@property(nonatomic, readwrite, strong) NSMutableSet *plugins;
+@end
+
```
tomlongridge: Is this the correct place for the declaration? Why not in `BugsnagConfiguration.h`?

Source/BugsnagClient.m
```diff
@@ -269,6 +275,7 @@ - (id)initWithConfiguration:(BugsnagConfiguration *)initConfiguration {
         [self metadataChanged:self.configuration.metadata];
         [self metadataChanged:self.configuration.config];
         [self metadataChanged:self.state];
+        _pluginClient = [[BugsnagPluginClient alloc] initWithPlugins:self.configuration.plugins];
```
tomlongridge: Any reason we shouldn't use `[self pluginClient...` to keep this consistent with other properties such as `SessionTracker` which have similar visibility/usage.

Source/BugsnagPlugin.h
```diff
@@ -1,16 +1,17 @@
...
  * Internal interface for adding custom behavior
  */
 @protocol BugsnagPlugin <NSObject>
 
 @required
 
-- (BOOL)isStarted;
-
-- (void)start;
+- (void)load;
```
tomlongridge: This should take a client instance as a parameter

tomlongridge:
```suggestion
/**
 * Loads a plugin with the given Client. When this method is invoked the plugin should
 * activate its behaviour - for example, by capturing an additional source of errors.
*/
- (void)load;
```

Source/BugsnagPlugin.h
```diff
@@ -1,16 +1,17 @@
...
+- (void)unload;
```
tomlongridge:
```suggestion
/**
 * Unloads a plugin. When this is invoked the plugin should cease all custom behaviour and
 * restore the application to its unloaded state.
 */
- (void)unload;
```

Source/BugsnagPluginClient.m
```diff
@@ -0,0 +1,31 @@
...
+- (instancetype _Nonnull)initWithPlugins:(NSMutableSet<id<BugsnagPlugin>> *_Nonnull)plugins {
+    if (self = [super init]) {
+        _plugins = [NSSet setWithSet:plugins];
+    }
+    return self;
+}
+
+- (void)loadPlugins {
+    for (id<BugsnagPlugin> plugin in self.plugins) {
+        [plugin load];
```
tomlongridge: Can we catch any exception here to prevent it affecting the client start? (This is what we do in Android.)

Source/BugsnagClient.m
fractalwrench: This is a [class extension](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html#//apple_ref/doc/uid/TP40011210-CH6-SW3) which is used in Objective-C to hide implementation details. `BugsnagClient` will eventually need to be a public class and when it is, this property should be hidden.

Source/BugsnagClient.m
fractalwrench: I'll defer to @robinmacharg on this one as I believe there's a lot of nuance and differing opinion on when `self` should be used vs accessing the property directly, and it's something we should standardise on.

Source/BugsnagPlugin.h
fractalwrench: `BugsnagClient` is not publicly accessible and there's no point passing in the `Bugsnag` facade because it can't be instantiated. I agree that the client would ideally be passed in as a parameter to this method if we were following the spec, but the amount of work this would take means it's something we would need to address separately. This is definitely something we want to discuss once the current crop of tickets have been implemented.

tomlongridge: Will you be able to write your RN plugin from this? Will it just use the `Bugsnag` static to access the client?

fractalwrench: Yep, the React Native plugin will have to use the `Bugsnag` facade to start off with. If the deviations with `BugsnagClient` are addressed then it should be possible to alter the plugin implementation later on.

Source/BugsnagClient.m
tomlongridge: I guess I was looking from a consistency point of view - this seems to be different to how we're accessing other properties that aren't intended for public use (e.g. SessionTracker)

fractalwrench: I'll update this to be consistent for now, then we can always discuss whether to use `self` or direct field access later on.

Source/BugsnagPluginClient.m
fractalwrench: Updated this to catch exceptions thrown when loading a plugin, although this is less likely compared to Android as exceptions are generally avoided in Objective-C.

Source/BugsnagClient.m
robinmacharg: The short of it is that using `_property` accesses the property directly. Using `self` uses synthetic getters/setters that we could, in theory, override in the future if we needed to.  ObjC gives us the getters/setters for free, unseen.  There's potentially a minuscule performance hit but opinion seems to be that this is probably optimised out and so `self.property` is preferable.

Source/BugsnagClient.m
robinmacharg: Put another way: this declaration gives `BugsnagClient` visibility of `BugsnagConfiguration`'s `plugins` property.

Source/BugsnagClient.m
```diff
@@ -206,9 +207,14 @@ @interface BugsnagClient ()
 @property(nonatomic, strong) BugsnagErrorReportApiClient *errorReportApiClient;
 @property(nonatomic, strong, readwrite) BugsnagSessionTracker *sessionTracker;
 @property (nonatomic, strong) BSGOutOfMemoryWatchdog *oomWatchdog;
+@property (nonatomic, strong) BugsnagPluginClient *pluginClient;
 @property (nonatomic) BOOL appCrashedLastLaunch;
 @end
 
+@interface BugsnagConfiguration ()
+@property(nonatomic, readwrite, strong) NSMutableSet *plugins;
```
robinmacharg: Are we confident that plugin load ordering won't be an issue?  If we later find that it is this could be replaced with an ordered set implementation.

Source/BugsnagPluginClient.m
```diff
@@ -0,0 +1,36 @@
...
+    return self;
+}
+
+- (void)loadPlugins {
+    for (id<BugsnagPlugin> plugin in self.plugins) {
+        @try {
+            [plugin load];
+        } @catch (NSException *exception) {
+            bsg_log_err(@"Failed to load plugin %@, continuing with initialisation.", plugin);
+        }
```
robinmacharg: This obviously only catches a subset of possible errors.  Are we happy with the compromise of loading plugins before the full set of Bugsnag error trapping is in place?

Tests/BugsnagPluginTest.m
```diff
@@ -0,0 +1,49 @@
...
+}
+
+- (void)testPluginLoaded {
+    FakePlugin *plugin = [FakePlugin new];
+    BugsnagConfiguration *config = [[BugsnagConfiguration alloc] initWithApiKey:DUMMY_APIKEY_32CHAR_1 error:nil];
+    [config addPlugin:plugin];
+    [Bugsnag startBugsnagWithConfiguration:config];
+    XCTAssertTrue(plugin.loaded);
+}
+@end
```
robinmacharg: The `@catch` clause in plugin startup could use a test, maybe a failing expectation?

Source/BugsnagClient.m
fractalwrench: Currently bugsnag-cocoa only uses plugins for cocos2dx - that won't be used at the same time as react-native. In future we could update the set implementation or make the plugins resilient to handling different load orders.

Tests/BugsnagPluginTest.m
fractalwrench: Added an additional test case to cover this.

Source/BugsnagPluginClient.m
fractalwrench: After discussion attempted to reduce this compromise by loading the plugins at the very end of `[BugsnagClient start]`, and adding an additional test to verify that exceptions are caught when loading plugins.

PR #485, opened 2020-03-13
Add onSendBlock methods

robinmacharg:
## Goal

Add missing `add/remove/clearOnSendBlock()` methods to `Bugsnag` and `BugsnagConfiguration`.  These are used in this notifier in lieu of equivalent `onError` methods in others due to the tight requirements of code run in error handlers (e.g. no ObjC, async etc).

## Design

Straightforward exposure of access to the `BugsnagConfiguration` onSendBlocks array.

## Changeset

- `Bugsnag`, `BugsnagConfiguration` and associated tests.

- Some small non-functional rearrangement refactoring of headers has also been done.

## Tests

Unit tests.  Not tested prior to `start` being called.

### Outstanding Questions

  - [X] Final review

  features)

### DIFF
--- a/Source/Bugsnag.h
+++ b/Source/Bugsnag.h
@@ -63,7 +63,9 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
  */
 + (BOOL)appDidCrashLastLaunch;
 
+// =============================================================================
 // MARK: - Notify
+// =============================================================================
 
 /** Send a custom or caught exception to Bugsnag.
  *
@@ -156,7 +158,9 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
                        value:(id _Nullable)value
     NS_SWIFT_NAME(addMetadata(_:key:value:));
 
+// =============================================================================
 // MARK: - Breadcrumbs
+// =============================================================================
 
 /**
  * Leave a "breadcrumb" log message, representing an action that occurred
@@ -212,7 +216,9 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
 + (void)setBreadcrumbCapacity:(NSUInteger)capacity
         __deprecated_msg("Use [BugsnagConfiguration setMaxBreadcrumbs:] instead");
 
+// =============================================================================
 // MARK: - Session
+// =============================================================================
 
 /**
  * Starts tracking a new session.
@@ -291,6 +297,9 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
 + (id _Nullable )getMetadata:(NSString *_Nonnull)section key:(NSString *_Nonnull)key
     NS_SWIFT_NAME(getMetadata(_:key:));
 
+// =============================================================================
+// MARK: - Other methods
+// =============================================================================
 /**
  * Remove a key/value from a named matadata section.  If either the section or the
  * key do not exist no action will occur.
@@ -344,5 +353,7 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
 + (void)setUser:(NSString *_Nullable)userId
        withName:(NSString *_Nullable)name
        andEmail:(NSString *_Nullable)email;
+
++ (NSDateFormatter *_Nonnull)payloadDateFormatter;
 
 @end
--- a/Source/Bugsnag.h
+++ b/Source/Bugsnag.h
@@ -371,9 +371,4 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
  */
 + (void)removeOnSendBlock:(BugsnagOnSendBlock _Nonnull)block;
 
-/**
- * Clear all onSend callbacks
- */
-+ (void)clearOnSendBlocks;
-
 @end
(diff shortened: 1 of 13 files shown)


Source/BugsnagConfiguration.h
```diff
@@ -61,8 +61,7 @@ typedef void (^BugsnagOnErrorBlock)(BugsnagEvent *_Nonnull report);
  *
  *  @return YES if the report should be sent
  */
-typedef bool (^BugsnagOnSendBlock)(NSDictionary *_Nonnull rawEventData,
-                                       BugsnagEvent *_Nonnull reports);
+typedef bool (^BugsnagOnSendBlock)(NSDictionary *_Nonnull rawEventData, BugsnagEvent *_Nonnull reports);
```
fractalwrench: I think it'd make sense to remove `rawEventData` and rename `reports` as `event`, as it could be a bit confusing having two parameters. In the past users have edited the `rawEventData` and been surprised that changes are not reflected in reports sent to the bugsnag dashboard.

Source/BugsnagConfiguration.m
```diff
@@ -185,10 +185,27 @@ - (void)setUserMetadataFromUser:(BugsnagUser *)user {
     [self.metadata addAttribute:BSGKeyEmail withValue:user.emailAddress toTabWithName:BSGKeyUser];
 }
```
fractalwrench: The `onSendBlocks` property is publicly accessible. This (and the other callback arrays) should be hidden as part of the internal implementation and accessed via a class extension or similar mechanism.

Source/BugsnagConfiguration.m
```diff
@@ -185,10 +185,27 @@ - (void)setUserMetadataFromUser:(BugsnagUser *)user {
...
+// MARK: - onSendBlock
+// =============================================================================
+
 - (void)addOnSendBlock:(BugsnagOnSendBlock)block {
     [(NSMutableArray *)self.onSendBlocks addObject:[block copy]];
 }
 
+- (void)removeOnSendBlock:(BugsnagOnSendBlock _Nonnull )block
+{
+    [(NSMutableArray *)self.onSendBlocks removeObject:block];
```
fractalwrench: This is more for my curiosity, but I'd be interested to know how this works, as `addOnSendBlock` makes a copy of the block whereas this is passed the original parameter. The test covering this scenario does seem to pass however.

Source/Bugsnag.h
```diff
@@ -345,4 +352,28 @@ static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
...
+ * Remove an onSend callback, if it exists
+ *
+ * @param block The block to remove
+ */
++ (void)removeOnSendBlock:(BugsnagOnSendBlock _Nonnull)block;
+
+/**
+ * Clear all onSend callbacks
+ */
++ (void)clearOnSendBlocks;
```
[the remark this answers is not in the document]
robinmacharg: Agreed, and yes, it does reduce the surface area, although part of me does like having the convenience methods around.

robinmacharg:
> is this functionality defined anywhere in the notifier spec?

No, it's a divergence.  Due to the limitations of the environment during a crash (no async, or ObjC etc) we can't implement `onError` like the other notifiers.  `onSend` is the best we can do.  There is a separate task to document the divergence.

fractalwrench:
>There is a separate task to document the divergence.

It would be preferable if the spec was up-to-date before we consider merging this PR.

Without a written specification that has been agreed upon, we run the risk of having to rework this interface or even scrap it entirely in favour of another solution. Documenting and scrutinising the requirements down beforehand reduces this risk.

Source/BugsnagConfiguration.h
robinmacharg: Thanks for the background.  Happy to change it, if we're happy that nothing exclusively in the raw data could influence the return value.

tomlongridge:
> Without a written specification that has been agreed upon, we run the risk of having to rework this interface or even scrap it entirely in favour of another solution. Documenting and scrutinising the requirements down beforehand reduces this risk.

I've raised an issue, but I'd seen this more of a rename to get it more like the spec until such time as we have resolved whether the concept of OnSend is a Cocoa deviation or something we'll introduce as a broader concept.

I'm happy to merge this once we have the add and remove methods (with no rawEventData parameter) but I'll kick off discussion on the issue to make sure it's covered before we release.

Source/BugsnagConfiguration.m
robinmacharg: Agreed.  Done for `onSend` and `onSession`.

Source/BugsnagConfiguration.m
robinmacharg: I’ve done a little digging; I think I understand it well enough to explain.  Failing that there’s a link at the end, and please correct me if I’m wrong.  Blocks are created initially on the stack, unlike other ObjC objects. The copy effectively _moves_ it to the heap, allowing it to be retained and used after the enclosing scope has exited.  The object pointer is updated.  So the usage in `remove()` will remove the correct - heap-stored - block from the array.  http://www.friday.com/bbum/2009/08/29/blocks-tips-tricks/

PR #486, opened 2020-03-18
(feat) Enhanced orientation breadcrumbs with explicit "from" and "to"

robinmacharg:
## Goal

Report both the old and new orientation values as metadata. Match the casing for the breadcrumb message.

## Design

On device orientation change notification, record the current device orientation to be used the next time orientation changes.  When both current and previous orientations are non-nil and differ record an appropriate breadcrumb.

## Changeset

The change is located in `BugsnagClient`.

## Tests

This is difficult to test.  For Unit testing mocking would allow the UIDevice to be stubbed out and simulate a rotation.  For CI/e2e tests the possibility exists (but has not been proven to work) of sending a rotation keypress (Cmd-Left) to the simulator.

The implementation is simple and in the interests of time this feature is left as a manual test until mocking libraries have been added.

### Outstanding Questions

  - [X] Final review

  features)


Source/BugsnagClient.m
```diff
@@ -811,14 +815,26 @@ - (void)orientationChanged:(NSNotification *)notif {
...
+        if ([orientation isEqualToString:metadata[BSGKeyOrientation]])
             return; // ignore duplicate orientation event
-        }
     }
 
-    [[self state] addAttribute:BSGKeyOrientation
-                     withValue:orientation
-                 toTabWithName:BSGKeyDeviceState];
+    // It's not a change
+    if ([orientation isEqualToString:lastOrientation])
```
fractalwrench: nit: could braces be added to new if statements throughout this file?

I don't know if we actually have any official code style guide for Cocoa, so maybe that's something to discuss.

Source/BugsnagClient.m
```diff
@@ -84,6 +84,10 @@
 static NSUInteger handledCount;
 static NSUInteger unhandledCount;
 static bool hasRecordedSessions;
+#if TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
+// The previous device orientation - iOS only
+static NSString *lastOrientation = NULL;
```
fractalwrench: Can this be declared as a property instead? This would make it easier to get to a point where it'd be possible to instantiate multiple `BugsnagClient` objects, and also simplify any future extraction of this logic to a separate class.

Source/BugsnagClient.m
```diff
@@ -811,14 +815,26 @@ - (void)orientationChanged:(NSNotification *)notif {
...
 
-    [[self state] addAttribute:BSGKeyOrientation
-                     withValue:orientation
-                 toTabWithName:BSGKeyDeviceState];
+    // It's not a change
+    if ([orientation isEqualToString:lastOrientation])
+        return;
+
+    // We previously had an orientation
+    if (lastOrientation) {
```
fractalwrench: This can be simplified by calculating the `lastOrientation` when `BugsnagClient` is initialised (assuming that the return value of `[UIDevice currentDevice].orientation` is non-null)

Source/BugsnagClient.m
robinmacharg: We don't.  I've been taking the arbitrary decision that single line conditionals don\t need the baggage of a closing brace.  I can see the other side of the argument.  I'll leave as is and raise a ticket to come up with a definitive style guide.

snmaynard: The style guide for C code is pretty universally always include braces around if statements because it's easy to miss that they are missing when adding future code.

Source/BugsnagClient.m
robinmacharg: We can't assume this.  Acording to [the docs](https://developer.apple.com/documentation/uikit/uidevice/1620041-begingeneratingdeviceorientation?language=objc) `beginGeneratingDeviceOrientationNotifications()` needs called.  We do this in `BugsnagClient.start()`.  As the code stands we call `BugsnagClient.initWithConfiguration()` first, then `BugsnagClient.start()`.  We could take the golden path and short-circuit on the first call instead.

Source/BugsnagClient.m
robinmacharg: I'm assuming you mean class property?  Yes, that makes sense.

fractalwrench: Yep, that's correct

Source/BugsnagClient.m
fractalwrench: Isn't it possible to move where `beginGeneratingDeviceOrientationNotifications()` is called to make calculation of an initial value feasible?

Source/BugsnagClient.m
robinmacharg: Task created to agree on Cocoa style.  Braces will be readded.

Source/BugsnagClient.m
robinmacharg: Yes, but I don't believe that there's any guarantee that the orientation has been determined at any specific time.  A test implementation bears this out.  One way round this deficit would be to use `[[UIApplication sharedApplication] statusBarOrientation]` but this does not allow for face up/down determination.  If the goal is to have a breadcrumb with valid from/to values I think the current approach with an early exit is the most consistent.

fractalwrench: Happy enough to go with the current implementation if it's not always feasible to get an initial value.

Source/BugsnagClient.m
```diff
@@ -774,51 +829,46 @@ - (void)batteryChanged:(NSNotification *)notif {
...
         }
     }
-
-    [[self state] addAttribute:BSGKeyOrientation
-                     withValue:orientation
+    
+    // Record the breadcrumb
+    [[self state] addAttribute:BSGKeyOrientationChange
+                     withValue:@{@"from" : _lastOrientation,
+                                 @"to" : orientation}
```
tomlongridge: This is the call to add metadata - think won't be adding it to the breadcrumb (as far as I can see), this seems to be done through `sendBreadcrumbForNotification`.

Source/BugsnagKeys.h
```diff
@@ -43,6 +43,7 @@ static NSString *const BSGKeyLabel = @"label";
 static NSString *const BSGKeySeverityReason = @"severityReason";
 static NSString *const BSGKeyLogLevel = @"logLevel";
 static NSString *const BSGKeyOrientation = @"orientation";
+static NSString *const BSGKeyOrientationChange = @"Orientation change";
```
tomlongridge: I'm not sure this is really a key - I think it should use the notification name.

Source/BugsnagClient.m
```diff
@@ -42,12 +42,6 @@
 #import "BSG_KSSystemInfo.h"
 #import "BSG_KSMach.h"
 
-#if TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
```
fractalwrench: Just wanted to check why it's necessary to move these imports to the header file?

Source/BugsnagClient.m
```diff
@@ -217,6 +252,23 @@ @interface BugsnagConfiguration ()
 
 @implementation BugsnagClient
 
+#if TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
+/**
+ * Storage for the device orientation.  It is "last" whenever an orientation change is received
+ */
+static NSString *_lastOrientation = nil;
```
fractalwrench: Why does this need to be static? Is it possible to just assign to `self.lastOrientation` declared in the class extension instead?

Source/BugsnagClient.m
```diff
@@ -774,51 +829,46 @@ - (void)batteryChanged:(NSNotification *)notif {
...
+    else if (!_lastOrientation || [orientation isEqualToString:_lastOrientation]) {
+        _lastOrientation = orientation;
+        return;
     }
 
-    NSDictionary *lastBreadcrumb =
-        [[self.configuration.breadcrumbs arrayValue] lastObject];
-    NSString *orientationNotifName =
-        BSGBreadcrumbNameForNotificationName(notif.name);
+    NSDictionary *lastBreadcrumb = [[self.configuration.breadcrumbs arrayValue] lastObject];
```
fractalwrench: This feels error-prone - users could choose to reject the breadcrumb via an `OnBreadcrumb` callback. It should also be possible to just use `self.lastOrientation` to remove duplicate orientation events, unless I've missed some additional complexity here.

Source/BugsnagKeys.h
robinmacharg: It's not a key per-se, but it is a constant string.  `keys.h` was an unobtrusive place to put it.  As per other constant strings scattered throughout the library I'll put it in `client.m`.

Source/BugsnagClient.m
robinmacharg: Good spot.  Hangover from trying to make `orientationChanged` testable.  Moved back.

Source/BugsnagClient.m
robinmacharg: Previous discussion agreed that this should be a class property.  My understanding is that class properties need static storage.  Am I misunderstanding something?

fractalwrench: Apologies - my terminology might be a little off as it's a while since I've used Objective-C in anger. The intent behind my review comment is that this would be defined in a [class extension](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html#//apple_ref/doc/uid/TP40011210-CH6-SW3), similar to this:

```
// within BugsnagClient.m

@interface BugsnagClient()
@property NSString *lastOrientation;
@end
```

This would avoid exposing the property publicly. Adding new static variables will make it harder to instantiate multiple objects of `BugsnagClient` - which is something we may want to do eventually. As far as I'm aware using a class extension wouldn't require static storage?

Source/BugsnagClient.m
robinmacharg: Doh.  Resolved after discussion.

Source/BugsnagKeys.h
robinmacharg: Resolved after discussion.  Removed, replaced by a key in `notificationNameMap`.

Source/BugsnagClient.m
robinmacharg: Restructured the entire method after discussion.

Source/BugsnagClient.m
robinmacharg: Discussed offline, moved from class to instance-level storage.

PR #487, opened 2020-03-18
Add observers for state change events

fractalwrench:
## Goal

The React Native layer needs to be informed when the state of the user/context/metadata change on Cocoa, so that it can update the information stored in the JS layer.

This changeset makes the properties observable and allows observers to pass a single block which will be invoked as a callback when any of the properties change.

This loosely follows the Android PR and design: https://github.com/bugsnag/bugsnag-android/pull/827

## Changeset
- Removed `BugsnagMetadataDelegate` in favour of observers
- Updated `BugsnagConfiguration` to allow adding a `BugsnagObserverBlock`, which is invoked whenever a property value changes
- Added `BugsnagStateEvent`, which functions as a data class to contain the event name and data about what changed, structured as per the design document

## Tests

Added unit tests to verify that the observer is called when a value is updated for each stateful property.

Additionally, pre-existing unit tests were added to the macOS.xcodeproj.

Verified in an example app to confirm that `Bugsnag.addMetadata` triggers the observer.

### DIFF
--- a/Source/BugsnagClient.m
+++ b/Source/BugsnagClient.m
@@ -47,6 +47,8 @@
 #import "Bugsnag.h"
 #import "BugsnagErrorTypes.h"
 #import "BugsnagNotifier.h"
+#import "BugsnagMetadataInternal.h"
+#import "BugsnagStateEvent.h"
 
 #if TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
 #import <UIKit/UIKit.h>
@@ -299,6 +301,7 @@ void BSGWriteSessionCrashData(BugsnagSession *session) {
 @property(readwrite, retain, nullable) BugsnagMetadata *config;
 @property(readonly, strong, nullable) BugsnagBreadcrumbs *breadcrumbs;
 - (BOOL)shouldRecordBreadcrumbType:(BSGBreadcrumbType)type;
+- (void)notifyObserver:(BugsnagStateEvent *)event;
 @end
 
 @interface BugsnagEvent ()
@@ -313,12 +316,6 @@ void BSGWriteSessionCrashData(BugsnagSession *session) {
                                   metadata:(BugsnagMetadata *_Nullable)metadata
                               handledState:(BugsnagHandledState *_Nonnull)handledState
                                    session:(BugsnagSession *_Nullable)session;
-@end
-
-@interface BugsnagMetadata ()
-@property(unsafe_unretained) id<BugsnagMetadataDelegate> _Nullable delegate;
-- (NSDictionary *_Nonnull)toDictionary;
-- (id)deepCopy;
 @end
 
 @interface BSGOutOfMemoryWatchdog ()
@@ -378,9 +375,6 @@ NSString *_lastOrientation = nil;
 
         self.extraRuntimeInfo = [NSMutableDictionary new];
         self.metadataLock = [[NSLock alloc] init];
-        self.configuration.metadata.delegate = self;
-        self.configuration.config.delegate = self;
-        self.state.delegate = self;
         self.crashSentry = [BugsnagCrashSentry new];
         self.errorReportApiClient = [[BugsnagErrorReportApiClient alloc] initWithConfig:configuration
                                                                               queueName:@"Error API queue"];
@@ -398,9 +392,20 @@ NSString *_lastOrientation = nil;
 
         // Start with a copy of the configuration metadata
         self.metadata = [[configuration metadata] deepCopy];
+        // sync initial state
         [self metadataChanged:self.configuration.metadata];
         [self metadataChanged:self.configuration.config];
         [self metadataChanged:self.state];
+
+        // add observers for future metadata changes
+        void (^observer)(BugsnagMetadata *) = ^(BugsnagMetadata *metadata) {
+            [self metadataChanged:metadata];
+        };
+        [self.metadata addObserver:observer];
+        [self.configuration.metadata addObserver:observer];
+        [self.configuration.config addObserver:observer];
+        [self.state addObserver:observer];
+
         self.pluginClient = [[BugsnagPluginClient alloc] initWithPlugins:self.configuration.plugins];
 
 #if TARGET_OS_IOS
@@ -993,6 +998,9 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
         } else {
             bsg_log_debug(@"Unknown metadata dictionary changed");
         }
+
+        BugsnagStateEvent *event = [[BugsnagStateEvent alloc] initWithName:kStateEventMetadata data:metadata];
+        [self.configuration notifyObserver:event];
     }
 }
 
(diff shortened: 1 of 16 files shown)


Source/BugsnagConfiguration.m
```diff
@@ -504,4 +521,19 @@ - (void)addPlugin:(id<BugsnagPlugin> _Nonnull)plugin {
...
+    if (self.block != nil) {
+        self.block(event);
+    }
+}
+
+- (void)registerStateObserverWithBlock:(BugsnagObserverBlock _Nonnull)block {
+    self.block = block;
+}
+
+- (void)metadataChanged:(BugsnagMetadata *_Nonnull)metadata {
```
tomlongridge: Should/could `BugsnagMetadata` do this directly rather than delegating to `Configuration` to notify observers?

Source/BugsnagConfiguration.m
```diff
@@ -48,6 +50,8 @@
 NSString * const kBugsnagUserName = @"BugsnagUserName";
 NSString * const kBugsnagUserUserId = @"BugsnagUserUserId";
 
+typedef void (^BugsnagObserverBlock)(BugsnagStateEvent *_Nonnull event);
```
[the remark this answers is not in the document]
fractalwrench: I've removed `BugsnagMetadataDelegate` and used the observer pattern throughout the changeset now.

Source/BugsnagClient.m
```diff
@@ -254,9 +254,8 @@ - (id)initWithConfiguration:(BugsnagConfiguration *)initConfiguration {
         }
 
         self.metadataLock = [[NSLock alloc] init];
-        self.configuration.metadata.delegate = self;
-        self.configuration.config.delegate = self;
```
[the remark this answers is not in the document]
fractalwrench: The equivalent has now been added below in the form of observers.

fractalwrench:
Rebased latest changes.

iOS/Bugsnag.xcodeproj/project.pbxproj
```diff
@@ -203,7 +203,6 @@
 		E7397E2B1F83BC2A0034242A /* BugsnagCollections.m in Sources */ = {isa = PBXBuildFile; fileRef = 8A2C8F421C6BBE3C00846019 /* BugsnagCollections.m */; };
 		E7397E2C1F83BC2A0034242A /* BugsnagConfiguration.m in Sources */ = {isa = PBXBuildFile; fileRef = 8A2C8F441C6BBE3C00846019 /* BugsnagConfiguration.m */; };
 		E7397E2D1F83BC2A0034242A /* BugsnagEvent.m in Sources */ = {isa = PBXBuildFile; fileRef = 8A2C8F461C6BBE3C00846019 /* BugsnagEvent.m */; };
-		E7397E2E1F83BC2A0034242A /* BugsnagMetadata.m in Sources */ = {isa = PBXBuildFile; fileRef = 8A2C8F4A1C6BBE3C00846019 /* BugsnagMetadata.m */; };
```
fractalwrench, line 206: note to self: this seems to have been removed in a merge conflict, and should be reinstated along with the header file in the copy phase

iOS/BugsnagTests/BugsnagStateEventTest.m
```diff
@@ -0,0 +1,63 @@
+//
+//  BugsnagStateEventTest.m
```
fractalwrench, line 2: note to self: this should live in the `Tests` directory

Source/BugsnagClient.m
```diff
@@ -398,9 +394,16 @@ - (id)initWithConfiguration:(BugsnagConfiguration *)initConfiguration {
...
         // Start with a copy of the configuration metadata
         self.metadata = [[configuration metadata] deepCopy];
+        // sync initial state
         [self metadataChanged:self.configuration.metadata];
         [self metadataChanged:self.configuration.config];
         [self metadataChanged:self.state];
+
+        // add observers for future metadata changes
+        [self registerStateObserverWithBlock:^(BugsnagStateEvent *event) {
+            [self metadataChanged:event.data];
```
tomlongridge, line 404: `metadataChanged` has a debug message "Unknown metadata dictionary changed" if it's not one of the three it cares about. This could imply to future readers that it's being triggered incorrectly. Suggest we remove? Also we could check the event type is `kStateEventMetadata ` to further filter the calls?

Source/BugsnagStateEvent.h
```diff
@@ -0,0 +1,27 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+static NSString *const kStateEventContext = @"ContextUpdate";
+static NSString *const kStateEventMetadata = @"MetadataUpdate";
+static NSString *const kStateEventUser = @"UserUpdate";
+
+@interface BugsnagStateEvent : NSObject
+- (instancetype)initWithName:(NSString *)name data:(id)data;
+@property NSString *name;
```
tomlongridge, line 19: Does calling this `type` rather than `name` make its purpose clearer?

Source/BugsnagMetadata.m
```diff
@@ -46,15 +46,47 @@ - (id)initWithDictionary:(NSDictionary *)dict {
...
+}
+
+- (void)notifyObservers {
+    for (BugsnagObserverBlock callback in self.stateEventBlocks) {
+        BugsnagStateEvent *event = [[BugsnagStateEvent alloc] initWithName:kStateEventMetadata data:self];
+        callback(event);
+    }
+}
+
+- (void)registerStateObserverWithBlock:(BugsnagObserverBlock _Nonnull)block {
```
tomlongridge, line 66: _Possibly_ more idiomatic to call this `addObserverUsingBlock`?

https://developer.apple.com/documentation/foundation/nsnotificationcenter/1411723-addobserverforname

Source/BugsnagClient.m
fractalwrench: Removed log call to avoid confusion.

Source/BugsnagMetadata.m
fractalwrench: Renamed to `addObserverUsingBlock`

Source/BugsnagStateEvent.h
fractalwrench: Renamed property

Source/BugsnagClient.m
```diff
@@ -398,9 +394,17 @@ - (id)initWithConfiguration:(BugsnagConfiguration *)initConfiguration {
...
         [self metadataChanged:self.configuration.metadata];
         [self metadataChanged:self.configuration.config];
         [self metadataChanged:self.state];
+
+        // add observers for future metadata changes
+        __weak __typeof__(self) weakSelf = self;
+        [self addObserverUsingBlock:^(BugsnagStateEvent *event) {
+            [weakSelf metadataChanged:event.data];
+        }];
+
```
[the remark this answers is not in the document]
fractalwrench: Added a comment to explain this.

PR #491, opened 2020-03-19
Remove leaveBreadcrumbWithBlock from public api

fractalwrench:
Removes the `leaveBreadcrumbWithBlock` method from our public API. This is still accessible internally as the implementation for adding breadcrumbs currently relies on this method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Bugsnag Notifiers on other platforms.
 
 ## Enhancements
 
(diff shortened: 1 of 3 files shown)


tomlongridge:
Looks like a MR compile issue: https://travis-ci.com/github/bugsnag/bugsnag-cocoa/jobs/299882365#L814

fractalwrench:
Amended the scenarios - the `leaveBreadcrumbWithBlock` method is converted to another method name in Swift which was not updated. I've verified that the breadcrumb scenario passes locally.

PR #492, opened 2020-03-19
Alter api key validation to log warning

fractalwrench:
## Goal

Alters api key validation so that an error is logged and configuration is populated as normal.

This has the side effect of allowing the NSError parameter to be removed from `initWithApiKey` as it is no longer populated in any codepath.

## Changeset

- Updated `initWithApiKey:` signature by removing `NSError` parameter
- Updated `initWithApiKey` to log an error and continue construction rather than returning nil
- Removed constants used for populating the NSError

## Tests

Relied on existing test coverage which was updated to use new method signatures.


tomlongridge:
> Actually, from discussion on the Android/JS changes: we should be throwing an error if the API key is empty so unfortunately I think this means we need to reinstate the error parameter...

Discussed IRL and agreed with @fractalwrench that the proposed changes are probably the more preferable - either way we have a slight spec deviation, but with this implementation it doesn't affect the signature.

fractalwrench:
Fixed a compile issue with the mazerunner scenarios and verified by running the breadcrumb scenario manually.

robinmacharg:
> LGTM as long as we're confident in the compile-time null check for which I'd like @robinmacharg's input.

From [Apples dev blog](https://developer.apple.com/swift/blog/?id=25): "The compiler will tell you if you try to break the rules."

PR #493, opened 2020-03-23
(feat) Increased detail in handled event breadcrumb

robinmacharg:
## Goal

Add additional metadata to handled error breadcrumbs.

## Design

Simply add two additional fields to the preexisting breadcrumb metadata .

## Changeset

`BugsnagClient`.  Some small additional formatting/renaming changes were incorporated.

## Tests

Unit test added.

### Outstanding Questions

  - [X] Final review

  features)


Tests/BugsnagClientTests.m
```diff
@@ -0,0 +1,84 @@
...
+    NSError *error;
+    BugsnagConfiguration *configuration = [[BugsnagConfiguration alloc] initWithApiKey:DUMMY_APIKEY_32CHAR_1 error:&error];
+    if (willNotify) {
+        [configuration addOnSendBlock:^bool(BugsnagEvent * _Nonnull event) { return false; }];
+    }
+    [Bugsnag startBugsnagWithConfiguration:configuration];
+}
+
+/**
+ * Handled events leave a breadcrumb when norify() is called.  Test that values are inserted
```
tomlongridge:
```suggestion
 * Handled events leave a breadcrumb when notify() is called.  Test that values are inserted
```

Tests/BugsnagClientTests.m
```diff
@@ -0,0 +1,84 @@
...
+ * correctly.
+ */
+- (void)testBreadcrumbEnhancedValues {
```
tomlongridge:
```suggestion
- (void)testAutomaticNotifyBreadcrumbData {
```

Source/BugsnagClient.m
```diff
@@ -691,35 +695,36 @@ - (void)notify:(NSException *)exception
...
-      crumb.type = BSGBreadcrumbTypeError;
-      crumb.message = reportName;
-      crumb.metadata = @{
-          BSGKeyMessage : reportMessage,
-          BSGKeySeverity : BSGFormatSeverity(report.severity)
-      };
+        crumb.type = BSGBreadcrumbTypeError;
+        crumb.message = eventErrorClass;
+        crumb.metadata = @{
+            BSGKeyMessage : eventMessage,
```
fractalwrench: How would an empty string for `eventMessage` display in the dashboard for an error report? If it would render as an empty string the implementation should be altered to omit the value from the dictionary if it is nil, otherwise no change is required here.

robinmacharg: Screenshot 2020-03-27 at 18 36 57

Using the example app and hardwiring an empty string gives the attached results, i.e. an empty string is displayed.  `[NSNull null]` resolves to "null" so you're correct, this field needs to be  conditionally added.

PR #495, opened 2020-03-25
fix(swift example app): expand the hit area of buttons in app

xander-jones:
## Goal
Modernise the Swift iOS example app:
- Expand the hit area of buttons
- Port the Out of Memory function from the Objective C example application.

## Changeset
- Removed `LaunchScreen.xib` in favour of `LaunchScreen.storyboard`
- Copied OOM Objective C code from `BigHonkinWebViewController.m/.h`
- Adjusted `VIewController.swift` to use tableView function to call example errors rather than direct `@IBAction` calls from each button. 
- Buttons replaced with labels allow for larger hit area.

## Tests
- Tested on iPhone simulators: iPhone 8 iOS 13.3; iPhone 11 iOS 13.3
- Tested on real iPhone: iPhone 11 Pro iOS 13.3.1

### Outstanding Questions

  - [x] Final review

I'm not in the Bugsnag dev group for Xcode so am unable to sign with this ident. Can you check this builds with that the Bugsnag dev group signing?

  features)


xander-jones:
Closing this out to re-open a new PR for v6 as this was initially for v5.

PR #496, opened 2020-03-25
Remove unused APIs from BugsnagConfiguration interface

fractalwrench:
## Goal

Removes/hides unused APIs from the `BugsnagConfiguration` interface. These are either no longer required or were originally part of the internal implementation and mistakenly included in the public header.

## Changeset

```
BugsnagBeforeNotifyHook
BugsnagMetadata and BugsnagConfig
breadcrumbs
reportBackgroundOOMs
notifyURL and sessionURL
shouldAutoCaptureSessions
autoNotify
shouldSendReports
errorApiHeaders
sessionApiHeaders
```

## Tests

Relied on existing test coverage

### DIFF
--- a/Source/BugsnagConfiguration.m
+++ b/Source/BugsnagConfiguration.m
@@ -178,7 +178,7 @@ NSString * const kBugsnagUserUserId = @"BugsnagUserUserId";
 // -----------------------------------------------------------------------------
 
 /**
- *  Whether reports shoould be sent, based on release stage options
+ *  Whether reports should be sent, based on release stage options
  *
  *  @return YES if reports should be sent based on this configuration
  */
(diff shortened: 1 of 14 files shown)


Source/BugsnagConfiguration.m
```diff
@@ -159,6 +177,11 @@ - (instancetype _Nonnull)initWithApiKey:(NSString *_Nonnull)apiKey
 // MARK: - Instance Methods
 // -----------------------------------------------------------------------------
 
+/**
+ *  Whether reports shoould be sent, based on release stage options
```
tomlongridge:
```suggestion
 *  Whether reports should be sent, based on release stage options
```

PR #498, opened 2020-03-25
Remove unused APIs on BugsnagEvent interface

fractalwrench:
## Goal

Removes/hides unused APIs from the `BugsnagEvent` interface. These are either no longer required or were originally part of the internal implementation and mistakenly included in the public header.

## Changeset

```
BSGParseSeverity
BSGFormatSeverity
serializableValueWithTopLevelData
shouldBeSent
toJson
attachCustomStacktrace:withType:
enhancedErrorMessageForThread
notifyReleaseStages
handledState
overrides
depth
error
isIncomplete
```

## Tests

Updated existing test coverage to avoid using deprecated APIs. Mazerunner scenarios which were testing individual stackframes have had these assertions removed as historically this could be quite flaky.

### DIFF
--- a/features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/HandledInternalNotifyScenario.swift
+++ b/features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/HandledInternalNotifyScenario.swift
@@ -19,7 +19,6 @@ import Bugsnag
                 ["method":"bar()", "file":"parser.js"],
                 ["method":"yes()"]
             ]
-            report.attachCustomStacktrace(frames, withType: "unreal")
         }
     }
 }
--- a/features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/HandledInternalNotifyScenario.swift
+++ b/features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/HandledInternalNotifyScenario.swift
@@ -19,6 +19,7 @@ import Bugsnag
                 ["method":"bar()", "file":"parser.js"],
                 ["method":"yes()"]
             ]
+            report.attachCustomStacktrace(frames, withType: "unreal")
         }
     }
 }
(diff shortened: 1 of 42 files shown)


features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/HandledInternalNotifyScenario.swift
```diff
@@ -19,7 +19,6 @@ import Bugsnag
                 ["method":"bar()", "file":"parser.js"],
                 ["method":"yes()"]
             ]
-            report.attachCustomStacktrace(frames, withType: "unreal")
```
snmaynard: there is a maze runner expectation on this type that is failing

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/UnhandledInternalNotifyScenario.swift
```diff
@@ -19,7 +19,6 @@ import Bugsnag
                 ["method":"baz()", "file":"[native code]"],
                 ["method":"is_done()"]
             ]
-            report.attachCustomStacktrace(frames, withType: "fake")
```
snmaynard: there is a maze runner expectation on this that is failing

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/HandledInternalNotifyScenario.swift
fractalwrench: Yep - left this in the draft state overnight as the mazerunner scenario was using one of the removed APIs. Planning on updating the scenario/assertions today.

tomlongridge:
It looks like we've had to remove some of the test rigor in this change now that we've made some field private. Can we not use some class extensions to give access in the tests?

fractalwrench:
After IRL discussion `attachCustomStacktrace` has been reinstated as it was before, as this allows us to retain the same test coverage. It will be removed as part of future work once the Error interface has been implemented and the existing scenarios can be updated to use this API.

PR #500, opened 2020-03-26
Fix automatic breadcrumb detection

robinmacharg:
## Goal

Work to implement `enabledBreadcrumbTypes` left automatic breadcrumb detection only partially working.  This PR aims to fix it.

## Design

- Add guards around specific pieces of code to hook up automatic breadcrumb event notifications.
- Reinstate code to iterate over a list of events to be notified about, with appropriate guards

## Changeset

Mainly `BugsnagClient` and `BugsnagConfiguration`.  No CHANGELOG since this work merely fixes an issue.

## Tests

- A unit test was negated due to a negation in expected behaviour
- An additional CI test was added to ensure the startup automatic breadcrumb was not left when configured as such.

NOTE: Manual testing of e.g. rotation, battery etc. events is highly recommended.

### Outstanding Questions

  - [X] Final review

  features)

### DIFF
--- a/Source/BugsnagClient.h
+++ b/Source/BugsnagClient.h
@@ -116,6 +116,11 @@
 - (void)crumbleNotification:(NSString *_Nonnull)notificationName;
 
 /**
+ *  Enable or disable automatic breadcrumb collection based on configuration
+ */
+- (void)updateAutomaticBreadcrumbDetectionSettings;
+
+/**
  * Enable or disable crash reporting based on configuration state
  */
 - (void)updateCrashDetectionSettings;
--- a/Source/BugsnagClient.h
+++ b/Source/BugsnagClient.h
@@ -109,18 +109,6 @@
 - (void)clearBreadcrumbs;
 
 /**
- *  Listen for notifications and attach breadcrumbs when received
- *
- *  @param notificationName name of the notification
- */
-- (void)crumbleNotification:(NSString *_Nonnull)notificationName;
-
-/**
- *  Enable or disable automatic breadcrumb collection based on configuration
- */
-- (void)updateAutomaticBreadcrumbDetectionSettings;
-
-/**
  * Enable or disable crash reporting based on configuration state
  */
 - (void)updateCrashDetectionSettings;
--- a/Source/BugsnagClient.m
+++ b/Source/BugsnagClient.m
@@ -875,37 +875,28 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
 }
 
 /**
- * Configure event listeners for automatic breadcrumbs.
+ * Configure event listeners (i.e. observers) for enabled automatic breadcrumbs.
  */
 - (void)updateAutomaticBreadcrumbDetectionSettings {
-
-    // No automatic breadcrumbs are enabled; remove notifications for them all and return.
-    if (![[self configuration] enabledBreadcrumbTypes]) {
-        NSArray *allEventNames = [[[[self automaticBreadcrumbStateEvents]
-            arrayByAddingObjectsFromArray:[self automaticBreadcrumbControlEvents]]
-            arrayByAddingObjectsFromArray:[self automaticBreadcrumbMenuItemEvents]]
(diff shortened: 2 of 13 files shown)


Source/BugsnagClient.m
```diff
@@ -843,6 +873,68 @@ - (void)updateCrashDetectionSettings {
...
+            arrayByAddingObjectsFromArray:[self automaticBreadcrumbMenuItemEvents]]
+            arrayByAddingObjectsFromArray:[self automaticBreadcrumbTableItemEvents]];
+        
+        for (NSString *eventName in allEventNames) {
+            [[NSNotificationCenter defaultCenter] removeObserver:self
+                                                            name:eventName
+                                                          object:nil];
+        }
+        return;
+    }
```
tomlongridge: I don't believe we don't need to remove observers as this is called on `start`, so they won't exist already.

Source/BugsnagClient.m
```diff
@@ -843,6 +873,68 @@ - (void)updateCrashDetectionSettings {
...
+
+    // 'State' events
+    if ([[self configuration] shouldRecordBreadcrumbType:BSGBreadcrumbTypeState]) {
+        for (NSString *name in [self automaticBreadcrumbStateEvents]) {
+            [self crumbleNotification:name];
+        }
+    }
+    
+    // 'User' events
+    if ([[self configuration] shouldRecordBreadcrumbType:BSGBreadcrumbTypeUser]) {
```
tomlongridge: sendBreadcrumbForTableViewNotification sets Navigation breadcrumbs

Source/BugsnagClient.m
```diff
@@ -843,6 +873,68 @@ - (void)updateCrashDetectionSettings {
...
+        for (NSString *name in [self automaticBreadcrumbControlEvents]) {
+            [[NSNotificationCenter defaultCenter]
+                addObserver:self
+                   selector:@selector(sendBreadcrumbForControlNotification:)
+                       name:name
+                     object:nil];
+        }
+        
+        // NSMenu events (Mac only)
+        for (NSString *name in [self automaticBreadcrumbMenuItemEvents]) {
```
tomlongridge: sendBreadcrumbForMenuItemNotification sets State type breadcrumbs

In fact, this seems questionable - might be worth raising a separate ticket to look at whether this should be navigation.

Source/BugsnagClient.h
```diff
@@ -115,6 +115,11 @@
  */
 - (void)crumbleNotification:(NSString *_Nonnull)notificationName;
 
+/**
+ *  Enable or disable automatic breadcrumb collection based on configuration
+ */
+- (void)updateAutomaticBreadcrumbDetectionSettings;
```
tomlongridge: Does this need to be public? It's not in the API, so suggest we move internal.

Source/BugsnagClient.m
robinmacharg: Yup, legacy code moved around without due consideration.

Source/BugsnagClient.m
```diff
@@ -423,10 +424,12 @@ - (void)start {
 
     [self.sessionTracker startNewSessionIfAutoCaptureEnabled];
 
-    [self addBreadcrumbWithBlock:^(BugsnagBreadcrumb *_Nonnull breadcrumb) {
-        breadcrumb.type = BSGBreadcrumbTypeState;
-        breadcrumb.message = BSGBreadcrumbLoadedMessage;
-    }];
+    if ([self.configuration shouldRecordBreadcrumbType:BSGBreadcrumbTypeState]) {
```
fractalwrench: It may be worth creating a method called something like `addAutoBreadcrumb`  which encapsulates this behaviour and reduces the risk of future automatic breadcrumbs forgetting to add this check. [This](https://github.com/bugsnag/bugsnag-android/pull/842/files#diff-d01d476975cb3e9ebc74fb3cec71c943R824) is what the Android implementation has done.

robinmacharg: After discussion It was agreed to make the suggested change as part of #486.

PR #508, opened 2020-03-27
Allow addition/removal of OnBreadcrumb callbacks

fractalwrench:
## Goal

Provides a method for adding/removing callbacks which can mutate every `Breadcrumb` before it is added to the application. This matches the notifier specification.

## Changeset

- Added `addOnBreadcrumbBlock` and `removeOnBreadcrumbBlock` to `Bugsnag` and `BugsnagConfiguration`
- Stored `OnBreadcrumb` callbacks in `BugsnagBreadcrumbs`
- Executed callbacks before adding a breadcrumb and discarded the crumb if false is returned from any callback

## Tests

Added new test coverage for adding/removing callbacks, and verifying that the callback discards/mutates breadcrumbs.

### DIFF
--- a/OSX/Bugsnag.xcodeproj/project.pbxproj
+++ b/OSX/Bugsnag.xcodeproj/project.pbxproj
@@ -162,6 +162,7 @@
 		E79E6BDA1F4E3850002B35F9 /* BSG_RFC3339DateTool.m in Sources */ = {isa = PBXBuildFile; fileRef = E79E6B6E1F4E3850002B35F9 /* BSG_RFC3339DateTool.m */; };
 		E79E6BDB1F4E3850002B35F9 /* BSG_KSCrashReportFilter.h in Headers */ = {isa = PBXBuildFile; fileRef = E79E6B711F4E3850002B35F9 /* BSG_KSCrashReportFilter.h */; };
 		E79E6BDC1F4E3850002B35F9 /* BSG_KSCrashReportFilterCompletion.h in Headers */ = {isa = PBXBuildFile; fileRef = E79E6B721F4E3850002B35F9 /* BSG_KSCrashReportFilterCompletion.h */; };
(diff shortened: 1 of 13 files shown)


OSX/Bugsnag.xcodeproj/project.pbxproj
```diff
@@ -162,6 +162,7 @@
 		E79E6BDA1F4E3850002B35F9 /* BSG_RFC3339DateTool.m in Sources */ = {isa = PBXBuildFile; fileRef = E79E6B6E1F4E3850002B35F9 /* BSG_RFC3339DateTool.m */; };
 		E79E6BDB1F4E3850002B35F9 /* BSG_KSCrashReportFilter.h in Headers */ = {isa = PBXBuildFile; fileRef = E79E6B711F4E3850002B35F9 /* BSG_KSCrashReportFilter.h */; };
 		E79E6BDC1F4E3850002B35F9 /* BSG_KSCrashReportFilterCompletion.h in Headers */ = {isa = PBXBuildFile; fileRef = E79E6B721F4E3850002B35F9 /* BSG_KSCrashReportFilterCompletion.h */; };
+		E7AB4B9E2423E184004F015A /* BugsnagOnBreadcrumbTest.m in Sources */ = {isa = PBXBuildFile; fileRef = E7AB4B9D2423E184004F015A /* BugsnagOnBreadcrumbTest.m */; };
```
tomlongridge: Does this need adding to tvOS too?

fractalwrench: We could get away with not adding it as we typically run tests from either the iOS/OSX workspace, but it can't hurt to add it in here. After the spec work there's a ticket to investigate simplifying the project configuration - currently it's quite easy to miss adding files to one of the 3 workspaces.

tomlongridge: I think we should add them as a matter of course until we consolidate the projects.

PR #509, opened 2020-03-27
Rename notifyReleaseStages to enabledReleaseStages

fractalwrench:
Renames the `notifyReleaseStages` configuration property to `enabledReleaseStages`. This makes bugsnag-cocoa more consistent with the naming of other notifiers and complies with the notifier specification.


Source/BugsnagEvent.m
```diff
@@ -319,7 +319,7 @@ - (instancetype)initWithKSReport:(NSDictionary *)report {
                 }
             }
         } else {
-            _notifyReleaseStages = BSGLoadConfigValue(report, @"notifyReleaseStages");
+            _enabledReleaseStages = BSGLoadConfigValue(report, @"notifyReleaseStages");
```
tomlongridge: Is this deliberately `notifyReleaseStages` to handle serialised configuration values? Should we attempt to read either and change the value of `BSGKeyEnabledReleaseStages`?

fractalwrench: Yep - this is serialized as a field within the KSCrash report. IMO we should not alter the key as this will add extra unnecessary complexity to the implementation - but the variable names have been altered everywhere else.

tomlongridge: I think it would be tidier to rename they key and read both going forward - at least then it's tidier in the code and there's just one place where we can read the old value with an appropriate comment saying why.

fractalwrench: That makes sense. I've updated this file to use the constant defined in `BugsnagKeys.h` instead of a string literal.

PR #513, opened 2020-03-31
Make metadata interface consistent across config client and event

robinmacharg:
## Goal

Implement a consistent metadata interface across Config, Client and Event classes.

## Design

A `BugsnagMetadataStore` instance-level protocol is defined, as well as a class-level version of the same, containing the following methods:

- `addMetadata:toSection:`
- `addMetadata:withKey:toSection:`
- `getMetadataFromSection:`
- `getMetadataFromSection:withKey:`
- `clearMetadataFromSection:`
- `clearMetadataFromSection:withKey:`

`Bugsnag`, `BugsnagClient`, `BugsnagEvent`, `BugsnagConfig` and `BugsnagMetadata` classes are all changed to conform to this.

Internal dictionary metadata implementations are replaced with `BugsnagMetadata` objects.

## Changeset

A wide range of files, not limited to those mentioned above, since many references to object metadata also needed updated. Test apps have also been updated.

## Tests

No significant new tests added, but many changes to existing ones to adopt the new non-dictionary access syntax.

### Outstanding Questions

  - [X] Final review

  features)
- [X] Consistency across platforms for structures or concepts added or modified


Source/BugsnagClient.m
```diff
@@ -649,23 +669,23 @@ - (void)notifyError:(NSError *)error
...
+               };
+               if (report.context == nil) { // set context as error domain
+                    report.context = [NSString stringWithFormat:@"%@ (%ld)", error.domain, (long)error.code];
+               }
+               report.metadata = metadata;
+
+               if (block) {
+                   block(report);
+               }
+           }];
```
robinmacharg: Formatting only

Source/BugsnagClient.m
```diff
@@ -1186,4 +1205,82 @@ - (void)sendBreadcrumbForControlNotification:(NSNotification *)note {
 #endif
 }
 
+// MARK: - <BugsnagMetadataStore>
```
robinmacharg: Dictionary-based conformance.  To be moved to Object-based.

Source/BugsnagEvent.h
```diff
@@ -75,66 +76,6 @@ initWithErrorName:(NSString *_Nonnull)name
...
-
-/**
- * Remove a named metadata value if the key exists in the named section.
- *
- * @param sectionName The name of the section to remove
- * @param key The key to remove
- */
-- (void)clearMetadataInSection:(NSString *_Nonnull)sectionName
-                       withKey:(NSString *_Nonnull)key;
-
```
robinmacharg: All removed in lieu of protocol.

Source/BugsnagEvent.m
```diff
@@ -788,6 +713,82 @@ - (BOOL)unhandled {
     return self.handledState.unhandled;
 }
 
+// MARK: - <BugsnagMetadataStore>
```
robinmacharg: Simple move under a MARK

Source/BugsnagMetadata.h
```diff
@@ -30,14 +30,41 @@
...
+ * @param values A dictionary of string -> id key/value pairs.
+ *               Values should be serializable to JSON.
+ */
+- (void)addMetadata:(NSDictionary *_Nullable)values
+          toSection:(NSString *_Nonnull)section;
+
+- (void)addMetadata:(id _Nullable)metadata
+            withKey:(NSString *_Nonnull)key
+          toSection:(NSString *_Nonnull)section;
+
```
robinmacharg: Moved to match protocol ordering

Tests/BugsnagEventTests.m
```diff
@@ -577,24 +578,25 @@ - (void)testInvalidKeyValueData {
...
+        XCTAssertEqual([[event.metadata toDictionary] count], 1);
+        XCTAssertNil([event.metadata getMetadataFromSection:@"mySection" withKey:@"myNewKey"]);
         
-        [event addMetadataToSectionNamed:@"mySection" key:@"myNewKey" value:@"realValue"];
-        XCTAssertEqual([[event metadata] count], 1);
-        XCTAssertNotNil([[[event metadata] objectForKey:@"mySection"] objectForKey:@"myNewKey"]);
+        [event addMetadata:@"realValue" withKey:@"myNewKey" toSection:@"mySection"];
+        XCTAssertEqual([[event.metadata toDictionary] count], 1);
+        XCTAssertNotNil([event.metadata getMetadataFromSection:@"mySection" withKey:@"myNewKey"]);
+//        XCTAssertNotNil([[[event metadata] objectForKey:@"mySection"] objectForKey:@"myNewKey"]);
```
fractalwrench, line 599: nit: this comment can be removed

Source/BugsnagClient.m
```diff
@@ -866,13 +861,12 @@ - (void)notify:(NSException *)exception
...
-        initWithErrorName:exceptionName
-             errorMessage:message
-            configuration:self.configuration
-                 metadata:[self.configuration.metadata toDictionary]
-             handledState:handledState
-                  session:self.sessionTracker.runningSession];
+    BugsnagEvent *event = [[BugsnagEvent alloc] initWithErrorName:exceptionName
+                                                     errorMessage:message
+                                                    configuration:self.configuration
+                                                         metadata:[[[Bugsnag client] metadata] deepCopy] //[self.configuration.metadata toDictionary]
```
fractalwrench, line 867: This can be simplified to use `self.metadata`

Source/BugsnagMetadata.m
```diff
@@ -43,81 +44,52 @@ - (id)init {
...
 
-- (void)clearMetadataInSection:(NSString *)section
-                           key:(NSString *)key
-{
+- (id)deepCopy {
     @synchronized(self) {
-        if ([[[self dictionary] objectForKey:section] objectForKey:key]) {
-            [[[self dictionary] objectForKey:section] removeObjectForKey:key];
-        }
+        return [[BugsnagMetadata alloc] initWithDictionary:[[self dictionary] mutableDeepCopy]];
```
fractalwrench, line 64: nit: the `mutableDeepCopy` can be avoided here as the constructor implementation already makes a defensive copy

Source/BugsnagMetadata.m
```diff
@@ -43,81 +44,52 @@ - (id)init {
...
     bool metadataChanged = false;
     @synchronized(self) {
-        if (value && value != [NSNull null]) {
-            id cleanedValue = BSGSanitizeObject(value);
+        if (metadata && metadata != [NSNull null]) {
+            id cleanedValue = BSGSanitizeObject(metadata);
             if (cleanedValue) {
                 // Value is OK, try and set it
-                NSMutableDictionary *section = [self getMetadata:tabName];
+                NSMutableDictionary *section = [[self getMetadataFromSection:sectionName] mutableCopy];
```
fractalwrench, line 83: I believe the `mutableCopy` can be avoided here as we can guarantee that the section value is always an `NSMutableDictionary`, since we set it ourselves. The same applies to the 2-param `addMetadata` method.

Source/BugsnagMetadata.m
```diff
@@ -158,12 +130,13 @@ - (void)addMetadataToSection:(NSString *)section
                         id cleanedValue = BSGSanitizeObject(value);
                         if (cleanedValue) {
                             // We only want to create a tab if we have a valid value.
-                            NSMutableDictionary *metadata = [self getMetadata:section];
+                            NSMutableDictionary *metadata = [[self getMetadataFromSection:sectionName] mutableCopy];
                             if (!metadata) {
                                 metadata = [NSMutableDictionary new];
-                                [[self dictionary] setObject:metadata forKey:section];
+                                [self dictionary][sectionName] = metadata;
```
fractalwrench, line 136: +1 for modern Objective-C syntax! 🎉

Source/BugsnagEvent.m
```diff
@@ -315,13 +320,14 @@ - (instancetype)initWithKSReport:(NSDictionary *)report {
...
-            _metadata = [NSMutableDictionary new];
+            self.metadata = [BugsnagMetadata new];
+            
             NSDictionary *sessionData = [report valueForKeyPath:@"user.state.oom.session"];
             if (sessionData) {
                 _session = [[BugsnagSession alloc] initWithDictionary:sessionData];
                 _session.unhandledCount += 1; // include own event
                 if (_session.user) {
-                    _metadata = @{@"user": [_session.user toJson]};
+                    self.metadata = [[BugsnagMetadata alloc] initWithDictionary:[@{@"user": [_session.user toJson]} mutableCopy]];
```
fractalwrench, line 330: nit: unnecessary mutableCopy

Source/BugsnagEvent.m
```diff
@@ -385,22 +398,22 @@ - (instancetype)initWithKSReport:(NSDictionary *)report {
...
+                                  metadata:(BugsnagMetadata *_Nullable)metadata
+                              handledState:(BugsnagHandledState *_Nonnull)handledState
+                                   session:(BugsnagSession *_Nullable)session
+{
     if (self = [super init]) {
         _errorClass = name;
         _errorMessage = message;
         _overrides = [NSDictionary new];
-        _metadata = metadata ?: [NSDictionary new];
+        self.metadata = metadata ?: [BugsnagMetadata new];
```
fractalwrench, line 412: There's a bit of a gotcha here in that the `metadata` passed in by the `BugsnagClient` will continue to be mutated. I'd feel better about making a defensive copy in the constructor than at each call site, as it reduces the risk that someone forgets to add in a call to `deepCopy` sometime in the future.

CHANGELOG.md
```diff
@@ -195,6 +195,10 @@ Bugsnag Notifiers on other platforms.
 * Enhanced device orientation change breadcrumbs.  These are now reported with "from" and "to" values
   in a form consistent with the Android notifier.
   (#486)[https://github.com/bugsnag/bugsnag-cocoa/pull/486]
+  
+* The metadata interface is now consistent across the `BugsnagConfig`, `BugsnagClient` and `BugsnagEvent`
```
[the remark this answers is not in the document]
robinmacharg: And `BugsnagMetadata`.

Source/BugsnagMetadataStore.h
```diff
@@ -0,0 +1,125 @@
...
+ * @param sectionName The name of the metadata section
+ *
+ * @param metadata A dictionary of string -> id key/value pairs.
+ *                 Values should be serializable to JSON.
+ */
+- (void)addMetadata:(NSDictionary *_Nonnull)metadata
+          toSection:(NSString *_Nonnull)sectionName
+    NS_SWIFT_NAME(addMetadata(_:section:));
+
+- (void)addMetadata:(id _Nullable)value
```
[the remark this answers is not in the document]
robinmacharg: Touche :-)

Source/Bugsnag.m
```diff
@@ -246,33 +227,65 @@ + (NSDateFormatter *)payloadDateFormatter {
...
     if ([self bugsnagStarted]) {
-        return [self.client getMetadata:section key:key];
-    } else {
-        return nil;
+        [self.client.metadata clearMetadataFromSection:sectionName
+                                                             withKey:key];
     }
 }
 
+// MARK: -
```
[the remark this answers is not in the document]
robinmacharg: Intentionally - properties/accessors benefit from the dividing lines in the XCode drop-down but don't need a specific title.  There is a task to tidy the entire codebase and perform these annotations and rearrangements more thoroughly.

Source/BugsnagMetadata.m
robinmacharg: In which case the protocol signature also needs updated to include Mutable.

Source/BugsnagEvent.m
robinmacharg: Actually, I think it's required since the signature is:

`- (id)initWithDictionary:(NSMutableDictionary *)dict`

Source/BugsnagEvent.m
robinmacharg: Agreed, although at present there was only one substantive call, you're right putting the defensive copy in a single place is preferred.

Source/BugsnagEvent.m
fractalwrench: Could the constructor be altered to use `NSDictionary`? This would allow both types to be passed in and avoid a need for a copy.

Source/BugsnagMetadata.m
fractalwrench: Sounds like a good approach 👍

Source/Bugsnag.m
```diff
@@ -246,33 +227,65 @@ + (NSDateFormatter *)payloadDateFormatter {
...
     if ([self bugsnagStarted]) {
-        return [self.client getMetadata:section];
-    } else {
-        return nil;
+        [self.client.metadata addMetadata:value
+                                toSection:section];
+    }
+}
+
++ (NSMutableDictionary *)getMetadataFromSection:(NSString *)section
```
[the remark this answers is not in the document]
robinmacharg: Following discussion we've agreed that a mutable interface is more consistent with other notifiers.  No guarantees are made about nested collections.

Source/BugsnagEvent.m
robinmacharg: You're right, if we change the sig. and I think we can without loss of functionality.  We make no promises as long as _some_ form of NS*Dictionary is passed in.

PR #514, opened 2020-03-31
Remove unused APIs from Bugsnag interface

fractalwrench:
## Goal

Removes/hides unused APIs from the `Bugsnag` interface. These are either no longer required or were originally part of the internal implementation and mistakenly included in the public header.

## Changeset

```
- clearBreadcrumbs
- setBreadcrumbCapacity
- payloadFormatter
```

## Tests

Updated existing test coverage to avoid using deprecated APIs.

### DIFF
--- a/Source/Bugsnag.h
+++ b/Source/Bugsnag.h
@@ -155,16 +155,6 @@
                            andType:(BSGBreadcrumbType)type
     NS_SWIFT_NAME(leaveBreadcrumb(_:metadata:type:));
 
-/**
- * Set the maximum number of breadcrumbs to keep and sent to Bugsnag.
- * By default, we'll keep and send the 20 most recent breadcrumb log
- * messages.
- *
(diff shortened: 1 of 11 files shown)


Source/Bugsnag.h
```diff
@@ -29,10 +29,6 @@
 #import "BugsnagMetadata.h"
 #import "BugsnagPlugin.h"
 
-static NSString *_Nonnull const BugsnagSeverityError = @"error";
-static NSString *_Nonnull const BugsnagSeverityWarning = @"warning";
-static NSString *_Nonnull const BugsnagSeverityInfo = @"info";
```
tomlongridge: Should we include in Upgrade Guide in case anyone uses them?

fractalwrench: Added a note to the upgrade guide in case anyone was using these.

PR #515, opened 2020-03-31
Remove unused APIs from BugsnagClient interface

fractalwrench:
## Goal

Removes/hides unused APIs from the `BugsnagClient` interface. These are either no longer required or were originally part of the internal implementation and mistakenly included.

This is in preparation of making `BugsnagClient` publicly accessible, which will be addressed in another PR.

## Changeset

- Unused APIs have been removed/hidden using a [class extension](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html#//apple_ref/doc/uid/TP40011210-CH6-SW6), in the `BugsnagClientInternal` header
- `appCrashedLastLaunch` renamed to `appDidCrashLastLaunch`
- `clearBreadcrumbs` removed from `Bugsnag` interface

## Tests

Updated existing test coverage to avoid using deprecated APIs.

### DIFF
--- a/Source/BugsnagClient.m
+++ b/Source/BugsnagClient.m
@@ -25,6 +25,7 @@
 //
 
 #import "BugsnagClient.h"
+#import "BugsnagClientInternal.h"
 #import "BSGConnectivity.h"
 #import "Bugsnag.h"
 #import "Private.h"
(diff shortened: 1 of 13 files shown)


Source/BugsnagClient.m
```diff
@@ -25,6 +25,7 @@
 //
 
 #import "BugsnagClient.h"
+#import "BugsnagClientInternal.h"
```
tomlongridge: Do you think this is an approach we should take with all our public classes? To provide an internal header rather than all the class extensions we have dotted around?

fractalwrench: Ultimately yes, this would be a good idea. There are quite a few existing class extensions lying around so I think for now it'd be best to follow this pattern only if a class has a lot of properties/methods to hide. I'll create an internal ticket so that we can address this fully at a future date as it's a non-breaking change.

PR #519, opened 2020-04-01
Add classes for structured Event data

fractalwrench:
## Goal

Adds empty source files for classes which will be used to hold structured data in `BugsnagEvent`.

These have been added here as they needed to be added to the iOS/tvOS/macOS xcodeprojs, and the headers need to be publicly visible, imported in the `Bugsnag.h` umbrella header, and added to the podspec.json. Adding them all at once reduces the chance that these steps are forgotten in the individual PRs which will be raised.

## Changeset

- Added the following public classes:

```
App
AppWithState
Device
DeviceWithState
Error
Stacktrace
Thread
```

- Updated `Bugsnag.h` imports so that structured classes are available in the [umbrella header](https://stackoverflow.com/questions/31238761/what-is-an-umbrella-header/31238936)
- Updated podspec so public header files are found by Cocoapods

## Tests

Verified that the iOS project compiles.

### DIFF
--- a/Bugsnag.podspec.json
+++ b/Bugsnag.podspec.json
@@ -30,6 +30,6 @@
   "requires_arc": true,
   "public_header_files": [
     "Source/BSG_KSCrashReportWriter.h",
-    "Source/Bugsnag{,Metadata,Configuration,Breadcrumb,Event,Plugin,Client}.h"
+    "Source/Bugsnag{,Metadata,Configuration,Breadcrumb,Event,App,AppWithState,Device,DeviceWithState,Error,Stackframe,Thread,Plugin,Client}.h"
   ]
 }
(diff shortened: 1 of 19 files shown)


Bugsnag.podspec.json
```diff
@@ -30,6 +30,6 @@
   "requires_arc": true,
   "public_header_files": [
     "Source/BSG_KSCrashReportWriter.h",
-    "Source/Bugsnag{,Metadata,Configuration,Breadcrumb,Event,Plugin,Client}.h"
+    "Source/Bugsnag{,Metadata,Configuration,Breadcrumb,Event,App,AppWithState,Device,DeviceWithState,Error,Stackframe,Thread,Plugin,Client}.h"
```
robinmacharg, line 33: I think this should now be split into separate lines (with the path duplication that implies).  Alphabetised as well to help with future diffing.

fractalwrench:
Reformatted the podspec and added some placeholder comments on the new classes.

PR #520, opened 2020-04-02
Convert event.app from NSDictionary to a structured class

fractalwrench:
## Goal

Converts `event.app` from an `NSDictionary` to a structured class where each field is represented as a property.

`BugsnagApp` is used for sessions, and contains stateless properties that do not change over the lifecycle of an application. `BugsnagAppWithState` extends this class and is used for events, and contains stateful properties that change for each error - e.g. durationInForeground.

## Changeset
- Added `BugsnagApp` and `BugsnagAppWithState` classes with properties for each field in the spec
- Updated `BugsnagConfiguration` to have a default value for `appType`, rather than parsing in the `BugsnagEvent`
- Altered `BugsnagEvent` to use readonly `BugsnagAppWithState` for `event.app`
- Refactored `BSGParseApp` and `BSGParseAppWithState` methods into new `BugsnagApp` classes, so that they can parse a KSCrash report and populate their fields
- Extracted JSON serialization into `BugsnagApp/WithState` from `BugsnagEvent`
- Removed `app.name` as it is the same as `app.id`

## Tests

- Added unit test to verify that `BugsnagApp` and `BugsnagAppWithState` populate correctly, and that they serialize JSON correctly
- Ran example app and verified that app data is sent to the dashboard for an event


Source/BugsnagEvent.m
```diff
@@ -26,6 +26,12 @@
 #import "BugsnagKeys.h"
 #import "BugsnagClient.h"
 
+@interface BugsnagAppWithState ()
++ (BugsnagAppWithState *)appWithDictionary:(NSDictionary *)event config:(BugsnagConfiguration *)config;
+- (NSDictionary *)toDict;
++ (BugsnagAppWithState *)appWithOomData:(NSDictionary *)event;
+@end
+
```
robinmacharg, line 34: It would be good if the class extensions could be collected together and `// MARK:-`ed as a group to help delineate dependencies from the main body of the class.

Source/BugsnagEvent.m
```diff
@@ -411,7 +415,7 @@ - (instancetype _Nonnull)initWithErrorName:(NSString *_Nonnull)name
         _errorMessage = message;
         _overrides = [NSDictionary new];
         self.metadata = [metadata deepCopy] ?: [BugsnagMetadata new];
-        _releaseStage = config.releaseStage;
+        self.releaseStage = config.releaseStage;
```
robinmacharg, line 418: Does it make sense to change other synthesized ivar accesses to dot notation now for consistency?

Source/BugsnagSessionTrackingPayload.h
```diff
@@ -10,9 +10,11 @@
 
 #import "BugsnagSession.h"
 
+@class BugsnagConfiguration;
+
 @interface BugsnagSessionTrackingPayload : NSObject
 
-- (instancetype)initWithSessions:(NSArray<BugsnagSession *> *)sessions;
+- (instancetype)initWithSessions:(NSArray<BugsnagSession *> *)sessions config:(BugsnagConfiguration *)config;
```
robinmacharg, line 17: nit: looks better formatted over two lines

Source/BugsnagSessionTrackingPayload.m
```diff
@@ -24,33 +26,57 @@ @interface Bugsnag ()
 + (BugsnagClient *)client;
 @end
 
+@interface BugsnagApp ()
++ (BugsnagApp *)appWithDictionary:(NSDictionary *)event config:(BugsnagConfiguration *)config;
```
robinmacharg, line 30: nit: looks better formatted over two lines

Source/BugsnagSessionTrackingPayload.m
```diff
@@ -24,33 +26,57 @@ @interface Bugsnag ()
...
+@end
+
+@interface BugsnagSessionTrackingPayload ()
+@property (nonatomic) BugsnagConfiguration *config;
+@end
+
 @implementation BugsnagSessionTrackingPayload
 
-- (instancetype)initWithSessions:(NSArray<BugsnagSession *> *)sessions {
+- (instancetype)initWithSessions:(NSArray<BugsnagSession *> *)sessions config:(BugsnagConfiguration *)config {
```
robinmacharg, line 41: nit: looks better formatted over three lines (brace on separate line).

Source/BugsnagSessionTrackingPayload.m
```diff
@@ -24,33 +26,57 @@ @interface Bugsnag ()
...
-    
+
     for (BugsnagSession *session in self.sessions) {
         [sessionData addObject:[session toJson]];
     }
     BSGDictInsertIfNotNil(dict, sessionData, @"sessions");
     BSGDictSetSafeObject(dict, [Bugsnag client].details, BSGKeyNotifier);
-    
+
+    // app/device data collection currently relies on KSCrash reports,
```
robinmacharg, line 59: nit: we don't need the word "currently".  It does, or it doesn't unless there's future plans or historical info to add.

Source/BugsnagSessionTrackingPayload.m
```diff
@@ -24,33 +26,57 @@ @interface Bugsnag ()
...
-                                                [Bugsnag configuration].releaseStage,
-                                                [Bugsnag configuration].codeBundleId), @"app");
+    NSDictionary *event = [self appInfo:systemInfo config:self.config];
+    BugsnagApp *app = [BugsnagApp appWithDictionary:event config:self.config];
+    BSGDictSetSafeObject(dict, [app toDict], @"app");
     BSGDictSetSafeObject(dict, BSGParseDeviceState(systemInfo), @"device");
     return dict;
 }
 
+- (NSDictionary *)appInfo:(NSDictionary *)systemInfo config:(BugsnagConfiguration *)config {
```
robinmacharg, line 69: nit: 3 lines

Tests/BugsnagAppTest.m
```diff
@@ -0,0 +1,153 @@
...
+
+#import <XCTest/XCTest.h>
+
+#import "BugsnagAppWithState.h"
+#import "BugsnagConfiguration.h"
+#import "BugsnagTestConstants.h"
+
+@interface BugsnagApp ()
++ (BugsnagApp *)appWithDictionary:(NSDictionary *)event config:(BugsnagConfiguration *)config;
+
```
robinmacharg, line 17: nit: remove extra line

Tests/BugsnagAppTest.m
```diff
@@ -0,0 +1,153 @@
...
+                    @"CFBundleShortVersionString": @"5.6.3",
+                    @"CFBundleVersion": @"1",
+                    @"app_uuid": @"dsym-uuid-123"
+            },
+            @"user": @{
+                    @"config": @{
+                            @"releaseStage": @"beta"
+                    }
+            }
+    };
```
robinmacharg, line 54: v. brief commentary explaining that this is a mock KSCrash report would be useful.

Tests/BugsnagSinkTests.m
```diff
@@ -299,12 +299,11 @@ - (void)testEventApp {
...
-    XCTAssertEqual(11, app.count);
+    XCTAssertEqual(9, app.count);
     
     XCTAssertEqualObjects(app[@"id"], @"net.hockeyapp.CrashProbeiOS");
     XCTAssertNotNil(app[@"type"]);
     XCTAssertEqualObjects(app[@"version"], @"1.0");
-    XCTAssertEqualObjects(app[@"name"], @"CrashProbeiOS");
     XCTAssertEqualObjects(app[@"bundleVersion"], @"1");
     XCTAssertEqualObjects(app[@"releaseStage"], @"production");
     XCTAssertEqualObjects(app[@"dsymUUIDs"], @[@"D0A41830-4FD2-3B02-A23B-0741AD4C7F52"]);
```
robinmacharg, line 309: This UUID could live in `BugsnagTestConstants.h`

Source/BugsnagAppWithState.m
```diff
@@ -7,7 +7,52 @@
 //
 
 #import "BugsnagAppWithState.h"
+#import "BugsnagKeys.h"
+#import "BugsnagConfiguration.h"
+#import "BugsnagCollections.h"
+
+@interface BugsnagApp ()
++ (void)populateFields:(BugsnagApp *)app dictionary:(NSDictionary *)event config:(BugsnagConfiguration *)config;
```
robinmacharg, line 15: nit: split over multiple lines

Source/BugsnagApp.m
```diff
@@ -7,7 +7,42 @@
...
 
 @implementation BugsnagApp
 
++ (BugsnagApp *)appWithDictionary:(NSDictionary *)event config:(BugsnagConfiguration *)config {
+    BugsnagApp *app = [BugsnagApp new];
+    [self populateFields:app dictionary:event config:config];
+    return app;
+}
+
++ (void)populateFields:(BugsnagApp *)app dictionary:(NSDictionary *)event config:(BugsnagConfiguration *)config {
```
robinmacharg, line 22: nit: split over multiple lines

Source/BugsnagSessionTrackingPayload.h
fractalwrench: Formatted all instances in this file like this, marked the rest as resolved.

Source/BugsnagSessionTrackingPayload.m
fractalwrench: Moved the brace for now. We should probably decide on a common style at some point down the line, as parts of the codebase are inconsistent on formatting.

Tests/BugsnagAppTest.m
fractalwrench: Added a brief comment here

Source/BugsnagEvent.m
fractalwrench: Added a few comments and rearranged accordingly.

Tests/BugsnagSinkTests.m
fractalwrench: IMO it's best to keep this ID as-is for now, since it's only used in one test source file and it's unrelated to this changeset. Can update if you feel strongly about this

Source/BugsnagEvent.m
fractalwrench: I think it'd be best to avoid doing this in the constructor - I've only changed this for `releaseStage` as the side-effect of the setter is required. In this case the setter sets an 'override property' which is then persisted in the KSCrashReport and can be read when the report is delivered.

Previous to this changeset we set this via `event.releaseStage`, but now because we're enforcing `event.app.releaseStage`, we need to serialize the value in the same place in the KSCrash report. Using this felt like the simplest mechanism.

fractalwrench:
>Is it worth putting new structured containers in a group in XCode to make working with them easier?

This is definitely worth doing - I'll raise a separate PR to address this.

Source/BugsnagEvent.m
robinmacharg: Good explanation of the thinking behind the change, thanks.

PR #526, opened 2020-04-03
Convert event.device from NSDictionary to a structured class

fractalwrench:
## Goal
Converts `event.device` from an `NSDictionary` to a structured class where each field is represented as a property.

`BugsnagDevice` is used for sessions, and contains stateless properties that do not change over the lifecycle of an application. `BugsnagDeviceWithState` extends this class and is used for events, and contains stateful properties that change for each error - e.g. `freeDisk`.

## Changeset
- Added `BugsnagDevice` and `BugsnagDeviceWithState` classes with properties for each field in the spec
- Altered `BugsnagEvent` to use readonly `BugsnagDeviceWithState` for `event.device`
- Refactored `BSGParseDevice` and `BSGParseDeviceWithState` methods into new `BugsnagDevice` classes, so that they can parse a KSCrash report and populate their fields
- Extracted JSON serialization into `BugsnagDevice/WithState` from `BugsnagEvent`
- Moved surplus device metadata fields into `event.metadata`

## Tests
- Added unit test to verify that `BugsnagDevice` and `BugsnagDeviceWithState` populate correctly, and that they serialize JSON correctly
- Ran example app and verified that device data is sent to the dashboard for an event


Source/BugsnagDevice.m
```diff
@@ -7,7 +7,48 @@
...
+    device.osVersion = system[@"system_version"];
+    device.totalMemory = system[@"memory"][@"usable"];
+
+    NSMutableDictionary *runtimeVersions = [NSMutableDictionary new];
+    runtimeVersions[@"osBuild"] = system[@"os_version"];
+    runtimeVersions[@"clangVersion"] = system[@"clang_version"];
+    device.runtimeVersions = runtimeVersions;
+}
+
+- (NSDictionary *)toDict {
```
robinmacharg, line 40: Prior classes (`BugsnagConfiguration`, `BugsnagMetadata` etc.) use `toDictionary` as the method name.  I think we should be consistent.  And longer is more Cocoa.

features/steps/session_tracking_steps.rb
```diff
@@ -7,7 +7,6 @@
     And the payload field "device.manufacturer" equals "Apple"
     And the payload field "device.jailbroken" is not null
     And the payload field "device.modelNumber" is not null
-    And the payload field "device.wordSize" is not null
```
robinmacharg, line 10: Are we sure about removing `wordSize`?  I may not be quite on the money but it seems to be used server-side in determining the primary dysm for symbolication of crash reports?

Source/BugsnagDeviceWithState.m
```diff
@@ -7,7 +7,85 @@
...
+
+    NSString *val = [event valueForKeyPath:@"report.timestamp"];
+
+    if (val != nil) {
+        device.time = [device.formatter dateFromString:val];
+    }
+    return device;
+}
+
+- (NSDictionary *)toDict {
```
robinmacharg, line 79: nit: `toDict` -> `toDictionary`

Source/BugsnagDeviceWithState.m
```diff
@@ -7,7 +7,85 @@
...
+    NSDictionary *fileSystemAttrs =
+            [fileManager attributesOfFileSystemForPath:path error:&error];
+
+    if (error) {
+        bsg_log_warn(@"Failed to read free disk space: %@", error);
+    } else {
+        freeBytes = [fileSystemAttrs objectForKey:NSFileSystemFreeSize];
+    }
+    return freeBytes;
+}
```
robinmacharg, line 29: Moving this function is a good opportunity to provide it with some documentation.

Source/BugsnagDeviceWithState.m
```diff
@@ -7,7 +7,85 @@
...
+@property (nonatomic, readonly) NSDateFormatter *formatter;
+@end
 
 @implementation BugsnagDeviceWithState
 
+- (instancetype)init {
+    if (self = [super init]) {
+        _formatter = [NSDateFormatter new];
+        _formatter.dateFormat = @"yyyy'-'MM'-'dd'T'HH':'mm':'ssZZZ";
+        _formatter.timeZone = [NSTimeZone timeZoneForSecondsFromGMT:0];
```
robinmacharg, line 48: `Bugsnag` has a `+ (NSDateFormatter *)payloadDateFormatter` method.  Is there any reason not to use that?  The tests also repeat formatter initialisation and could be simplified.

Source/BugsnagKSCrashSysInfoParser.m
```diff
@@ -10,77 +10,20 @@
 #import "Bugsnag.h"
 #import "BugsnagCollections.h"
 #import "BugsnagKeys.h"
-#import "BugsnagConfiguration.h"
 #import "Private.h"
```
robinmacharg, line 13: I think `Bugsnag.h`, `BugsnagKeys.h` and `proivate.h` can be removed here.

Source/BugsnagKSCrashSysInfoParser.h
```diff
@@ -10,5 +10,4 @@
 
 #define PLATFORM_WORD_SIZE sizeof(void*)*8
```
robinmacharg, line 11: To my mind this should now live in the `BugsnagDevice.h` header.

Source/BugsnagSessionTrackingPayload.m
```diff
@@ -26,6 +26,11 @@ @interface Bugsnag ()
 + (BugsnagClient *)client;
 @end
 
+@interface BugsnagDevice ()
++ (BugsnagDevice *)deviceWithDictionary:(NSDictionary *)event;
+- (NSDictionary *)toDict;
```
robinmacharg, line 31: nit: `toDict` -> `toDictionary`

iOS/BugsnagTests/BugsnagKSCrashSysInfoParserTest.m
```diff
@@ -0,0 +1,31 @@
...
+    XCTAssertGreaterThan([freeBytes integerValue], 1000, @"expect at least 1k of free space on test device");
+}
+
+- (void)testDeviceFreeSpaceShouldBeNilWhenFailsToRetrieveIt {
+    NSSearchPathDirectory notAccessibleDirectory = NSAdminApplicationDirectory;
+    NSNumber *freeBytes = BSGDeviceFreeSpace(notAccessibleDirectory);
+    XCTAssertNil(freeBytes, @"expect nil when fails to retrieve free space for the directory");
+}
+
+@end
```
robinmacharg, line 31: Are these tests actually targeting `BugsnagDeviceWithState`?  If `BugsnagKSCrashSysInfoParser` is removed, once the tests are relocated then so can this file.

Tests/BugsnagEventTests.m
```diff
@@ -393,6 +393,7 @@ - (void)testHandledReportMetaData {
     NSDictionary *dict = @{@"user.handledState": [state toJson], @"user.metaData": [metadata toDictionary]};
 
     BugsnagEvent *event = [[BugsnagEvent alloc] initWithKSReport:dict];
+    [event clearMetadataFromSection:@"device"];
```
robinmacharg, line 396: Do we also want at least one test to check that the device property contains something?

Source/BugsnagDevice.m
fractalwrench: Renamed to `toDictionary`

Tests/BugsnagEventTests.m
fractalwrench: The existing test is asserting that the metadata dictionary has a count of 1 item after calling `addMetadata`. This assumption was invalidated because `event.metadata.device` now captures several fields that used to be on `event.device`.

As this is just a precondition for a test I'm not sure that this requires any action

Source/BugsnagKSCrashSysInfoParser.h
fractalwrench: This would make the macro publicly accessible which wouldn't be desirable as it might need changing/removing in future. Additionally this macro is also used in `BSGOutOfMemoryWatchdog`, which doesn't currently have knowledge about `BugsnagDevice`

fractalwrench: Moved to `BSGOutOfMemoryWatchdog.h` - as this is non-public and the only other place that uses it

Source/BugsnagKSCrashSysInfoParser.m
fractalwrench: Moved the function definition to `BugsnagDeviceWithState.m` to avoid ballooning `BugsnagEvent`, and added a forward declaration where it is used in the event class.

iOS/BugsnagTests/BugsnagKSCrashSysInfoParserTest.m
fractalwrench: Removed this test after relocating to `BugsnagDeviceTest`

Source/BugsnagDeviceWithState.m
fractalwrench: Added a few docs

Source/BugsnagDeviceWithState.m
fractalwrench: Updated the implementation to use `[Bugsnag payloadDateFormatter]`. I've left the tests using a different formatter as that should hopefully catch the case if something ever goes wrong with the `payloadDateFormatter`.

robinmacharg: Late-to-the-game addenda on the evils of repeated date formatter instantiation can be found e.g. [here](https://gist.github.com/JoshuaSullivan/6c0d0134e66c60c6cd8c).

PR #527, opened 2020-04-06
Update project source code organisation

fractalwrench:
Updates the source code organisation to use groups in the iOS project.

This makes the files a bit easier to browse as they previously wouldn't fit on the screen. Logical ordering should also make it easier to find the necessary file.

A brief amount of documentation on how the files have been grouped has also been added.


fractalwrench:
Updated changeset to link to the markdown from `CONTRIBUTING.md`, and updated osx/tvos projects also.

PR #528, opened 2020-04-06
Create structured BugsnagStackframe class

fractalwrench:
## Goal
Creates a publicly accessible `BugsnagStackframe` class which holds fields in properties, rather than an `NSDictionary`.

A project-visible `BugsnagStacktrace` has been created to sanitise the stacktrace - notably by trimming its length to <200 frames, as per the spec.

The `BugsnagStackframe` class will eventually be present on the `BugsnagError` and `BugsnagThread` classes, which will be implemented in future PRs. For now the JSON serializer has been updated to use the object internally.

## Changeset
- Add properties to `BugsnagStackframe` which can be edited by the user. Note that these deviate from the spec
- Create `BugsnagStacktrace` which limits the number of `BugsnagStackframe` to <200, and encapsulates stacktraces to allow any similar modifications in future
- Updated `BugsnagEvent` JSON serialization to use `BugsnagStacktrace` internally
- Refactored `BSGFormatFrame` to serializers in `BugsnagStackframe/BugsnagStacktrace`

## Tests
Added unit tests to verify that stacktraces can be parsed from a dictionary/serialized to JSON. Relied on existing unit test coverage to verify the event JSON remains unchanged.


Source/BugsnagStackframe.m
```diff
@@ -7,7 +7,60 @@
...
+}
+
++ (BugsnagStackframe *)frameFromDict:(NSDictionary *)dict
+                          withImages:(NSArray *)binaryImages {
+    BugsnagStackframe *frame = [BugsnagStackframe new];
+    frame.frameAddress = [dict[@"instruction_addr"] unsignedLongValue];
+    frame.symbolAddress = [dict[@"symbol_addr"] unsignedLongValue];
+    frame.machoLoadAddress = [dict[@"object_addr"] unsignedLongValue];
+    frame.machoFile = dict[@"object_name"];
+    frame.method = dict[@"symbol_name"];
```
robinmacharg, line 31: With these, is there any reason not to use the `KSCrashReportFields` constants?

Source/BugsnagStackframe.m
```diff
@@ -7,7 +7,60 @@
...
+    if (image != nil) {
+        frame.machoUuid = image[@"uuid"];
+        frame.machoVmAddress = [image[@"image_vmaddr"] unsignedLongValue];
+        return frame;
+    } else { // invalid frame, skip
+        return nil;
+    }
+}
+
+- (NSDictionary *)toDict {
```
robinmacharg, line 46: nit: `toDict` -> `toDictionary`

Source/BugsnagStackframe.m
```diff
@@ -7,7 +7,60 @@
...
+    NSMutableDictionary *dict = [NSMutableDictionary new];
+    BSGDictInsertIfNotNil(dict, self.machoFile, BSGKeyMachoFile);
+    BSGDictInsertIfNotNil(dict, self.method, @"method");
+    BSGDictInsertIfNotNil(dict, self.machoUuid, BSGKeyMachoUUID);
+
+    NSString *frameAddr = [NSString stringWithFormat:BSGKeyFrameAddrFormat, self.frameAddress];
+    BSGDictSetSafeObject(dict, frameAddr, @"frameAddress");
```
robinmacharg, line 53: Can this be added to `BugsnagKeys`?

Source/BugsnagStacktrace.h
```diff
@@ -0,0 +1,25 @@
...
+ * Representation of a stacktrace in a bugsnag error report
+ */
+@interface BugsnagStacktrace : NSObject
+
+- (instancetype)initWithTrace:(NSArray<NSDictionary *> *)trace
+                 binaryImages:(NSArray<NSDictionary *> *)binaryImages;
+
+- (NSArray *)toArray;
+
+@property NSMutableArray<BugsnagStackframe *> *trace;
```
robinmacharg, line 23: I think this can be private. It's only used in the class and tests.

Source/BugsnagStacktrace.m
```diff
@@ -0,0 +1,46 @@
...
+
+#import "BugsnagStacktrace.h"
+#import "BugsnagStackframe.h"
+#import "BugsnagCollections.h"
+#import "BugsnagKeys.h"
+
+@interface BugsnagStackframe ()
++ (BugsnagStackframe *)frameFromDict:(NSDictionary *)dict
+                          withImages:(NSArray *)binaryImages;
+- (NSDictionary *)toDict;
```
robinmacharg, line 17: nit: `toDictionary`

Source/BugsnagStacktrace.m
```diff
@@ -0,0 +1,46 @@
...
+//  Bugsnag
+//
+//  Created by Jamie Lynch on 06/04/2020.
+//  Copyright © 2020 Bugsnag. All rights reserved.
+//
+
+#import "BugsnagStacktrace.h"
+#import "BugsnagStackframe.h"
+#import "BugsnagCollections.h"
+#import "BugsnagKeys.h"
```
robinmacharg, line 12: I don't think these two imports are required.

Source/BugsnagStacktrace.m
```diff
@@ -0,0 +1,46 @@
...
+                 binaryImages:(NSArray<NSDictionary *> *)binaryImages {
+    if (self = [super init]) {
+        self.trace = [NSMutableArray new];
+
+        for (NSDictionary *obj in trace) {
+            BugsnagStackframe *frame = [BugsnagStackframe frameFromDict:obj withImages:binaryImages];
+
+            if (frame != nil && [self.trace count] < 200) {
+                [self.trace addObject:frame];
+            }
```
robinmacharg, line 32: Is it worth adding logging here if we're discarding frames?

Source/BugsnagStackframe.m
fractalwrench: Renamed this and other similar methods.

Source/BugsnagStacktrace.m
fractalwrench: Removed the unused imports

Source/BugsnagStacktrace.h
fractalwrench: Moved to a class extension inside `BugsnagStacktrace.m`

Source/BugsnagStackframe.m
fractalwrench: No reason - I've amended the implementation to use constants instead.

Source/BugsnagStacktrace.m
fractalwrench: I believe the current behaviour is to discard at least one frame from most KSCrash reports, which is done silently. Presumably the reason behind this is that otherwise we'd log a message out for each event that is captured.

fractalwrench:
Addressed inline comments and resolved the XCode source code organisation nit. I've also rebased against v6 to fix the merge conflicts.

Source/BugsnagStacktrace.m
robinmacharg: My thinking was that exceeding 200 frames is an unusual event, and may in itself point to a situation that needs investigated during development, so it was logging (once) when the upper limit is exceeded, not every single discard.

fractalwrench: Agreed after IRL discussion to raise this as a point with the spec, and address separately if any action required

robinmacharg: It's been agreed that this will be looked at at a later date.

Issue #521, opened 2020-04-02
-watchLifecycleEvents: is listening to the same notification for foreground/background.


bolsinga:
### Description
I was reading the code to understand more and found the issue I describe here.

### Issue
```
#if TARGET_OS_TV || TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
    foregroundName = UIApplicationWillEnterForegroundNotification;
    backgroundName = UIApplicationWillEnterForegroundNotification;
    #elif TARGET_OS_MAC
    foregroundName = NSApplicationWillBecomeActiveNotification;
    backgroundName = NSApplicationDidFinishLaunchingNotification;
    #endif

```
### Environment

Library versions:

master

it should be
    `backgroundName = UIApplicationDidEnterBackgroundNotification;`

xander-jones:
Hey @bolsinga, thanks for this report. We will take a look into this.

xander-jones:
Hey @bolsinga, the fix for this has been released in [v5.23.1](https://github.com/bugsnag/bugsnag-cocoa/releases/tag/v5.23.1)

PR #529, opened 2020-04-06
(fix) Automatic session reporting now correctly records a new session…

robinmacharg:
… if returning to the foreground after more than 60 seconds

## Goal

#521 highlighted a copy-and-paste error in the event listening code that meant that apps returning to the foreground after more than 60 seconds were not correctly reporting that as the start of a new session.  This PR fixes that.

## Design

A single wrong notification name in `BugsnagNotifier` needed corrected.

## Tests

An end-to-end test has been added that ensures that after an appropriate sequence of foreground/background changes an app reports two sessions with differing ids.

### Outstanding Questions

  - [X] Final review

  features)


robinmacharg:
FP'd minor corrective updates to the feature.

PR #532, opened 2020-04-08
Create structured BugsnagThread class

fractalwrench:
## Goal
Creates a publicly accessible `BugsnagThread` class which holds fields in properties, rather than an `NSDictionary`.

This is used as a readonly array property on `BugsnagEvent`, where the user can alter thread information in callbacks. The changeset relies on #528 which has not yet been merged.

## Changeset
- Add properties to `BugsnagThread` which can be edited by the user. Note that these deviate from the spec, specifically `thread.name` whose value will always be nil by default
- Updated `BugsnagEvent` thread generation to occur in constructor rather than in the `toJson` method

- Updated `BugsnagEvent` JSON serialization to use `BugsnagThread` internally

## Tests
Added unit tests to verify that threads can be parsed from a dictionary/serialized to JSON. Added additional coverage to ensure that thread information is enhanced with `isPC` and `isLR`. Relied on existing unit test coverage to verify the event JSON remains unchanged.


Source/BugsnagEvent.m
```diff
@@ -79,6 +80,26 @@ + (BugsnagStackframe *)frameFromDict:(NSDictionary *)dict
                           withImages:(NSArray *)binaryImages;
 @end
 
+@interface BugsnagThread ()
+@property BugsnagStacktrace *trace;
+- (NSDictionary *)toDict;
```
robinmacharg, line 85: nit: `toDictionary`

Source/BugsnagThread.m
```diff
@@ -7,7 +7,103 @@
...
+    BSGDictInsertIfNotNil(dict, self.id, @"id");
+    BSGDictInsertIfNotNil(dict, self.name, @"name");
+    BSGDictSetSafeObject(dict, @(self.errorReportingThread), @"errorReportingThread");
+    BSGDictSetSafeObject(dict, self.type == BSGThreadTypeCocoa ? @"cocoa" : @"reactnativejs", @"type");
+    BSGDictSetSafeObject(dict, @(self.errorReportingThread), @"errorReportingThread");
+    BSGDictSetSafeObject(dict, [self.trace toArray], @"stacktrace");
+    return dict;
+}
+
++ (NSMutableArray *)serializeThreads:(NSMutableDictionary *)event threads:(NSArray<BugsnagThread *> *)threads {
```
robinmacharg, line 55: Is `event` used?  This method is only called from `BugsnagEvent` so the parameter could be left off.

Source/BugsnagThread.m
```diff
@@ -7,7 +7,103 @@
...
+
+@interface BugsnagThread ()
+@property BugsnagStacktrace *trace;
+@end
+
+@interface BugsnagStacktrace ()
+@property NSMutableArray<BugsnagStackframe *> *trace;
+@end
 
 @implementation BugsnagThread
```
robinmacharg, line 26: I think this class would benefit from some method documentation.  The methods are well named but a single line summary of e.g. what enhancing a thread sets out to do would aid understanding.

Source/BugsnagThread.m
```diff
@@ -7,7 +7,103 @@
...
+    for (NSDictionary *thread in threads) {
+        NSDictionary *threadInfo = [self enhanceThreadInfo:thread depth:depth errorType:errorType];
+        BugsnagThread *obj = [[BugsnagThread alloc] initWithThread:threadInfo binaryImages:binaryImages];
+        [bugsnagThreads addObject:obj];
+    }
+    return bugsnagThreads;
+}
+
++ (NSDictionary *)enhanceThreadInfo:(NSDictionary *)thread
+                              depth:(NSUInteger)depth
```
robinmacharg, line 78: I assume that this is the frame discard depth, and I know this is declared historically as `depth` in  `BugsnagEvent`.  So this comment may be out of scope for the change, but it would be good to see more explicit naming of this.  `depth` feels a little too generic.

Source/BugsnagThread.m
```diff
@@ -7,7 +7,103 @@
...
+                          errorType:(NSString *)errorType {
+    NSArray *backtrace = thread[@"backtrace"][@"contents"];
+    BOOL isReportingThread = [thread[@"crashed"] boolValue];
+
+    if (isReportingThread) {
```
robinmacharg, line 83: A comment on the implications of being the reporting thread would help understanding.

Source/BugsnagEvent.m
fractalwrench: Renamed method

Source/BugsnagThread.m
fractalwrench: It doesn't seem to be used - I'll remove the parameter.

Source/BugsnagThread.m
fractalwrench: Added a few more comments in general - particularly around the thread enhancement.

Source/BugsnagThread.m
fractalwrench: I think there is value in leaving this as-is unless you feel strongly, as the concept of 'depth' is established and documented: https://docs.bugsnag.com/platforms/ios-objc/reporting-handled-exceptions/#depth

robinmacharg: Agreed; this should be addressed separately if changes need to be made consistently across multiple notifiers and doc sets.

PR #534, opened 2020-04-14
Make all callbacks return boolean values

fractalwrench:
## Goal

Updates the `OnSendBlock` and `OnSessionBlock` callbacks to return boolean values which control whether the event/session is sent or not.

## Changeset
- Updated return type of blocks to `BOOL` for both callback types
- Moved `OnSessionBlock` callbacks into the `SessionTracker`. If a callback returns false, the new session is discarded and the original session (if any) is used

Note: the session callback is effectively passed an empty dictionary as its parameter for now, which has been noted as a TODO comment. Future work has been scheduled which will alter the parameter to take a structured class with session information instead.

## Tests
Added new unit tests to verify the behaviour for session blocks.

### DIFF
--- a/Source/BugsnagSessionTracker.m
+++ b/Source/BugsnagSessionTracker.m
@@ -117,21 +117,20 @@ NSString *const BSGSessionUpdateNotification = @"BugsnagSessionChanged";
         return;
     }
 
-    BugsnagSession *newSession = [[BugsnagSession alloc] initWithId:[[NSUUID UUID] UUIDString]
-                                                          startDate:[NSDate date]
-                                                               user:self.config.user
-                                                       autoCaptured:isAutoCaptured];
     // TODO JL: refactor to pass as structured data into callback
-    BugsnagSessionTrackingPayload *payload = [[BugsnagSessionTrackingPayload alloc] initWithSessions:@[]
-                                                                                              config:self.config];
-    NSMutableDictionary *data = [payload toJson];
-    for (BugsnagOnSessionBlock cb in self.config.onSessionBlocks) {
-        if (!cb(data)) {
+    BugsnagSessionTrackingPayload *sessions = [[BugsnagSessionTrackingPayload alloc] initWithSessions:@[]
+                                                                                               config:self.config];
+    NSMutableDictionary *sessionsJson = [sessions toJson];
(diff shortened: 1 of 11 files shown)


Source/BugsnagSessionTracker.m
```diff
@@ -116,11 +117,21 @@ - (void)startNewSessionWithAutoCaptureValue:(BOOL)isAutoCaptured {
...
+                                                       autoCaptured:isAutoCaptured];
+    // TODO JL: refactor to pass as structured data into callback
+    BugsnagSessionTrackingPayload *payload = [[BugsnagSessionTrackingPayload alloc] initWithSessions:@[]
+                                                                                              config:self.config];
+    NSMutableDictionary *data = [payload toJson];
+    for (BugsnagOnSessionBlock cb in self.config.onSessionBlocks) {
+        if (!cb(data)) {
+            return;
+        }
+    }
```
robinmacharg, line 132: nit: this refactoring is a good opportunity to improve naming for clarity.  What do you think about:

* `payload` -> `sessions`
* `data` -> `sessionsJson`
* `cb` -> `onSessionBlock`

Source/BugsnagSessionTracker.m
```diff
@@ -116,11 +117,21 @@ - (void)startNewSessionWithAutoCaptureValue:(BOOL)isAutoCaptured {
...
     }
 
-    self.currentSession = [[BugsnagSession alloc] initWithId:[[NSUUID UUID] UUIDString]
-                                                   startDate:[NSDate date]
-                                                        user:self.config.user
-                                                autoCaptured:isAutoCaptured];
+    BugsnagSession *newSession = [[BugsnagSession alloc] initWithId:[[NSUUID UUID] UUIDString]
+                                                          startDate:[NSDate date]
+                                                               user:self.config.user
+                                                       autoCaptured:isAutoCaptured];
```
robinmacharg, line 123: This has become separated from its first use.

Source/BugsnagSessionTracker.m
fractalwrench: sounds good :) renamed

Source/BugsnagSessionTracker.m
fractalwrench: Move closer to first assignment.

Issue #523, opened 2020-04-02
Unable to build Bugsnag 5.23.0 for tvOS using Carthage


faithfracture:
### Description
I'm trying to update Bugsnag-cocoa from 5.22.6 to 5.23.0 and it fails to build. 5.22.6, .7, .8 all fail for all platforms completely. 5.22.9 and 5.23.0 build for iOS, but that's it. Judging from the log output it looks like the build scheme isn't configured correctly - all of the tvOS simulators are listed as "Ineligible destinations"

### Issue
It fails to build

### Environment
Xcode 11.4

Library versions:

* Bugsnag version (from your Podfile, Podfile.lock, or elsewhere): 5.23.0
* Carthage version (if any): 0.34.0

### Example code snippet
Cartfile:

```
github "AFNetworking/AFNetworking" ~> 4.0
github "jdg/MBProgressHUD" ~> 1.1
github "erikdoe/ocmock" ~> 3.4
github "bugsnag/bugsnag-cocoa"
```

<details><summary>Error messages:</summary>

Command Line
```
 $ ✗  carthage build --platform tvos
*** xcodebuild output can be found in /var/folders/vx/yjcdkqmd4x9_6yc851v4ch9m0000gn/T/carthage-xcodebuild.p187R4.log
*** Downloading ocmock.framework binary at "OCMock 3.6"
*** Building scheme "AFNetworking tvOS" in AFNetworking.xcworkspace
*** Building scheme "Bugsnag" in Bugsnag.xcodeproj
*** Building scheme "Bugsnag" in Bugsnag.xcodeproj
Build Failed
	Task failed with exit code 70:
	/usr/bin/xcrun xcodebuild -project /project/Carthage/Checkouts/bugsnag-cocoa/iOS/Bugsnag.xcodeproj -scheme Bugsnag -configuration Release -derivedDataPath /Users/myuser/Library/Caches/org.carthage.CarthageKit/DerivedData/11.4_11E146/bugsnag-cocoa/v5.23.0 -sdk appletvsimulator -destination platform=tvOS\ Simulator,id=5E249F73-CE56-4285-956B-6DFBE09DD3E2 -destination-timeout 3 ONLY_ACTIVE_ARCH=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES build (launched in /project/Carthage/Checkouts/bugsnag-cocoa)

[...]

PR #535, opened 2020-04-14
(fix) Fixes Carthage support for tvOS

robinmacharg:
## Goal

[#523](https://github.com/bugsnag/bugsnag-cocoa/issues/523) notes an issue when using Carthage to build a project utilising Bugsnag that was causing tvOS builds to attempt to use the iOS project and thereby fail to build.  This PR addresses this issue.

## Design

The XCode build setting `SUPPORTED_PLATFORMS` was blank in the iOS project implying using the base SDK and masking the correct project for building tvOS.  Carthage, in turn, then used the iOS project to build the tvOS framework.  By setting this property to `iOS` Carthage then correctly finds the `tvOS` project.

## Tests

The Travic CI Makefile has been updated to also now run the Carthage tvOS build where previously it was only building MacOS/OSX and iOS.

### Outstanding Questions

  - [ ] Final review
  - [X] Release

  features)

### DIFF
--- a/iOS/Bugsnag.xcodeproj/project.pbxproj
+++ b/iOS/Bugsnag.xcodeproj/project.pbxproj
@@ -1355,6 +1355,7 @@
 				MTL_ENABLE_DEBUG_INFO = YES;
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = iphoneos;
+				SUPPORTED_PLATFORMS = "iphonesimulator iphoneos";
 				TARGETED_DEVICE_FAMILY = "1,2";
 				VERSIONING_SYSTEM = "apple-generic";
 				VERSION_INFO_PREFIX = "";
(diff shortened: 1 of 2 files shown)


iOS/Bugsnag.xcodeproj/project.pbxproj
```diff
@@ -1514,6 +1518,7 @@
 				);
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SKIP_INSTALL = YES;
+				SUPPORTS_MACCATALYST = YES;
```
tomlongridge, line 1521: Is this change required?

robinmacharg: Yes.  This change allows the CI system (which runs MacOS 10.14 - without Catalyst support) to (trivially) find and replace the line with `SUPPORTS_MACCATALYST = NO;` and allow the static library to build under CI conditions.  A manual test of a Catalyst-compatible static library build on 10.15 is also performed during UAT.

fractalwrench:
Closing as this should be addressed by #670 instead

PR #540, opened 2020-04-15
Add redactedKeys for removing sensitive values from metadata

fractalwrench:
## Goal
Adds the `redactedKeys` property to `Configuration` which allows a user to specify sensitive keys. Any sensitive values present in the metadata of `BugsnagEvent` will be replaced by `"[REDACTED]"` when serialized to JSON.

## Changeset
- Added `redactedKeys` property to `Configuration`, which by default is an array of "password"
- Set the `redactedKeys` on each `BugsnagEvent` using the `Configuration` value
- Sanitised the metadata at the point of JSON serialization when 
- `BugsnagEvent` now redacts sensitive values with any key that is considered redacted

## Tests
- Added unit tests to verify the default value of `redactedKeys` and its effect on JSON serialization
- Added tests to macOS/tvOS projects which already existed on the file system but which were not part of the xcodeproj file

### DIFF
--- a/Source/BugsnagConfiguration.h
+++ b/Source/BugsnagConfiguration.h
@@ -107,6 +107,17 @@ typedef NS_OPTIONS(NSUInteger, BSGErrorType) {
  *  Release stages which are allowed to notify Bugsnag
  */
 @property(readwrite, retain, nullable) NSArray *enabledReleaseStages;
+
+/**
+ * Sets which values should be removed from any Metadata objects before
+ * sending them to Bugsnag. Use this if you want to ensure you don't send
+ * sensitive data such as passwords, and credit card numbers to our
+ * servers. Any keys which contain these strings will be filtered.
+ *
+ * By default, redactedKeys is set to "password"
+ */
+@property(readwrite, retain, nullable) NSArray<NSString *> *redactedKeys;
+
 /**
  *  A general summary of what was occuring in the application
  */
--- a/Source/BugsnagConfiguration.h
+++ b/Source/BugsnagConfiguration.h
@@ -112,9 +112,10 @@ typedef NS_OPTIONS(NSUInteger, BSGErrorType) {
  * Sets which values should be removed from any Metadata objects before
  * sending them to Bugsnag. Use this if you want to ensure you don't send
  * sensitive data such as passwords, and credit card numbers to our
- * servers. Any keys which contain these strings will be filtered.
+ * servers. Any keys which contain a match will be filtered.
  *
- * By default, redactedKeys is set to "password"
+ * By default, redactedKeys is set to ["password"]. Both string literals and regex
+ * values can be supplied to this property.
  */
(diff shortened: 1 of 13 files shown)


Source/BugsnagEvent.m
```diff
@@ -308,6 +308,8 @@ - (BOOL)shouldBeSent;
  *  The release stage of the application
  */
 @property(readwrite, copy, nullable) NSString *releaseStage;
+
+@property NSArray *redactedKeys;
```
robinmacharg, line 312: I think the comment should point to the Config comment.

Source/BugsnagConfiguration.h
```diff
@@ -107,6 +107,17 @@ typedef NS_OPTIONS(NSUInteger, BSGErrorType) {
...
  */
 @property(readwrite, retain, nullable) NSArray *enabledReleaseStages;
+
+/**
+ * Sets which values should be removed from any Metadata objects before
+ * sending them to Bugsnag. Use this if you want to ensure you don't send
+ * sensitive data such as passwords, and credit card numbers to our
+ * servers. Any keys which contain these strings will be filtered.
+ *
+ * By default, redactedKeys is set to "password"
```
robinmacharg, line 117: nit: the default value is `[@"password"]`

Source/BugsnagConfiguration.h
```diff
@@ -107,6 +107,17 @@ typedef NS_OPTIONS(NSUInteger, BSGErrorType) {
...
+ */
+@property(readwrite, retain, nullable) NSArray<NSString *> *redactedKeys;
```
robinmacharg, line 119: This will need to be an array of ids to support RegEx matching.

Tests/BugsnagSinkTests.m
```diff
@@ -20,7 +20,7 @@ @interface BugsnagSinkTests : XCTestCase
 @end
 
 @interface BugsnagSink ()
-- (NSDictionary *)getBodyFromReports:(NSArray *)reports;
+- (NSDictionary *)getBodyFromReports:(NSArray *)events;
```
robinmacharg, line 23: This method is only mentioned here and in the test, a total of 6 times, AFAICT.  I'd be happy for it to be renamed to `getBodyFromEvents:` as part of this change; I think it's clear enough housekeeping.

Source/BugsnagConfiguration.h
fractalwrench: Updated docs to reflect this.

Tests/BugsnagSinkTests.m
fractalwrench: Agreed - renamed this method.

Source/BugsnagConfiguration.h
fractalwrench: Updated implementation to support regex testing, added an additional test case.

PR #546, opened 2020-04-17
Add structured app/device fields to BugsnagSession

fractalwrench:
## Goal

Adds structured app/device fields to `BugsnagSession` and passes the structured session object into `OnSessionCallback` to allow users to modify/inspect the payload.

## Changeset
- Updated `OnSessionCallback` to take a `BugsnagSession` parameter rather than a dictionary
- Updated `BugsnagSessionTracker` to generate a structured `BugsnagSession` object
- Renamed `session.sessionId` to `session.id`
- Add `session.app` and `session.device`
- Make `session.user` conform to notifier spec with set method + getter
- Add methods for serializing `BugsnagSession` to JSON
- Add method for deserializing `BugsnagSession` from a cached payload
- Update `[BSG_KSSystemInfo systemInfo]` to include memory information by default, to ensure `device.totalMemory` is populated for sessions
- Updated `SessionTrackingApiClient` logic to avoid batching of session payloads, as the app/device information can now differ between individual sessions
- Made `BugsnagSession` and `BugsnagUser` headers publicly accessible

## Tests
- Added new tests for serialization/deserialization of sessions
- Verified that a `BugsnagSession` object is populated during a callback in the example app


Source/BugsnagSession.h
```diff
@@ -9,24 +9,34 @@
...
-                       handledCount:(NSUInteger)handledCount
-                     unhandledCount:(NSUInteger)unhandledCount;
-
-@property(readonly) NSString *_Nonnull sessionId;
-@property(readonly) NSDate *_Nonnull startedAt;
-@property(readonly) BugsnagUser *_Nullable user;
+@property NSString *_Nonnull id;
+@property NSDate *_Nonnull startedAt;
+@property(readonly) BugsnagApp *_Nonnull app;
+@property(readonly) BugsnagDevice *_Nonnull device;
```
robinmacharg, line 20: I think these two properties can be made private and added to the category on `BugsnagSessionTrackingPayload`

fractalwrench: These properties are mandated as public in the notifier spec so a user can inspect/mutate them in a callback - otherwise hiding them would be the right approach.

PR #549, opened 2020-04-20
Add sendThreads property to `BugsnagConfiguration`

fractalwrench:
## Goal

Adds the `sendThreads` to `BugsnagConfiguration`. This allows a user to record thread information always, for unhandled errors only, or never.

Note that the stacktrace of the crashing thread is always recorded and reported in the thread tab.

## Changeset

- Added publicly visible `BSGThreadSendPolicy` enum and `sendThreads` property
- Passed value of `sendThreads` into KSCrash during its initialisation
- Altered KSCrash to take an int rather than a bool for the `threadTracingEnabled` value
- Added logic to only serialize all threads if the user has specified to do so

## Tests

- Added unit tests to verify default value and copying of thread information in `BugsnagConfiguration`
- Manually verified behaviour in an example app

### DIFF
--- a/Source/BugsnagCrashSentry.m
+++ b/Source/BugsnagCrashSentry.m
@@ -24,12 +24,14 @@ NSUInteger const BSG_MAX_STORED_REPORTS = 12;
         onCrash:(BSGReportCallback)onCrash
 {
     BugsnagSink *sink = [[BugsnagSink alloc] initWithApiClient:apiClient];
-    [BSG_KSCrash sharedInstance].sink = sink;
-    [BSG_KSCrash sharedInstance].introspectMemory = YES;
-    [BSG_KSCrash sharedInstance].deleteBehaviorAfterSendAll =
+    BSG_KSCrash *ksCrash = [BSG_KSCrash sharedInstance];
+    ksCrash.sink = sink;
+    ksCrash.introspectMemory = YES;
+    ksCrash.deleteBehaviorAfterSendAll =
         BSG_KSCDeleteOnSuccess;
-    [BSG_KSCrash sharedInstance].onCrash = onCrash;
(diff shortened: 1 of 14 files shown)


Source/BugsnagCrashSentry.m
```diff
@@ -44,7 +46,7 @@ - (void)install:(BugsnagConfiguration *)config
     
     bsg_kscrash_setHandlingCrashTypes(crashTypes);
     
-    if (![[BSG_KSCrash sharedInstance] install])
+    if (![ksCrash install])
         bsg_log_err(@"Failed to install crash handler. No exceptions will be reported!");
```
robinmacharg, line 50: nit: braces round conditional.  Also, while this is being modified, would a rename to e.g. `tryInstall` read better?  I'm slightly swithering on that.

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry.h
```diff
@@ -60,7 +60,7 @@ typedef struct BSG_KSCrash_SentryContext {
     bool reportWhenDebuggerIsAttached;
 
     /** If true, will trace threads and report binary images. */
```
robinmacharg, line 62: nit: The comment's not quite accurate w.r.t. the enum

fractalwrench: Updated comment.

Source/BugsnagCrashSentry.m
fractalwrench: Added braces - I'll keep the name as `install` for now as it reduces the size of the changeset, and it's something we can alter later if needed.

PR #554, opened 2020-04-21
Add getter for bugsnag.context

fractalwrench:
## Goal

Adds a getter method for context on `Bugsnag`, and converts `context into a property on `BugsnagClient`. This allows users to read the value if necessary which is a requirement in the spec.

### DIFF
--- a/Source/Bugsnag.h
+++ b/Source/Bugsnag.h
@@ -217,11 +217,11 @@
 // =============================================================================
 
 /**
- * Replicates BugsnagConfiguration.context
- *
- * @param context A general summary of what was happening in the application
+ * Retrieves the context - a general summary of what was happening in the application
  */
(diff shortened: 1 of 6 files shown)


Source/Bugsnag.h
```diff
@@ -217,12 +217,12 @@
...
 
 /**
- * Replicates BugsnagConfiguration.context
- *
- * @param context A general summary of what was happening in the application
+ * Retrieves the context - a general summary of what was happening in the application
  */
 + (void)setContext:(NSString *_Nullable)context;
 
++ (NSString *_Nullable)getContext;
```
robinmacharg, line 224: I think this can be just `context`, no `get` required?

fractalwrench: Yep - I've amended this to be `context` instead.

PR #561, opened 2020-04-22
Add BugsnagErrorTypes for controlling captured errors

fractalwrench:
## Goal

Adds the `BugsnagErrorTypes` class which controls which errors are captured by Bugsnag. By default, all error types are captured. This replaces the bitmask which was used in the previous implementation.

## Changeset

- Created "Configuration" group in XCode which logically groups configuration source files together
- Added publicly visible `BugsnagErrorTypes` class
- Made existing implementation use new syntax

## Tests

Updated existing test coverage to use new syntax.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Bugsnag Notifiers on other platforms.
 
 ## Enhancements
 
+* Add `BugsnagErrorTypes` for controlling captured errors
+  [#561](https://github.com/bugsnag/bugsnag-cocoa/pull/561)
+
 * Rename `BugsnagUser` properties
   [#560](https://github.com/bugsnag/bugsnag-cocoa/pull/560)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,6 @@ Bugsnag Notifiers on other platforms.
 
 ## Enhancements
 
-* Add `BugsnagErrorTypes` for controlling captured errors
-  [#561](https://github.com/bugsnag/bugsnag-cocoa/pull/561)
-
 * Rename `BugsnagUser` properties
   [#560](https://github.com/bugsnag/bugsnag-cocoa/pull/560)
 
@@ -212,10 +209,11 @@ Bugsnag Notifiers on other platforms.
 * Add metadata accessor methods to `BugsnagEvent`
   [#465](https://github.com/bugsnag/bugsnag-cocoa/pull/465)
   
-* Added a user-configurable `enabledErrorTypes` bitfield property to `BugsnagConfiguration`.
-  This allows users to choose which types of events are reported.  If automatic crash detection
+* Added a user-configurable `enabledErrorTypes` property to `BugsnagConfiguration`.
+  The `BugsnagErrorTypes` property allows users to choose which types of events are reported.  If automatic crash detection
   is disabled this value is ignored.  User-generated `notify()` events are reported in all cases.
   [#477](https://github.com/bugsnag/bugsnag-cocoa/pull/477)
+  [#561](https://github.com/bugsnag/bugsnag-cocoa/pull/561)
 
 * Internal logging has been unified.  Where before two preprocessor macros were
   required to configure both `Bugsnag` and `KSCrash` portions, now the Bugsnag
--- a/Source/BSGOutOfMemoryWatchdog.m
+++ b/Source/BSGOutOfMemoryWatchdog.m
@@ -13,6 +13,7 @@
 #import "Bugsnag.h"
 #import "BugsnagSessionTracker.h"
 #import "Private.h"
+#import "BugsnagErrorTypes.h"
 
 @interface BSGOutOfMemoryWatchdog ()
 @property(nonatomic, getter=isWatching) BOOL watching;
@@ -199,7 +200,7 @@
             BOOL sameVersions = [lastBootOSVersion isEqualToString:osVersion] &&
                                 [lastBootBundleVersion isEqualToString:bundleVersion] &&
                                 [lastBootAppVersion isEqualToString:appVersion];
-            BOOL shouldReport = (config.enabledErrorTypes & BSGErrorTypesOOMs)
+            BOOL shouldReport = (config.enabledErrorTypes.OOMs)
                 && (lastBootInForeground && lastBootWasActive);
             [self deleteSentinelFile];
             return sameVersions && shouldReport;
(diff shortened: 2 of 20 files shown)


Source/BSGOutOfMemoryWatchdog.m
```diff
@@ -199,7 +200,7 @@ - (BOOL)computeDidOOMLastLaunchWithConfig:(BugsnagConfiguration *)config {
             BOOL sameVersions = [lastBootOSVersion isEqualToString:osVersion] &&
                                 [lastBootBundleVersion isEqualToString:bundleVersion] &&
                                 [lastBootAppVersion isEqualToString:appVersion];
-            BOOL shouldReport = (config.enabledErrorTypes & BSGErrorTypesOOMs)
+            BOOL shouldReport = (config.enabledErrorTypes.OOMs)
```
tomlongridge, line 203: Is it correct that we're only checking enabledErrorTypes and not autoDetectErrors?

Source/BugsnagConfiguration.m
```diff
@@ -218,9 +216,9 @@ - (instancetype _Nonnull)initWithApiKey:(NSString *_Nonnull)apiKey
     // persistUser isn't settable until post-init.
     _user = [self getPersistedUserData];
     [self setUserMetadataFromUser:_user];
-    
+
     #if !DEBUG
-        _enabledErrorTypes |= BSGErrorTypesOOMs;
+        _enabledErrorTypes.OOMs = true;
```
tomlongridge, line 221: Do we need this if you're setting it in the `BugsnagErrorTypes`'s `init` function too?

Source/BugsnagClient.m
```diff
@@ -1097,7 +1097,7 @@ - (void)updateCrashDetectionSettings {
     if (self.configuration.autoDetectErrors) {
         // Enable all crash detection
         bsg_kscrash_setHandlingCrashTypes(BSG_KSCrashTypeAll);
```
tomlongridge, line 1099: I believe this should be checking the enabledErrorTypes as well, rather than enabling all. However, does it even need to update KSCrash now that configuration can't change?

Source/BugsnagErrorTypes.h
```diff
@@ -0,0 +1,49 @@
...
+ * This flag is true by default.
+ */
+@property BOOL signals;
+
+/**
+ * Determines whether C errors should be reported to bugsnag.
+ *
+ * This flag is true by default.
+ */
+@property BOOL C;
```
tomlongridge, line 39: Are we justified in calling this C rather than C++? It seems correct, but not sure why the KS constant is `BSG_KSCrashTypeCPPException`?

Source/BugsnagErrorTypes.h
```diff
@@ -0,0 +1,49 @@
...
+#import <Foundation/Foundation.h>
+
+@interface BugsnagErrorTypes : NSObject
+
+/**
+ * Determines whether Out of Memory events should be reported to bugsnag.
+ *
+ * This flag is true by default.
+ */
+@property BOOL OOMs;
```
tomlongridge, line 18: There's an unfortunate set of casing in these fields! I can see why you've chosen these, but is there anything we can do to improve the naming?

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ Bugsnag Notifiers on other platforms.
 
 ## Enhancements
 
+* Add `BugsnagErrorTypes` for controlling captured errors
```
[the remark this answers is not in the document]
fractalwrench: Combined changelog entries.

Source/Bugsnag.h
```diff
@@ -36,6 +36,7 @@
 #import "BugsnagDeviceWithState.h"
 #import "BugsnagEndpointConfiguration.h"
 #import "BugsnagError.h"
+#import "BugsnagErrorTypes.h"
```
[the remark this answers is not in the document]
fractalwrench: This import is required as `Bugsnag.h` is an umbrella header. This means it needs to import all public header files so that users can access them when importing the module.

Source/BugsnagErrorTypes.h
fractalwrench: Agreed to rename as the following for now:

- ooms
- unhandledExceptions
- signals
- cppExceptions
- machExceptions

Source/BugsnagErrorTypes.h
fractalwrench: Addressed by other comments

Source/BugsnagConfiguration.m
fractalwrench: nope, this can go.

Source/BSGOutOfMemoryWatchdog.m
fractalwrench: Nope, this is a redundant check which I'll remove: https://github.com/bugsnag/bugsnag-cocoa/blob/c3065a78666c13213f6a96f647caac83ea808fc0/Source/BugsnagClient.m#L534

Source/BugsnagClient.m
fractalwrench: You're right - there's no need to update, it looks like this method is obsolete now that configuration is immutable. I've removed this as `BugsnagCrashSentry` performs the calculation (and correctly checks enabledErrorTypes)

Source/BugsnagConfiguration.m
```diff
@@ -520,22 +514,6 @@ - (void)setReleaseStage:(NSString *)newReleaseStage {
...
-- (void)setAutoDetectErrors:(BOOL)autoDetectErrors {
-    if (autoDetectErrors == _autoDetectErrors) {
-        return;
-    }
-    [self willChangeValueForKey:NSStringFromSelector(@selector(autoDetectErrors))];
-    _autoDetectErrors = autoDetectErrors;
-    [[Bugsnag client] updateCrashDetectionSettings];
-    [self didChangeValueForKey:NSStringFromSelector(@selector(autoDetectErrors))];
-}
-
```
tomlongridge, line 538: Is this removing the getter/setting altogether? This still needs to be settable by the user to pass it to `start`.

Also, should we remove the `autoNotify` backwards compatible property below?

fractalwrench: The property is declared on the header and Objective-C will automatically generate accessors for it. This is just overriding those accessors to include side-effects which are no longer required, so this won't have any impact on users being able to set the value.

I'll remove the `autoNotify` methods as they're not required.

PR #564, opened 2020-04-23
Add onCrashHandler data to BugsnagEvent metadata

fractalwrench:
## Goal

Data which is added in the [onCrashHandler](https://docs.bugsnag.com/platforms/ios/configuration-options/#oncrashhandler) is now added to error reports in an `onCrash` metadata tab. This provides a replacement for the previous API which allowed for these values to be inspected (but not reported) in the now removed `rawData` parameter.

Screenshot 2020-04-23 at 10 14 38

## Changeset

- Read `user` information when constructing a `BugsnagEvent` from a KSCrash report
- Removed other keys which can be serialized into the `user` object that are not user-added data
- Added any remaining information to an `onCrash` metadata tab

## Tests

Added a unit test to verify behaviour of `onCrash` data parsing. Manually tested `onCrashHandler` functionality in an example app and verified the value was sent to the dashboard.

### DIFF
--- a/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReportFields.h
+++ b/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReportFields.h
@@ -174,7 +174,7 @@
 #define BSG_KSCrashField_Threads "threads"
 #define BSG_KSCrashField_User "user"
 #define BSG_KSCrashField_UserAtCrash "user_atcrash"
-#define BSG_KSCrashField_OnCrash "onCrash"
+#define BSG_KSCrashField_OnCrashMetadataSectionName "onCrash"
 
 #pragma mark Incomplete
 #define BSG_KSCrashField_Incomplete "incomplete"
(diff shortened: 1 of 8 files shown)


Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReportFields.h
```diff
@@ -174,6 +174,7 @@
 #define BSG_KSCrashField_Threads "threads"
 #define BSG_KSCrashField_User "user"
 #define BSG_KSCrashField_UserAtCrash "user_atcrash"
+#define BSG_KSCrashField_OnCrash "onCrash"
```
tomlongridge, line 177: Constant name should describe the intent not the value of the constant.

```suggestion
#define BSG_KSCrashField_OnCrashMetadataSectionName "onCrash"
```

Tests/BugsnagOnCrashTest.m
```diff
@@ -0,0 +1,99 @@
...
+
+@interface BugsnagOnCrashTest : XCTestCase
+
+@end
+
+@implementation BugsnagOnCrashTest
+
+- (void)testEmptyData {
+    BugsnagEvent *event = [[BugsnagEvent alloc] initWithKSReport:@{}];
+    XCTAssertNil([event getMetadataFromSection:@"onCrash"]);
```
tomlongridge, line 25: Use constant for key

fractalwrench: Agreed to keep using string literal as this a public field, so we want to verify the constant name itself is correct and doesn't change over time

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReportFields.h
fractalwrench: Renamed constant

PR #568, opened 2020-04-28
Import Bugsnag headers before reading preprocessor defines

bolsinga:
Summary:
- This change is a little different than https://github.com/bugsnag/bugsnag-react-native/pull/455 in that it just has .m files import their header files first thing.
- Also do not query `TARGET_OS_MAC` first as that is true for `TARGET_IOS_SIMULATOR`.

Test Plan:
- Debugging in Bugsnag for this code.
- I added a compiler error in every `TARGET_OS_IPHONE` conditional and ensured that they all worked. This is how I found the ones found in this diff.

## Goal

Fix building Bugsnag via Bazel

## Design

Ensure class headers are imported first, which brings in Foundation, which brings in TargetConditionals.h. This is typical code structure.

## Changeset

Change the imports of files so that preprocessor macros are defined when read.

## Tests

I added nonsense characters, to ensure that before the change would show that code wrapped in macro was not compiling, and that after the change it was. At runtime I was able to step into the `BSGOutOfMemoryWatchdog` which is usually preprocessor wrapped with `BSGOOMAvailable`, and it wasn't compiling for us.

### Outstanding Questions

  - [x] Final review
  - [ ] Release

- [x] The correct target branch has been selected (`master` for fixes, `next` for
  features)

### DIFF
--- a/Source/BSGOutOfMemoryWatchdog.m
+++ b/Source/BSGOutOfMemoryWatchdog.m
@@ -1,3 +1,5 @@
+#import "BSGOutOfMemoryWatchdog.h"
+
 #if (TARGET_OS_TV || TARGET_OS_IPHONE)
 #define BSGOOMAvailable 1
 #else
@@ -7,7 +9,6 @@
 #if BSGOOMAvailable
 #import <UIKit/UIKit.h>
 #endif
-#import "BSGOutOfMemoryWatchdog.h"
 #import "BSG_KSSystemInfo.h"
 #import "BugsnagLogger.h"
 #import "Bugsnag.h"
--- a/Source/BugsnagCrashReport.m
+++ b/Source/BugsnagCrashReport.m
@@ -6,8 +6,9 @@
 //
 //
 
-#if TARGET_OS_MAC || TARGET_OS_TV
-#elif TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
+#import "BugsnagCrashReport.h"
+
+#if TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
 #import <UIKit/UIKit.h>
 #include <sys/utsname.h>
 #endif
--- a/Source/BugsnagNotifier.m
+++ b/Source/BugsnagNotifier.m
@@ -458,8 +458,7 @@ NSString *const kAppWillTerminate = @"App Will Terminate";
     [[NSNotificationCenter defaultCenter] removeObserver:self];
     [self.networkReachable stopWatchingConnectivity];
 
-#if TARGET_OS_TV || TARGET_OS_MAC
-#elif TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
+#if TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
     [UIDevice currentDevice].batteryMonitoringEnabled = NO;
     [[UIDevice currentDevice] endGeneratingDeviceOrientationNotifications];
 #endif
--- a/Source/BugsnagNotifier.m
+++ b/Source/BugsnagNotifier.m
@@ -458,7 +458,8 @@ NSString *const kAppWillTerminate = @"App Will Terminate";
     [[NSNotificationCenter defaultCenter] removeObserver:self];
     [self.networkReachable stopWatchingConnectivity];
 
-#if TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
+#if TARGET_OS_TV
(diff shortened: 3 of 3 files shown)


jerrymarino:
@bolsinga  I didn't see how this code would work in Xcode but not Bazel. As far as I can tell, this is broken for all bugsnag users. Would you mind clarifying so we can make sure to isolate any other issues related to this? We should 100% support building bugsnag with PodToBUILD and Bazel.

Anyhow, I submitted bugsnag/bugsnag-react-native#455 I don't have a strong preference on how what the actual fix is.

bolsinga:
AFAIK it works because somehow the code must be seeing the macros defined when it has been tested and used by others. We know others successfully see OOM reports from Bugsnag. Our build system exposed this issue. However I do not think it is solely a build system issue nor a code issue. I think the combination of the two shows these files are not importing their own headers first. When they do this they automatically get Foundation and TargetConditionals.

I think this fix is simple and it is atypical to manually import TargetConditionals.

mattdyoung:
@jerrymarino @bolsinga 
I can see you've both been discussing these build problems, both on this PR and on https://github.com/bugsnag/bugsnag-react-native/pull/455.

Are these both aiming to solve the same problem or are both required? What ultimately are these issues blocking you from achieving and is this an issue for bugsnag-cocoa or bugsnag-react-native or both?

bolsinga:
> @jerrymarino @bolsinga
> I can see you've both been discussing these build problems, both on this PR and on [bugsnag/bugsnag-react-native#455](https://github.com/bugsnag/bugsnag-react-native/pull/455).
> 
> Are these both aiming to solve the same problem or are both required? What ultimately are these issues blocking you from achieving and is this an issue for bugsnag-cocoa or bugsnag-react-native or both?

We are both aiming to solve the same problem.

We import bugsnag-react-native into our application. As you know, bugsnag-cocoa is a dependency. So we need the fix in both, but it probably ought to apply here.

We use https://github.com/pinterest/PodToBUILD to bring Bugsnag-React-Native into our application which builds with Bazel. Under these circumstances, it is true that the TARGET_OS_IPHONE macro is not defined when the OOM watchdog is going to define BSGOOMAvailable macro. This fix just has the .m files import their .h files first thing, and everything works out. Thanks.

mattdyoung:
@bolsinga - thanks! We're taking a look at both PRs and will decide on the best way forward.

jerrymarino:
Ah ok cool, I just wanted to confirm that this bug wasn't related to PodToBUILD or using Bazel 👍

mattdyoung:
Hi @jerrymarino @bolsinga 
Yes, the macros we're using look to be incorrect in some cases and we'd like to investigate fully.

We're just working towards a new major release of bugsnag-cocoa in the coming weeks, so will be looking to resolve this as part of that work. If you need the fix in the short term I'd suggest using a fork with this PR.

bolsinga:
Thanks for the update. We currently have a local workaround and look forward to the real fix. 

Would you be able to verify if the oom watchdog builds for other iOS clients?

johnkiely1:
Hi @bolsinga,

Im not sure I understand what you mean by your question
> Would you be able to verify if the oom watchdog builds for other iOS clients?
Could you explain further on this please?

bolsinga:
Sure. For the way we build this code, the OOM watch dog code wasn’t compiling due to it referencing the OS macros prior to the source importing a header that defines these macros first. 

I assume that this code builds and works for other iOS clients. I haven’t personally confirmed this. My question was basically to confirm that bugsnag does get oom reports from other client code, and thus we’d know that the build issue we see here is unique to our situation.

johnkiely1:
Thanks for clarifying that. Yes you assume correctly, it builds and works fine for other iOS clients.

bolsinga:
Thanks! Our current workaround in our build system is to explicitly import TargetConditionals.h for every BugSnag file. 

That workaround is not necessary with this PR applied. 

I think this PR is a simple logical cleanup, but our workaround will hold up until this PR lands. 

Thanks again for your time and understanding.

mattdyoung:
We believe this issue has now been fixed in https://github.com/bugsnag/bugsnag-cocoa/pull/654, released in [v6.0.0](https://github.com/bugsnag/bugsnag-cocoa/releases/tag/v6.0.0).

Note: There are a number of breaking changes in v6.0.0 as part of an effort to improve the interface and make it more consistent across our error reporting libraries, so please refer to the upgrade guide:
https://github.com/bugsnag/bugsnag-cocoa/blob/master/UPGRADING.md

PR #569, opened 2020-04-29
Add BugsnagNotifier for payload serialization

fractalwrench:
## Goal

Refactors the serialization of `event.notifier` and `session.notifier` so that a structured class, `BugsnagNotifier`, holds this information. This also adds the new `notifier.dependencies` field, which allows for dependent notifiers to be added to this information. These changes will be required for the React Native refresh.

## Changeset
- Added structured class `BugsnagNotifier`
- Replaced previous implementation on `BugsnagClient`
- Created `toDict` method for JSON serialization

## Tests
Added unit tests for default values/serialization of `BugsnagNotifier` class


Source/BugsnagNotifier.m
```diff
@@ -0,0 +1,50 @@
...
+        self.name =
+#if TARGET_OS_TV
+                @"tvOS Bugsnag Notifier";
+#elif TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
+                @"iOS Bugsnag Notifier";
+#elif TARGET_OS_MAC
+        @"OSX Bugsnag Notifier";
+#else
+            @"Bugsnag Objective-C";
+#endif
```
tomlongridge, line 25: Nit: indentation

Source/BugsnagNotifier.m
```diff
@@ -0,0 +1,50 @@
...
+        self.version = @"5.23.0";
```
tomlongridge, line 26: We need to ensure the Makefile works to set the version (it appears to be broken on v6 as it looks for `NOTIFIER_VERSION` in this file (from it's previous purpose as the client-like class).

Suggest you move `NOTIFIER_VERSION` and `NOTIFIER_URL` to this file?

fractalwrench: Updated the Makefile to alter the version on this line instead. I've removed the unused `NOTIFIER_VERSION` and `NOTIFIER_URL` constants.

PR #577, opened 2020-05-01
Fix potential deadlock when generating a crash report

robinmacharg:
## Goal

An [issue](https://github.com/bugsnag/bugsnag-react-native/issues/448) has been identified that can lead to an app deadlocking when reporting a crash.

The sequence of events is:

1. Some program component accesses `dyld` and takes a lock out.
2. A crash is reported.
3. Bugsnag, as part of its normal crash reporting behaviour, suspends all but a few critical threads, including the thread that holds the `dyld` lock.
4. Bugsnag attempts to read the Mach binary images loaded into the app.  This also attempts to take out a lock but is unable to, leading to deadlock.

This PR fixes this issue.

## Design

There are two main mechanisms by which a program may find out about `dyld` behaviour: by direct querying (e.g. `_dyld_get_image_header()` etc., used by the KSCrash component of Bugsnag prior to this issue) and by callback (`_dyld_register_func_for_add_image()` etc.).  This change modifies the KSCrash component to use callbacks and caches information about loaded binary images for when a crash does occur.  Thus, when threads are suspended and a crash report is generated no additional calls to `dyld` need to be made and deadlock is avoided.

## Changeset

`dyld` binary image capture was moved from `BSG_KSCrashReport.c` to `BSG_KSCrash.m`.  A static cache is used.  An additional `.m` file contains functions to ease bridging between Objective C and C.

## Tests

__TBD__

### Outstanding Questions

  - [ ] Final review
  - [ ] Release

  features)

### DIFF
--- a/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
+++ b/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
@@ -24,6 +24,7 @@
 // THE SOFTWARE.
 //
 
+#import <mach-o/dyld.h>
 #import "BSG_KSCrashAdvanced.h"
 
 #import "BSG_KSCrashC.h"
@@ -31,6 +32,7 @@
 #import "BSG_KSJSONCodecObjC.h"
 #import "BSG_KSSingleton.h"
 #import "BSG_KSSystemCapabilities.h"
+#import "BSG_KSMachHeaders.h"
 #import "NSError+BSG_SimpleConstructor.h"
 
 //#define BSG_KSLogger_LocalLevel TRACE
@@ -226,6 +228,9 @@ IMPLEMENT_EXCLUSIVE_SHARED_INSTANCE(BSG_KSCrash)
         return false;
     }
 
+    // Maintain a cache of info about dynamically loaded binary images
+    [self listenForLoadedBinaries];
+    
 #if BSG_KSCRASH_HAS_UIKIT
     NSNotificationCenter *nCenter = [NSNotificationCenter defaultCenter];
     [nCenter addObserver:self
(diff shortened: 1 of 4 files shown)


Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
```diff
@@ -253,6 +258,20 @@ - (BOOL)install {
...
+/**
+ * Set up listeners for un/loaded frameworks.  Maintaining our own list of framework Mach
+ * headers means that we avoid potential deadlock situations where we try and suspend
+ * lock-holding threads prior to loading mach headers as part of our normal event handling
+ * behaviour.
+ */
+- (void)listenForLoadedBinaries {
+    bsg_initialise_mach_binary_headers();
+    _dyld_register_func_for_add_image(&bsg_mach_binary_image_added);
+    _dyld_register_func_for_remove_image(&bsg_mach_binary_image_removed);
```
fractalwrench, line 270: Potentially worth registering `bsg_mach_binary_image_removed` first. This avoids the case where some binary image is removed on another thread _while_ the `bsg_mach_binary_image_added` callback is executed for the first time.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
```diff
@@ -226,6 +228,9 @@ - (BOOL)install {
         return false;
     }
 
+    // Maintain a cache of info about dynamically loaded binary images
+    [self listenForLoadedBinaries];
```
fractalwrench, line 232: Potentially worth putting this before `bsg_kscrash_install`, or at least initialising an empty array of binary images to prevent against the case where a crash occurs during/at the same time as initialisation.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,190 @@
...
+{
+    BSG_Mach_Binary_Image_Info info = { 0 };
+    if (populate_info(header, &info)) {
+        [bsg_mach_binary_images_info addObject:[NSValue valueWithBytes:&info
+                                                          objCType:@encode(BSG_Mach_Binary_Image_Info)]];
+    }
+}
+
+/**
+ * Compare two mach biunary image structures
```
fractalwrench, line 119: spelling nit: biunary ->binary

Source/KSCrash/Source/KSCrash/Recording/BSG_KSMachHeaders.h
```diff
@@ -0,0 +1,51 @@
...
+ */
+typedef struct {
+    const struct mach_header *mh;      /* The mach_header - 32 or 64 bit */
+    uint64_t imageVmAddr;
+    uint64_t imageSize;
+    uint8_t *uuid;
+    const char* name;
+    cpu_type_t cputype;          /* cpu specifier */
+    cpu_subtype_t cpusubtype;    /* machine specifier */
+} BSG_Mach_Binary_Image_Info;
```
fractalwrench, line 24: Encapsulating this info in a struct works well 👍

Source/KSCrash/Source/KSCrash/Recording/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * @param header A mach_header structure
+ *
+ * @param slide The VM offset of the binary image
+ */
+void bsg_mach_binary_image_added(const struct mach_header *header, intptr_t slide)
+{
+    BSG_Mach_Binary_Image_Info info = { 0 };
+    if (populate_info(header, &info)) {
+        [bsg_mach_binary_images_info addObject:[NSValue valueWithBytes:&info
```
fractalwrench, line 113: Using an Objective-C object makes me a bit nervous - as if `dyld` continues to execute this callback during a signal crash, then it could hang the process.

Would it be possible to initialise a C array to start with, then allocate more memory as required?

Source/KSCrash/Source/KSCrash/Recording/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,190 @@
...
+        info1->mh->flags == info2->mh->flags &&
+        info1->mh->magic == info2->mh->magic &&
+        info1->mh->ncmds == info2->mh->ncmds &&
+        info1->mh->sizeofcmds == info2->mh->sizeofcmds;
+}
+
+/**
+ * Called when a binary image is unloaded.
+ */
+void bsg_mach_binary_image_removed(const struct mach_header *header, intptr_t slide)
```
fractalwrench, line 138: Will this method need any form of synchronization if images are loaded/unloaded on different threads? Or is that a non-issue?

Source/KSCrash/Source/KSCrash/Recording/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,190 @@
...
+                break;
+            }
+        }
+    }
+}
+
+/**
+ * Returns a C array of structs describing the loaded Mach binaries
+ */
+BSG_Mach_Binary_Image_Info* bsg_mach_header_array(size_t *count) {
```
fractalwrench, line 163: My gut feel is that this function could be done away with if the structs are stored as a C array

robinmacharg:
This PR is superseded by #580.  All noted issues have been addressed there.

PR #578, opened 2020-05-04
Allow enhancing default runtimeVersions info

fractalwrench:
## Goal

Allows for downstream notifiers such as bugsnag-react-native to set the runtime version information recorded for events and sessions. This changeset largely mirrors the Android implementation.

## Changeset

- Created method on `Bugsnag` and `BugsnagClient` for adding runtime version information
- Updated the session tracker + client implementation to hold extra runtime version info when it is added
- Whenever a new session/error is captured, the extra runtime version info is added

## Tests
Verified with a debugger that the object passed to an onSession/onError block contains the appropriate runtime version information.

Note that there is a known issue with the Cocoa implementation where the Event is not fully populated/persisted, so runtimeVersions information will not be sent in the actual payload. This will be addressed by a separate internal ticket.


fractalwrench:
Updated to add a unit test to verify that the `BugsnagDevice` appends any extra runtime info to the `runtimeVersions` property.

PR #580, opened 2020-05-04
Fix potential deadlock when generating a crash report

robinmacharg:
[Note: this PR supersedes #577, and incorporates feedback from initial review.] 

## Goal

An [issue](https://github.com/bugsnag/bugsnag-react-native/issues/448) has been identified that can lead to an app deadlocking when reporting a crash.

The sequence of events is:

1. Some program component accesses `dyld` and takes a lock out.
2. A crash is reported.
3. Bugsnag, as part of its normal crash reporting behaviour, suspends all but a few critical threads, including the thread that holds the `dyld` lock.
4. Bugsnag attempts to read the Mach binary images loaded into the app.  This also attempts to take out a lock but is unable to, leading to deadlock.

This PR fixes this issue.

## Design

There are two main mechanisms by which a program may find out about `dyld` behaviour: by direct querying (e.g. `_dyld_get_image_header()` etc., used by the KSCrash component of Bugsnag prior to this issue) and by callback (`_dyld_register_func_for_add_image()` etc.).  This change modifies the KSCrash component to use callbacks and caches information about loaded binary images for when a crash does occur.  Thus, when threads are suspended and a crash report is generated no additional calls to `dyld` need to be made and deadlock is avoided.

Thread safety is a concern.  Reading [the released dyld source code](https://github.com/opensource-apple/dyld/blob/3f928f32597888c5eac6003b9199d972d49857b5/src/dyld.cpp#L526) implies that access to the internal binary image storage in modern OS releases is mediated by a locking mechanism.  However to date no documented guarantee of this has been found.  For the avoidance of doubt the KSCrash-level `add/remove` callbacks use a legacy-compatible shared lock to ensure synchronisation.  By the time a crash report is being written all other non-KS threads will have been suspended and so no additional locking is required.

Performance is also a concern, especially given that the proposed solution moves processing from post-crash (where users are likely to be less concerned about a couple of lost seconds) to up-front.  The library startup code was crudely timed in the sample iOS app on a real device (iPhone XS), with results as follows:

```objc
    // Placed in AppDelegate
    CFTimeInterval start = CACurrentMediaTime();
    [Bugsnag startBugsnagWithApiKey:@"<API KEY>"];
    CFTimeInterval end = CACurrentMediaTime();
    NSLog(@"Took: %f", end - start);
```

|     | Without fix | With fix  |
|-----|-------------|-----------|
| 1   | 0.010215    | 0.011997  |
| 2   | 0.014457    | 0.011432  |
| 3   | 0.011843    | 0.015702  |
| 4   | 0.014623    | 0.013538  |
| 5   | 0.011528    | 0.01601   |
| 6   | 0.015008    | 0.017804  |
| 7   | 0.01258     | 0.015619  |
| 8   | 0.012356    | 0.012655  |
| 9   | 0.011991    | 0.014726  |
| 10  | 0.012292    | 0.013338  |
| __Avg__ | __0.0126893__   | __0.0142821__ |

i.e. a difference of ~2ms.  This was for 385 binary images.  (Anecdotally, turning on debugging adds ~0.1s to the startup time).

## Changeset

`dyld` binary image capture was moved from `BSG_KSCrashReport.c` to `BSG_KSCrash.m`.  A file-static dynamic array cache is used.  An additional `.m` file contains functions to ease bridging between Objective C and C.

Bugsnag/KSCrash makes use of `dyld` in two main - but related -  places: `BSG_KSCrashReport.c` writing and `BSG_KSDynamicLinker.c`.  These were modified to use an internal API similar to that presented by `dyld` that accesses the cache of headers.

## Tests

- A dynamic C array is used to record loaded binary images.  Its behaviour is tested.
- Manual performance testing has been carried out to ascertain the impact on app startup times of recording binary images.
- The generated `json` reports are have been manually tested.
- Both symbolicated and unsymbolicated crash reports need checked on the dashboard for correctness.

### Outstanding Questions

  - [X] Final review
  - [ ] Release

  features)

### DIFF
--- a/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
+++ b/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
@@ -36,6 +36,7 @@
 #include "BSG_KSObjC.h"
 #include "BSG_KSSignalInfo.h"
 #include "BSG_KSString.h"
+#include "BSG_KSMachHeaders.h"
 
 //#define BSG_kSLogger_LocalLevel TRACE
 #include "BSG_KSLogger.h"
@@ -1084,70 +1085,21 @@ int bsg_kscrw_i_threadIndex(const thread_t thread) {
  *
  * @param key The object key, if needed.
  *
- * @param index Which image to write about.
+ * @param info Cached info about the binary image.
  */
 void bsg_kscrw_i_writeBinaryImage(const BSG_KSCrashReportWriter *const writer,
-                                  const char *const key, const uint32_t index) {
-    const struct mach_header *header = _dyld_get_image_header(index);
-    if (header == NULL) {
-        return;
-    }
-
-    uintptr_t cmdPtr = bsg_ksdlfirstCmdAfterHeader(header);
-    if (cmdPtr == 0) {
-        return;
-    }
-
-    // Look for the TEXT segment to get the image size.
-    // Also look for a UUID command.
-    uint64_t imageSize = 0;
-    uint64_t imageVmAddr = 0;
-    uint8_t *uuid = NULL;
-
-    for (uint32_t iCmd = 0; iCmd < header->ncmds; iCmd++) {
-        struct load_command *loadCmd = (struct load_command *)cmdPtr;
-        switch (loadCmd->cmd) {
-        case LC_SEGMENT: {
-            struct segment_command *segCmd = (struct segment_command *)cmdPtr;
-            if (strcmp(segCmd->segname, SEG_TEXT) == 0) {
-                imageSize = segCmd->vmsize;
-                imageVmAddr = segCmd->vmaddr;
-            }
-            break;
-        }
-        case LC_SEGMENT_64: {
-            struct segment_command_64 *segCmd =
-                (struct segment_command_64 *)cmdPtr;
-            if (strcmp(segCmd->segname, SEG_TEXT) == 0) {
-                imageSize = segCmd->vmsize;
-                imageVmAddr = segCmd->vmaddr;
-            }
-            break;
-        }
-        case LC_UUID: {
-            struct uuid_command *uuidCmd = (struct uuid_command *)cmdPtr;
-            uuid = uuidCmd->uuid;
-            break;
-        }
-        }
-        cmdPtr += loadCmd->cmdsize;
-    }
-
+                                  const char *const key,
+                                  const BSG_Mach_Binary_Image_Info *info)
+{
     writer->beginObject(writer, key);
     {
-        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageAddress,
-                                   (uintptr_t)header);
-        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageVmAddress,
-                                   imageVmAddr);
-        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageSize,
-                                   imageSize);
-        writer->addStringElement(writer, BSG_KSCrashField_Name,
-                                 _dyld_get_image_name(index));
-        writer->addUUIDElement(writer, BSG_KSCrashField_UUID, uuid);
-        writer->addIntegerElement(writer, BSG_KSCrashField_CPUType,
-                                  header->cputype);
-        writer->addIntegerElement(writer, BSG_KSCrashField_CPUSubType,
-                                  header->cpusubtype);
+        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageAddress, (uintptr_t)info->mh);
+        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageVmAddress,          info->imageVmAddr);
+        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageSize,               info->imageSize);
+        writer->addStringElement(writer, BSG_KSCrashField_Name,                      info->name);
+        writer->addUUIDElement(writer, BSG_KSCrashField_UUID,                        info->uuid);
+        writer->addIntegerElement(writer, BSG_KSCrashField_CPUType,                  info->cputype);
+        writer->addIntegerElement(writer, BSG_KSCrashField_CPUSubType,               info->cpusubtype);
     }
     writer->endContainer(writer);
 }
@@ -1159,13 +1111,16 @@ void bsg_kscrw_i_writeBinaryImage(const BSG_KSCrashReportWriter *const writer,
  * @param key The object key, if needed.
  */
 void bsg_kscrw_i_writeBinaryImages(const BSG_KSCrashReportWriter *const writer,
-                                   const char *const key) {
-    const uint32_t imageCount = _dyld_image_count();
+                                   const char *const key)
+{
+    // Get the array of loaded binary images and a count
+    size_t imageCount;
+    BSG_Mach_Binary_Image_Info *info = bsg_mach_header_array(&imageCount);
 
     writer->beginArray(writer, key);
     {
         for (uint32_t iImg = 0; iImg < imageCount; iImg++) {
-            bsg_kscrw_i_writeBinaryImage(writer, NULL, iImg);
+            bsg_kscrw_i_writeBinaryImage(writer, NULL, &info[iImg]);
         }
     }
     writer->endContainer(writer);
--- a/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
+++ b/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
@@ -1085,21 +1085,23 @@ int bsg_kscrw_i_threadIndex(const thread_t thread) {
  *
  * @param key The object key, if needed.
  *
- * @param info Cached info about the binary image.
+ * @param index Cached info about the binary image.
  */
 void bsg_kscrw_i_writeBinaryImage(const BSG_KSCrashReportWriter *const writer,
                                   const char *const key,
-                                  const BSG_Mach_Binary_Image_Info *info)
+                                  const uint32_t index)
 {
+    BSG_Mach_Binary_Image_Info info = *bsg_dyld_get_image_info(index);
+    
     writer->beginObject(writer, key);
     {
-        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageAddress, (uintptr_t)info->header);
-        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageVmAddress,          info->imageVmAddr);
-        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageSize,               info->imageSize);
-        writer->addStringElement(writer, BSG_KSCrashField_Name,                      info->name);
-        writer->addUUIDElement(writer, BSG_KSCrashField_UUID,                        info->uuid);
-        writer->addIntegerElement(writer, BSG_KSCrashField_CPUType,                  info->header->cputype);
-        writer->addIntegerElement(writer, BSG_KSCrashField_CPUSubType,               info->header->cpusubtype);
+        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageAddress, (uintptr_t)info.header);
+        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageVmAddress,          info.imageVmAddr);
+        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageSize,               info.imageSize);
+        writer->addStringElement(writer, BSG_KSCrashField_Name,                      info.name);
+        writer->addUUIDElement(writer, BSG_KSCrashField_UUID,                        info.uuid);
+        writer->addIntegerElement(writer, BSG_KSCrashField_CPUType,                  info.header->cputype);
+        writer->addIntegerElement(writer, BSG_KSCrashField_CPUSubType,               info.header->cpusubtype);
     }
     writer->endContainer(writer);
 }
@@ -1114,13 +1116,12 @@ void bsg_kscrw_i_writeBinaryImages(const BSG_KSCrashReportWriter *const writer,
                                    const char *const key)
 {
     const uint32_t imageCount = (uint32_t)bsg_dyld_image_count();
-    
+
     writer->beginArray(writer, key);
     {
-        for (uint32_t i = 0; i < imageCount; i++) {
+        for (uint32_t iImg = 0; iImg < imageCount; iImg++) {
             // Threads are suspended at this point; no need to synchronise/lock
-            BSG_Mach_Binary_Image_Info *info = bsg_dyld_get_image_info(i);
-            bsg_kscrw_i_writeBinaryImage(writer, NULL, info);
+            bsg_kscrw_i_writeBinaryImage(writer, NULL, iImg);
         }
     }
     writer->endContainer(writer);
--- a/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
+++ b/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
@@ -1115,7 +1115,7 @@ void bsg_kscrw_i_writeBinaryImage(const BSG_KSCrashReportWriter *const writer,
 void bsg_kscrw_i_writeBinaryImages(const BSG_KSCrashReportWriter *const writer,
                                    const char *const key)
 {
-    const uint32_t imageCount = (uint32_t)bsg_dyld_image_count();
+    const uint32_t imageCount = bsg_dyld_image_count();
 
     writer->beginArray(writer, key);
     {
--- a/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
+++ b/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
@@ -1093,13 +1093,13 @@ void bsg_kscrw_i_writeBinaryImage(const BSG_KSCrashReportWriter *const writer,
 {
     writer->beginObject(writer, key);
     {
-        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageAddress, (uintptr_t)info->mh);
+        writer->addUIntegerElement(writer, BSG_KSCrashField_ImageAddress, (uintptr_t)info->header);
         writer->addUIntegerElement(writer, BSG_KSCrashField_ImageVmAddress,          info->imageVmAddr);
         writer->addUIntegerElement(writer, BSG_KSCrashField_ImageSize,               info->imageSize);
         writer->addStringElement(writer, BSG_KSCrashField_Name,                      info->name);
         writer->addUUIDElement(writer, BSG_KSCrashField_UUID,                        info->uuid);
-        writer->addIntegerElement(writer, BSG_KSCrashField_CPUType,                  info->cputype);
-        writer->addIntegerElement(writer, BSG_KSCrashField_CPUSubType,               info->cpusubtype);
+        writer->addIntegerElement(writer, BSG_KSCrashField_CPUType,                  info->header->cputype);
+        writer->addIntegerElement(writer, BSG_KSCrashField_CPUSubType,               info->header->cpusubtype);
     }
     writer->endContainer(writer);
 }
--- a/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
+++ b/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
@@ -1113,14 +1113,12 @@ void bsg_kscrw_i_writeBinaryImage(const BSG_KSCrashReportWriter *const writer,
 void bsg_kscrw_i_writeBinaryImages(const BSG_KSCrashReportWriter *const writer,
                                    const char *const key)
 {
-    // Get the array of loaded binary images and a count
-    size_t imageCount;
-    BSG_Mach_Binary_Image_Info *info = bsg_mach_header_array(&imageCount);
-
     writer->beginArray(writer, key);
     {
-        for (uint32_t iImg = 0; iImg < imageCount; iImg++) {
-            bsg_kscrw_i_writeBinaryImage(writer, NULL, &info[iImg]);
+        BSG_Mach_Binary_Images *images = bsg_get_mach_binary_images();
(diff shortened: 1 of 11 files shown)


Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,182 @@
...
+  free(array->contents);
+  array->contents = NULL;
+  array->used = array->size = 0;
+}
+
+void bsg_initialise_mach_binary_headers() {
+    bsg_initialize_binary_images_array(&bsg_mach_binary_images, 100);
+}
+
+uintptr_t bsg_ksdlfirstCmdAfterHeader(const struct mach_header *const header);
```
tomlongridge, line 80: Include `BSG_KSDynamicLinker.h` instead?

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.h
```diff
@@ -0,0 +1,64 @@
...
+
+static BSG_Mach_Binary_Images bsg_mach_binary_images;
+
+static os_unfair_lock bsg_mach_binary_images_access_lock = OS_UNFAIR_LOCK_INIT;
+
+void bsg_initialize_binary_images_array(BSG_Mach_Binary_Images *array, size_t initialSize);
+void bsg_add_mach_binary_image(BSG_Mach_Binary_Images *array, BSG_Mach_Binary_Image_Info element);
+void bsg_remove_mach_binary_image(BSG_Mach_Binary_Images *array, const char *element_name);
+void bsg_free_binary_images_array(BSG_Mach_Binary_Images *array);
+BSG_Mach_Binary_Images *bsg_get_mach_binary_images(void);
```
tomlongridge, line 47: These are internal - should they be declared in the `.c` file?

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,182 @@
...
+//  Created by Robin Macharg on 04/05/2020.
+//  Copyright © 2020 Bugsnag. All rights reserved.
+//
+
+#import <mach-o/dyld.h>
+#import <dlfcn.h>
+#import <Foundation/Foundation.h>
+#import "BSG_KSMachHeaders.h"
+
+void bsg_initialize_binary_images_array(BSG_Mach_Binary_Images *array, size_t initialSize) {
```
tomlongridge, line 14: Do we need to pass the array into these functions? Can they not access the static directly?

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,182 @@
...
+ * position.
+ */
+void bsg_remove_mach_binary_image(BSG_Mach_Binary_Images *array, const char *element_name) {
+    
+    os_unfair_lock_lock(&bsg_mach_binary_images_access_lock);
+    
+    for (size_t i=0; i<array->used; i++) {
+        BSG_Mach_Binary_Image_Info item = array->contents[i];
+        if (strcmp(element_name, item.name) == 0) {
+            if (array->used >= 2) {
```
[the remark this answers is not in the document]
robinmacharg: Handled correctly, albeit with a redundant copy from last -> last, at the same time as decrementing the 'used' counter.  Comment added for clarity, along with test cases to ensure this behaviour is as expected.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.h
robinmacharg: Yes, excepting `bsg_get_mach_binary_images()`

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
robinmacharg: Good spot.  Initially written to be a more general-purpose container, the storage is file-static so yes, parameter removed.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.h
```diff
@@ -0,0 +1,66 @@
+    //
```
tomlongridge, line 1: Stray tab

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,180 @@
...
+            break; // an image can only be loaded singularly; exit loop once found
+        }
+    }
+    
+    os_unfair_lock_unlock(&bsg_mach_binary_images_access_lock);
+}
+
+void bsg_initialise_mach_binary_headers() {
+    bsg_initialize_binary_images_array(BSG_INITIAL_MACH_BINARY_IMAGE_ARRAY_SIZE);
+}
```
tomlongridge, line 78: Nit: do we still need this extra function?

robinmacharg: This facilitated easier testing but, if you're happy for the default array size value to be delocalised slightly and live in `BSG_KSCrash.m`, then we can do away with the extra level.

tomlongridge: No strong view - your call.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+/**
+ * Called when a binary image is unloaded.
+ */
+void bsg_mach_binary_image_removed(const struct mach_header *header, intptr_t slide)
+{
+    // Convert header and slide into an info struct
+    BSG_Mach_Binary_Image_Info info;
```
fractalwrench, line 169: Consider initializing the struct in the same manner as `bsg_mach_binary_image_added` for consistency

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,173 @@
...
+ * This should be called in a threadsafe way; we lock against a simultaneous add and remove.
+ */
+void bsg_add_mach_binary_image(BSG_Mach_Binary_Image_Info element) {
+    
+    os_unfair_lock_lock(&bsg_mach_binary_images_access_lock);
+    
+    // Expand array if necessary
+    if (bsg_mach_binary_images.used == bsg_mach_binary_images.size) {
+        bsg_mach_binary_images.size *= 2;
+        bsg_mach_binary_images.contents = (BSG_Mach_Binary_Image_Info *)realloc(bsg_mach_binary_images.contents, bsg_mach_binary_images.size * sizeof(BSG_Mach_Binary_Image_Info));
```
fractalwrench, line 31: style nit: consider reducing the length of this line

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,173 @@
...
+ * The array doubles on filling-up.  Typical sizes is expected to be in < 1000 (i.e. 2-3 doublings, at app start-up)
+ * This should be called in a threadsafe way; we lock against a simultaneous add and remove.
+ */
+void bsg_add_mach_binary_image(BSG_Mach_Binary_Image_Info element) {
+    
+    os_unfair_lock_lock(&bsg_mach_binary_images_access_lock);
+    
+    // Expand array if necessary
+    if (bsg_mach_binary_images.used == bsg_mach_binary_images.size) {
+        bsg_mach_binary_images.size *= 2;
```
fractalwrench, line 30: strictly speaking this is not always true - `realloc` could fail to find enough memory to allocate the entire requested value. The paranoid way of doing things would be to query the size using the returned pointer and store that as the value.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,173 @@
...
+ * position.
+ */
+void bsg_remove_mach_binary_image(const char *element_name) {
+    
+    os_unfair_lock_lock(&bsg_mach_binary_images_access_lock);
+    
+    for (size_t i=0; i<bsg_mach_binary_images.used; i++) {
+        BSG_Mach_Binary_Image_Info item = bsg_mach_binary_images.contents[i];
+        
+        if (strcmp(element_name, item.name) == 0) {
```
fractalwrench, line 52: Will name definitely always be unique for binary images? `vmAddress` seems like it would definitely always be unique so it might be good to use that here. Perhaps being overly paranoid - if the filename is guaranteed to be different then this would be fine.

iOS/Bugsnag.xcodeproj/project.pbxproj
```diff
@@ -883,6 +891,8 @@
 				E7107C011F4C97F100BB3F98 /* BSG_KSMach_x86_32.c */,
 				E7107C021F4C97F100BB3F98 /* BSG_KSMach_x86_64.c */,
 				E7107C031F4C97F100BB3F98 /* BSG_KSMachApple.h */,
+				0071CB4A2460213200F562B1 /* BSG_KSMachHeaders.m */,
+				0071CB4D2460216200F562B1 /* BSG_KSMachHeaders.h */,
```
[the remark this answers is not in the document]
robinmacharg: Added a note to the PR template as a reminder.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,173 @@
...
+#import "BSG_KSDynamicLinker.h"
+#import "BSG_KSMachHeaders.h"
+
+BSG_Mach_Binary_Images *bsg_get_mach_binary_images() {
+    return &bsg_mach_binary_images;
+}
+
+/**
+ * Store a Mach binary image-excapsulating struct in a dynamic array.
+ * The array doubles on filling-up.  Typical sizes is expected to be in < 1000 (i.e. 2-3 doublings, at app start-up)
```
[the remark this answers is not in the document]
robinmacharg: Updated to 400.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
robinmacharg: I think filename _should_ be unique, but can imagine a pathological scenario where libraries had filenames changed prior to loading (and that's an interesting experiment to try); I've changed the code to use vmAddr as suggested since this is internal and once loaded this should be unique.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
robinmacharg: While an OOM (at the small scales we're dealing with here) is indicative of bigger problems I agree with your paranoia.  I've opted for a simple approach whereby a failing `realloc` simply doesn't store the image info or adjust the array size.  I don't try smaller resizings either; better, I think, to lose the image info than introduce complex, potentially error prone, reallocation heuristics.  Happy to dig further if you disagree.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,173 @@
...
+ * This should be called in a threadsafe way; we lock against a simultaneous add and remove.
+ */
+void bsg_add_mach_binary_image(BSG_Mach_Binary_Image_Info element) {
+    
+    os_unfair_lock_lock(&bsg_mach_binary_images_access_lock);
```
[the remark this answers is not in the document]
robinmacharg: I've added some preprocessor OS version guards and `lock()`/`unlock()` wrappers around the appropriate sections.  The (now deprecated) `OSSpinLock` mechanism is used on older OSs.  Chosen over `pthreads_mutex` for simplicity of implementation which should still be sufficient for our needs in this section of the code.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.h
```diff
@@ -39,9 +39,16 @@ typedef struct {
...
- * A lock, used to synchronise access to the array of binary image info.
+ * An OS-version-specific lock, used to synchronise access to the array of binary image info.
  */
-static os_unfair_lock bsg_mach_binary_images_access_lock = OS_UNFAIR_LOCK_INIT;
+#if defined(__IPHONE_10_0) || defined(__MAC_10_12) || defined(__TVOS_10_0) || defined(__WATCHOS_3_0)
+    #import <os/lock.h>
+    static os_unfair_lock bsg_mach_binary_images_access_lock = OS_UNFAIR_LOCK_INIT;
+#else
+    #import <libkern/OSAtomic.h>
+    #define OS_SPINLOCK_INIT 0
```
fractalwrench, line 49: This seems to be defined in the header already, does this statement need to be included for functional reasons or can it be removed?

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -16,40 +16,73 @@
...
+    // but we should still do *our* best not to crash the app.
     if (bsg_mach_binary_images.used == bsg_mach_binary_images.size) {
         bsg_mach_binary_images.size *= 2;
-        bsg_mach_binary_images.contents = (BSG_Mach_Binary_Image_Info *)realloc(bsg_mach_binary_images.contents, bsg_mach_binary_images.size * sizeof(BSG_Mach_Binary_Image_Info));
+        size_t newSize = bsg_mach_binary_images.size * sizeof(BSG_Mach_Binary_Image_Info);
+        void **newData = realloc(bsg_mach_binary_images.contents, newSize);
+        
+        // Exit early if we can't allocate memory
+        if (newData == NULL) {
+            return;
```
fractalwrench, line 61: This could lead to deadlock as the lock taken out earlier hasn't been released. `realloc()` also doesn't return `NULL` according to the header documentation:

>The realloc() function tries to change the size of the allocation pointed
     to by ptr to size, and returns ptr.  If there is not enough room to
     enlarge the memory allocation pointed to by ptr, realloc() creates a new
     allocation, copies as much of the old data pointed to by ptr as will fit
     to the new allocation, frees the old allocation, and returns a pointer to
     the allocated memory.  If ptr is NULL, realloc() is identical to a call
     to malloc() for size bytes.  If size is zero and ptr is not NULL, a new,
     minimum sized object is allocated and the original object is freed.  When
     extending a region allocated with calloc(3), realloc(3) does not guaran-
     tee that the additional memory is also zero-filled.

For maximum safety I think it'd be best to always assign the return value of `realloc()` to `bsg_mach_binary_images.contents` and avoid returning early.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.h
robinmacharg: Hah.   Missed due to the `#if` blocking symbol resolution.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -0,0 +1,206 @@
...
+#endif
+}
+
+void unlock() {
+#if defined(__IPHONE_10_0) || defined(__MAC_10_12) || defined(__TVOS_10_0) || defined(__WATCHOS_3_0)
+    os_unfair_lock_unlock(&bsg_mach_binary_images_access_lock);
+#else
+    OSSpinLockUnlock(&bsg_mach_binary_images_access_lock);
+#endif
+}
```
tomlongridge, line 41: Method names need to be expanded with `bsg_` prefix and probably more description.

Is it worth making a define for the method name to use for lock and unlock in the header and not having separate functions and repeated conditionals in this file?

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
robinmacharg: Thanks for encouraging me to dig a little deeper.  👍 TIL.  The RETURN VALUES section of [the man page](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/realloc.3.html) says:

> RETURN VALUES
>      If successful, calloc(), malloc(), realloc(), reallocf(), valloc(), and aligned_alloc() functions return a pointer to allocated memory.  If there is an error, they return a NULL pointer and set errno to ENOMEM.
> ...
> For realloc(), the input pointer is still valid if reallocation failed....

The [realloc implementation](https://opensource.apple.com/source/Libc/Libc-594.1.4/gen/malloc.c) concurs.

I've added an additional check for success and exit early if the failure conditions outlined above are met.  This way temporary memory starvation only fails to record the mach header info but will keep trying.

Does this approach seem safe enough?

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
robinmacharg: Thanks for the comment - that does indeed lead to a neater implementation.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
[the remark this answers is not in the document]
robinmacharg: w.r.t. a later comment this has been reimplemented as `#if` version-guarded `#define`d method names.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1084,70 +1085,21 @@ int bsg_kscrw_i_threadIndex(const thread_t thread) {
...
-            struct uuid_command *uuidCmd = (struct uuid_command *)cmdPtr;
-            uuid = uuidCmd->uuid;
-            break;
-        }
-        }
-        cmdPtr += loadCmd->cmdsize;
-    }
-
+                                  const char *const key,
+                                  const BSG_Mach_Binary_Image_Info *info)
```
tomlongridge, line 1092: Just a small nit, but it might be more consistent to revert back to the original signature of this method and obtain the image info from within `bsg_kscrw_i_writeBinaryImage` (everywhere else we pass around an index).

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.h
```diff
@@ -0,0 +1,103 @@
...
+    intptr_t slide;
+} BSG_Mach_Binary_Image_Info;
+
+/**
+ * MARK: - A Dynamic array container
+ * See: https://stackoverflow.com/a/3536261/2431627
+ */
+typedef struct {
+    size_t used;
+    size_t size;
```
tomlongridge, line 35: Consider making these `uint32_t` to better match dyld and avoid a cast in `bsg_kscrw_i_writeBinaryImages`?

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSDynamicLinker.c
```diff
@@ -115,16 +116,20 @@ uint32_t bsg_ksdlimageIndexContainingAddress(const uintptr_t address) {
                 if (loadCmd->cmd == LC_SEGMENT) {
                     const struct segment_command *segCmd =
                         (struct segment_command *)cmdPtr;
-                    if (addressWSlide >= segCmd->vmaddr &&
-                        addressWSlide < segCmd->vmaddr + segCmd->vmsize) {
-                        return iImg;
+                    if ( strcmp(segCmd->segname, "__TEXT") == 0 ) {
```
tomlongridge, line 119: Are you sure this is needed? The check for "__TEXT" is happening when the data is populated to the image cache, so I wouldn't have thought it necessary here, so we should reduce the changeset.

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.h
```diff
@@ -0,0 +1,103 @@
...
+ * detailing a crash report's binary images.
+ */
+typedef struct {
+    const struct mach_header *mh;      /* The mach_header - 32 or 64 bit */
+    uint64_t imageVmAddr;
+    uint64_t imageSize;
+    uint8_t *uuid;
+    const char* name;
+    cpu_type_t cputype;          /* cpu specifier */
+    cpu_subtype_t cpusubtype;    /* machine specifier */
```
tomlongridge, line 25: Do we need to denormalise cpu type/subtype? Can't they just be accessed via the header?

Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.h
```diff
@@ -0,0 +1,103 @@
...
+#define BSG_KSMachHeaders_h
+
+#import <mach/machine.h>
+
+/**
+ * An encapsulation of the Mach header - either 64 or 32 bit, along with some additional information required for
+ * detailing a crash report's binary images.
+ */
+typedef struct {
+    const struct mach_header *mh;      /* The mach_header - 32 or 64 bit */
```
tomlongridge, line 19:
```suggestion
    const struct mach_header *header;      /* The mach_header - 32 or 64 bit */
```

PR #582, opened 2020-05-04
Load configuration from Plist files

fractalwrench:
## Goal

Allows loading of Configuration from Plist files, without creating a configuration object up front in code.

## Changeset

- Added `BugsnagConfiguration.loadConfig()` method, which reads from the main bundle's plist. Note that the naming is a required deviation from the spec due to an [existing method](https://developer.apple.com/documentation/objectivec/nsobject/1418815-load?language=objc) on `NSObject`
- Added non-public class `BSGConfigurationBuilder` which loads configuration options from the plist under the 'bugsnag' key
- Updated `BugsnagConfiguration` to set the `apiKey` property to an empty string if it is not a valid value or a string value. Previous behaviour was to keep it as the passed in value, even if the object was of a different type (such as a dictionary)

Reads the following values:

```
apiKey
appType
appVersion
autoDetectErrors
autoTrackSessions
bundleVersion
persistUser
releaseStage
enabledReleaseStages
endpoints
maxBreadcrumbs
redactedKeys
sendThreads
```

## Tests

- Added unit tests to verify the default and overridden values of configuration options loaded from a plist
- Added a mazerunner scenario which loads configuration from a plist

### DIFF
--- a/OSX/Info.plist
+++ b/OSX/Info.plist
@@ -19,7 +19,7 @@
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<string>1</string>
 	<key>NSHumanReadableCopyright</key>
 	<string>Copyright © 2016 Bugsnag. All rights reserved.</string>
 	<key>NSPrincipalClass</key>
--- a/OSX/Info.plist
+++ b/OSX/Info.plist
@@ -21,7 +21,7 @@
 	<key>CFBundleVersion</key>
 	<string>1</string>
 	<key>NSHumanReadableCopyright</key>
-	<string>Copyright © 2016 Bugsnag. All rights reserved.</string>
+	<string>Copyright © 2020 Bugsnag. All rights reserved.</string>
 	<key>NSPrincipalClass</key>
 	<string></string>
 </dict>
--- a/Source/BSGConfigurationBuilder.h
+++ b/Source/BSGConfigurationBuilder.h
@@ -1,0 +1,16 @@
+#import <Foundation/Foundation.h>
+
+@class BugsnagConfiguration;
+
+@interface BSGConfigurationBuilder : NSObject
+
+/**
+ * Creates a configuration, applying supported options before returning the new
+ * config object. Unsupported options warn.
+ *
+ * @return a new BugsnagConfiguration object or nil if the a valid object could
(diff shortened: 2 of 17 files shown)


Tests/BugsnagConfigurationTests.m
```diff
@@ -3,8 +3,9 @@
  */
 
 #import <XCTest/XCTest.h>
-#import "BugsnagTestConstants.h"
+#import <objc/runtime.h>
```
robinmacharg, line 6: nit: I don't believe runtime.h is required?

fractalwrench: It's not and this is a relic from an earlier approach - I've removed the import and rebased to resolve the previous conflicts.

OSX/Info.plist
```diff
@@ -19,7 +19,7 @@
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<string>1</string>
```
robinmacharg, line 22: Should this (and other occurrences) follow the library versioning, i.e. 6.0 for this)?  And if so should we update the release process to automatically apply that?

OSX/Info.plist
```diff
@@ -19,7 +19,7 @@
...
 	<key>NSHumanReadableCopyright</key>
 	<string>Copyright © 2016 Bugsnag. All rights reserved.</string>
```
robinmacharg, line 24: nit: while I'm looking at this, the copyrights could be updated.

Source/BSGConfigurationBuilder.h
```diff
@@ -0,0 +1,16 @@
...
+
+@class BugsnagConfiguration;
+
+@interface BSGConfigurationBuilder : NSObject
+
+/**
+ * Creates a configuration, applying supported options before returning the new
+ * config object. Unsupported options warn.
+ *
+ * @return a new BugsnagConfiguration object or nil if the a valid object could
```
robinmacharg, line 11: nit: I don't see nil being returned or warnings emitted.

fractalwrench: I've removed this part of the doc as that behaviour would be inconsistent with Android/the spec, and doesn't reflect reality.

OSX/Info.plist
fractalwrench: Updated copyright

PR #583, opened 2020-05-06
Allow construction of app/device objects from json

fractalwrench:
## Goal

Adds non-public methods for constructing `BugsnagAppWithState` and `BugsnagDeviceWithState` from JSON. These will be used by React Native when reading JS payloads.

### DIFF
--- a/Source/BugsnagAppWithState.m
+++ b/Source/BugsnagAppWithState.m
@@ -22,6 +22,34 @@
 
 @implementation BugsnagAppWithState
 
++ (BugsnagAppWithState *)appFromJson:(NSDictionary *)json {
+    BugsnagAppWithState *app = [BugsnagAppWithState new];
+
+    id duration = json[@"duration"];
+    if (duration && [duration isKindOfClass:[NSNumber class]]) {
+        app.duration = [(NSNumber *) duration unsignedIntValue];
+    }
+
+    id durationInForeground = json[@"durationInForeground"];
+    if (durationInForeground && [durationInForeground isKindOfClass:[NSNumber class]]) {
+        app.durationInForeground = [(NSNumber *) durationInForeground unsignedIntValue];
+    }
+
+    id inForeground = json[@"inForeground"];
+    if (inForeground) {
+        app.inForeground = [(NSNumber *) inForeground boolValue];
+    }
+
+    app.bundleVersion = json[@"bundleVersion"];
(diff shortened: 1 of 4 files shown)


Source/BugsnagDeviceWithState.m
```diff
@@ -75,6 +75,33 @@ - (instancetype)init {
...
+    device.freeMemory = json[@"freeMemory"];
+    device.freeDisk = json[@"freeDisk"];
+    device.locale = json[@"locale"];
+    device.manufacturer = json[@"manufacturer"];
+    device.model = json[@"model"];
+    device.modelNumber = json[@"modelNumber"];
+    device.orientation = json[@"orientation"];
+    device.osName = json[@"osName"];
+    device.osVersion = json[@"osVersion"];
+    device.runtimeVersions = json [@"runtimeVersions"];
```
tomlongridge, line 90:
```suggestion
    device.runtimeVersions = json[@"runtimeVersions"];
```

Source/BugsnagDeviceWithState.m
```diff
@@ -75,6 +75,33 @@ - (instancetype)init {
...
+    device.totalMemory = json[@"totalMemory"];
+
+    id jailbroken = json[@"jailbroken"];
+    if (jailbroken) {
+        device.jailbroken = [(NSNumber *) jailbroken boolValue];
```
tomlongridge, line 95: Jailbroken is not listed as nullable - do we need a default?

Source/BugsnagAppWithState.m
```diff
@@ -22,6 +22,34 @@ - (NSDictionary *)toDict;
...
+
+    id durationInForeground = json[@"durationInForeground"];
+    if (durationInForeground && [durationInForeground isKindOfClass:[NSNumber class]]) {
+        app.durationInForeground = [(NSNumber *) durationInForeground unsignedIntValue];
+    }
+
+    id inForeground = json[@"inForeground"];
+    if (inForeground) {
+        app.inForeground = [(NSNumber *) inForeground boolValue];
+    }
```
tomlongridge, line 41: These properties aren't nullable - do we need a default?

Source/BugsnagDeviceWithState.m
fractalwrench: The default will be false for both the `jailbroken` and `inForeground` properties as that's the default value for the `BOOL` type in Objective-C.

PR #585, opened 2020-05-11
Persist bugsnag event information for non-fatal errors

fractalwrench:
## Goal

`OnError` callbacks run on handled errors, meaning the user can alter property values on the `BugsnagEvent`. These property values need to be persisted to disk and deserialized when the `BugsnagEvent` is read from a cached report again.

This changeset makes the notifier persist this information when it is written to disk, and ensures that `BugsnagEvent` is capable of deserializing the recorded JSON information.

**Note**: the payload information is still not populated when passed to the `OnError` callback - this will be addressed in separate work.

## Changeset

- Whenever `notifyInternal` is invoked on `BugsnagClient`, generate a payload matching the event object in the [Error Reporting API schema](https://bugsnagerrorreportingapi.docs.apiary.io/#reference/0/notify/send-error-reports) and set it as an override property on the `BugsnagEvent`
- The [CrashSentry](https://github.com/bugsnag/bugsnag-cocoa/compare/v6...v6-event-persistence?expand=1#diff-e931bae5f812bd60eb62723280009513R946) captures the user reported error and generates a KSCrashReport. This KSCrashReport contains a `user` section where it is possible to add arbitrary JSON. The payload information is therefore serialized under the key `user.overrides.bugsnag_event`
- The `BugsnagEvent` constructor `initWithKSReport` deserializes the persisted KSCrashReport. If the payload information is contained within `user.overridesbugsnag_event`, this information is parsed and overrides any default values set by KSCrash
- Removed a redundant accessor for `context` which uses the override property to ensure that "context" is persisted as an individual field. This accessor is no longer required and has the additional benefit that calling `setContext` no longer waits for a mutex

## Tests
- Added unit tests to verify that the `BugsnagEvent` can read a payload
- Inspected JSON payload for a handled error to verify the overrides are serialized
- Verified that an error sent to the dashboard displays the correct context


Source/BugsnagClient.m
```diff
@@ -931,6 +933,11 @@ - (void)notifyInternal:(BugsnagEvent *_Nonnull)event
         exc = event.originalError;
     }
 
+    // handled errors should persist any information edited by the user
+    // in a section within the KSCrash report so it can be read
+    // when the error is delivered
+    [event setOverrideProperty:@"bugsnag_event" value:[event toJson]];
```
tomlongridge, line 939: I think I'd prefer to see this data being kept separately to `user.overrides` to avoid blurring this relationship as it's potentially confusing - how about just `user.event`?

fractalwrench:
Updated to persist information using the `user.event` key instead. As this required touching the internals of KSCrash it'd be good to get a quick re-review. Tested by setting the context on a handled error and verifying the value reported to the dashboard was accurate.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.h
```diff
@@ -137,7 +137,8 @@ typedef enum {
                    metadata:(NSDictionary *)metadata
                      config:(NSDictionary *)config
                discardDepth:(int)depth
-           terminateProgram:(BOOL)terminateProgram;
+           terminateProgram:(BOOL)terminateProgram
+                  eventJson:(NSDictionary *)eventJson;
```
tomlongridge, line 141: The "Json" suffix seems a little inaccurate here - can we call the parameter `eventOverrides` and move it next to `callbackOverrides`?

Also - the doc-comments above need updating.

fractalwrench:
Renamed and changed order of parameters.

PR #586, opened 2020-05-11
Fix warnings with -Wpedantic

keith:
## Goal

This fixes a few small warnings that popped up when using super strict
warning flags.

image

## Design

Internally we build Bugsnag with Bazel, to fix these externally I added `-Wpedantic -Wall -Wextra` to `OTHER_CFLAGS` for testing (I did not push that).

## Tests

This shouldn't introduce any behavioral change

### Outstanding Questions

  - [x] Final review
  - [ ] Release


Source/BugsnagCrashReport.m
```diff
@@ -671,12 +671,9 @@ + (instancetype)errorDataFromThreads:(NSArray *)threads {
             }
             NSString *contentValue = data[@"value"];
 
-#pragma clang diagnostic push
-#pragma ide diagnostic ignored "OCDFAInspection"
```
keith, line 675: I couldn't find anything online on what this actually did, but the compiler warning implied that it was ignored

keith:
If you're interested being compatible with these super strict warnings in general I'm happy to push the bit I was testing with which was something like this:

```
+OTHER_CFLAGS = $(inherited) -Wpedantic -Wall -Wextra -Wno-gnu-zero-variadic-macro-arguments -Wno-gnu-conditional-omitted-operand -Wno-unused-parameter -Wno-missing-field-initializers -Wno-gnu-statement-expression -Wno-zero-length-array
```

keith:
🤔 the test failure here appears to be unrelated

mattdyoung:
Hi @keith 

Thanks for the PR! We'll review and consider this in due course.

tomlongridge:
This is now released under v6.0.1.

PR #589, opened 2020-05-12
Deserialize metadata/crumbs/severityReason in BugsnagEvent

fractalwrench:
## Goal

Deserializes payload information which is persisted as part of #585. This allows information added to callbacks for handled errors to be persisted after the error is written to disk.

## Changeset

Deserializes handled error information for the following fields:

- metadata
- breadcrumbs
- severityReason

### DIFF
--- a/Source/BugsnagEvent.m
+++ b/Source/BugsnagEvent.m
@@ -48,6 +48,7 @@ NSDictionary *_Nonnull BSGParseDeviceMetadata(NSDictionary *_Nonnull event);
 + (instancetype _Nullable)breadcrumbWithBlock:
         (BSGBreadcrumbConfiguration _Nonnull)block;
 + (instancetype _Nullable)breadcrumbFromDict:(NSDictionary *_Nonnull)dict;
++ (NSArray<BugsnagBreadcrumb *> *)breadcrumbArrayFromJson:(NSArray *)json;
 @end
 
 @interface BugsnagUser ()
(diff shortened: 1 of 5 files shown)


Source/BugsnagEvent.m
```diff
@@ -430,6 +430,37 @@ - (void)parseHandledCrashData:(NSDictionary *)event {
...
+    NSDictionary *attrs = data[@"attributes"];
+
+    if (attrs != nil && [attrs count] == 1) { // only 1 attrValue is ever present
+        attrValue = [attrs allValues][0];
+    }
+    SeverityReasonType reason = [BugsnagHandledState severityReasonFromString:data[@"type"]];
+    _handledState = [[BugsnagHandledState alloc] initWithSeverityReason:reason
+                                                               severity:severity
+                                                              unhandled:unhandled
+                                                              attrValue:attrValue];
```
tomlongridge, line 463: Can you abstract this to a `handledStateFromJson` (or similar) in `BugsnagHandledState` as you have done for app/device/error? Possibly also breadcrumbs?

fractalwrench: Factored out methods for both breadcrumbs + handled state.

Source/BugsnagHandledState.m
```diff
@@ -47,6 +47,24 @@ BSGSeverity BSGParseSeverity(NSString *severity) {
...
+    NSDictionary *attrs = data[@"attributes"];
+
+    if (attrs != nil && [attrs count] == 1) { // only 1 attrValue is ever present
+        attrValue = [attrs allValues][0];
+    }
+    SeverityReasonType reason = [BugsnagHandledState severityReasonFromString:data[@"type"]];
+    return [[BugsnagHandledState alloc] initWithSeverityReason:reason
+                                                               severity:severity
+                                                              unhandled:unhandled
+                                                              attrValue:attrValue];
```
tomlongridge, line 65: Alignment glitch

PR #591, opened 2020-05-13
Deserialize Thread info stored in BugsnagEvent

fractalwrench:
## Goal

Deserializes thread payload information which is persisted as part of #585. This allows information added to callbacks for handled errors to be persisted after the error is written to disk.

## Changeset

Deserializes handled error information. This has required the addition of deserializer functions for threads, which are covered by a new unit test.

### DIFF
--- a/Source/BugsnagThread.m
+++ b/Source/BugsnagThread.m
@@ -20,10 +20,42 @@
 @end
 
 @interface BugsnagStacktrace ()
++ (instancetype)stacktraceFromJson:(NSDictionary *)json;
 @property NSMutableArray<BugsnagStackframe *> *trace;
 @end
 
 @implementation BugsnagThread
+
++ (instancetype)threadFromJson:(NSDictionary *)json {
+    if (json == nil) {
+        return nil;
+    }
+    NSString *type = json[@"type"];
+    BSGThreadType threadType = [@"cocoa" isEqualToString:type] ? BSGThreadTypeCocoa : BSGThreadTypeReactNativeJs;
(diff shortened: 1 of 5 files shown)


Source/BugsnagThread.m
```diff
@@ -20,11 +20,43 @@ @interface BugsnagThread ()
...
 @end
 
 @implementation BugsnagThread
 
++ (instancetype)threadFromJson:(NSDictionary *)json {
+    if (json == nil) {
+        return nil;
+    }
+    NSString *type = json[@"type"];
+    BSGThreadType threadType = [@"cocoa" isEqualToString:type] ? BSGThreadTypeCocoa : BSGThreadTypeReactNativeJs;
```
tomlongridge, line 34: Although relatively simple, it would be good to see the serialisation and parsing of thread type factored out into separate functions as per `BSGParseErrorType` on #588 to avoid repetition of the logic buried down this far.

fractalwrench: Factored out to separate functions.

Source/BugsnagThread.m
```diff
@@ -11,6 +11,14 @@
 #import "BugsnagStacktrace.h"
 #import "BugsnagKeys.h"
 
+BSGThreadType BSGDeserializeThreadType(NSString *type) {
```
tomlongridge, line 14: We seem to have more of a precedent for calling this parsing:

```suggestion
BSGThreadType BSGParseThreadType(NSString *type) {
```

PR #594, opened 2020-05-13
Populate BugsnagEvent with initial data

fractalwrench:
## Goal

Populates `BugsnagEvent` objects generated in `notifyInternal` with initial data about the application state. This allows users to read/write this information in `OnError` callbacks for handled errors, whose information is now persisted thanks to changes made in #585 

## Changeset

The following fields are now populated:

- context
- apiKey
- app
- breadcrumbs
- device
- metadata
- user

This changeset leaves the `errors.stacktrace` and `threads` properties unpopulated as these are currently harder to populate due to needing to parse binary images. These will be addressed in a future PR.

## Tests

Verified that the event is populated by inspecting the contents of `BugsnagEvent` in an `OnError` callback.

### DIFF
--- a/Source/BugsnagEvent.m
+++ b/Source/BugsnagEvent.m
@@ -435,7 +435,7 @@ NSDictionary *BSGParseCustomException(NSDictionary *report,
     if (bugsnagPayload == nil || ![bugsnagPayload isKindOfClass:[NSDictionary class]]) {
         return;
     }
-    self.apiKey = bugsnagPayload[@"apiKey"];
+    _apiKey = bugsnagPayload[@"apiKey"];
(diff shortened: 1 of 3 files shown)


Source/BugsnagEvent.m
```diff
@@ -435,7 +435,7 @@ - (void)parseHandledCrashData:(NSDictionary *)event {
     if (bugsnagPayload == nil || ![bugsnagPayload isKindOfClass:[NSDictionary class]]) {
         return;
     }
-
+    self.apiKey = bugsnagPayload[@"apiKey"];
```
tomlongridge, line 438:
```suggestion
    _apiKey = bugsnagPayload[@"apiKey"];
```

PR #598, opened 2020-05-13
Remove cpp extension from podspec

keith:
There are no `*.cpp` in this project.


mattdyoung:
Thanks @keith! We'll look at reviewing and merging.

robinmacharg:
Thanks for spotting this, @keith; we appreciate your attention to detail.  I've rolled the change into our work on v6 of the notifier, linked to in #633, above.

PR #600, opened 2020-05-14
Reinstate app.name field

robinmacharg:
## Goal

The `app.name` field was removed during recent work to create structured objects.  This change reinstates it.  the value is sourced from `CFBundleExecutable`.

## Design

A simple search for all `app.*`-referencing locations (e.g. `app.id`) and insertion of suitable code.

## Tests

Non-nil-checking unit tests were already present.  Manual testing that the app name is sent through to the appropriate tab on the dashboard were performed.

### Outstanding Questions

  - [X] Final review
  - [ ] Release

  features)


Source/BugsnagEvent.m
```diff
@@ -353,12 +361,15 @@ - (instancetype)initWithKSReport:(NSDictionary *)event {
                 self.metadata = [BugsnagMetadata new];
             }
 
-            NSDictionary *deviceMetadata = BSGParseDeviceMetadata(event);
-            [self.metadata addMetadata:deviceMetadata toSection:@"device"];
-
+            // Cocoa-specific, non-spec., device and app data
+            [self.metadata addMetadata:BSGParseDeviceMetadata(event) toSection:@"device"];
```
tomlongridge, line 365: Should we add the device metadata to OOMs too?

robinmacharg: Out of scope, but no reason not to.  There's also a case for adding `charging` and `battery` to the sentinel file as well; they look to be missing.  Out of scope for this task.

PR #601, opened 2020-05-14
Allow capturing threads without writing to disk

fractalwrench:
## Goal

Captures the thread information from KSCrash without writing to disk. This is required for handled reports, where it's necessary to display the thread information up front so a user can invoke a callback.

This changeset fully populates the `event.threads` and `event.errors` fields before writing to disk.

### Thread capture
- Added a method to `BSG_KSCrash` for deserializing JSON from a char array, and for constructing threads out of a JSON dictionary
- Added `bsg_kscrash_captureThreadTrace()`, which is an internal API for suspending threads and captures their information as a JSON string
- Added convenience methods for suspending/resuming threads
- `BSG_KSCrashReport` has its JSON serializers for thread/stacktrace information to separate methods, which are invoked both when collecting thread information and when writing a report to disk
- Different `BSG_KSCrashReportWriter` implementations are passed when collecting thread information - one writes JSON to a char array, the other writes JSON to disk
- The JSON codec does not requires a null terminator when writing to disk, this is added when copying to the char array by using a temporary variable

### Stacktrace collection
- Updated default discard depth to 3 and made all `notify` calls within `BugsnagClient` have a consistent count of stackframes by inlining implementations where appropriate
- Removed unused parameters from `reportUserException` on `BugsnagCrashSentry`
- Fill out context fields in `bsg_kscrash_captureThreadTrace` which are required for stacktrace serialization
- Extracted stacktrace collection from `bsg_kscrashsentry_reportUserException` and `BugsnagClient` into `captureThreads`

## Tests

Added a unit test for JSON deserialization and verified that the thread information is displayed in the dashboard for an example app.

## TBD

- address E2E test failures (e.g. swift fatalError)
- remove unused parameters?

### DIFF
--- a/Source/BugsnagClient.m
+++ b/Source/BugsnagClient.m
@@ -63,7 +63,7 @@ NSString *const BSTabCrash = @"crash";
 NSString *const BSAttributeDepth = @"depth";
 NSString *const BSEventLowMemoryWarning = @"lowMemoryWarning";
 
-static NSInteger const BSGNotifierStackFrameCount = 3;
+static NSInteger const BSGNotifierStackFrameCount = 4;
 
 struct bugsnag_data_t {
     // Contains the state of the event (handled/unhandled)
@@ -832,8 +832,19 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
 
 // MARK: - Notify
 
+// note - some duplication between notifyError calls is required to ensure
+// the same number of stackframes are used for each call.
+// see notify:handledState:block for further info
+
 - (void)notifyError:(NSError *_Nonnull)error {
-    [self notifyError:error block:nil];
+    BugsnagHandledState *state = [BugsnagHandledState handledStateWithSeverityReason:HandledError
+                                                                            severity:BSGSeverityWarning
+                                                                           attrValue:error.domain];
+    [self notify:[self createNSErrorWrapper:error]
+    handledState:state
+           block:^BOOL(BugsnagEvent *_Nonnull event) {
+               return [self appendNSErrorInfo:error block:nil event:event];
+           }];
 }
 
 - (void)notifyError:(NSError *)error
@@ -842,28 +853,37 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
     BugsnagHandledState *state = [BugsnagHandledState handledStateWithSeverityReason:HandledError
                                                                             severity:BSGSeverityWarning
                                                                            attrValue:error.domain];
-    NSException *wrapper = [NSException exceptionWithName:NSStringFromClass([error class])
-                                                   reason:error.localizedDescription
-                                                 userInfo:error.userInfo];
-    [self notify:wrapper
+    [self notify:[self createNSErrorWrapper:error]
     handledState:state
            block:^BOOL(BugsnagEvent *_Nonnull event) {
-                event.originalError = error;
-                [event addMetadata:@{
-                                        @"code" : @(error.code),
-                                        @"domain" : error.domain,
-                                        BSGKeyReason : error.localizedFailureReason ?: @""
-                                    }
-                         toSection:@"nserror"];
-               if (event.context == nil) { // set context as error domain
-                    event.context = [NSString stringWithFormat:@"%@ (%ld)", error.domain, (long)error.code];
-               }
-
-               if (block) {
-                   return block(event);
-               }
-               return true;
+               return [self appendNSErrorInfo:error block:block event:event];
            }];
+}
+
+- (NSException *)createNSErrorWrapper:(NSError *)error {
+    return [NSException exceptionWithName:NSStringFromClass([error class])
+                                   reason:error.localizedDescription
+                                 userInfo:error.userInfo];
+}
+
+- (BOOL)appendNSErrorInfo:(NSError *)error
+                    block:(BugsnagOnErrorBlock)block
+                    event:(BugsnagEvent *)event {
+    event.originalError = error;
+    [event addMetadata:@{
+                            @"code" : @(error.code),
+                            @"domain" : error.domain,
+                            BSGKeyReason : error.localizedFailureReason ?: @""
+                        }
+             toSection:@"nserror"];
+    if (event.context == nil) { // set context as error domain
+         event.context = [NSString stringWithFormat:@"%@ (%ld)", error.domain, (long)error.code];
+    }
+
+    if (block) {
+        return block(event);
+    }
+    return true;
 }
 
 - (void)notify:(NSException *_Nonnull)exception {
@@ -930,14 +950,17 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
      * This helps remove bugsnag frames from showing in NSErrors as their
      * trace is synthesized.
      *
-     * 0. +[Bugsnag notifyError:block:]
-     * 1. -[BugsnagClient notifyError:block:]
+     * For example, for [Bugsnag notifyError:block:], bugsnag adds the following
+     * frames which must be removed:
+     *
+     * 1. +[Bugsnag notifyError:block:]
      * 2. -[BugsnagClient notifyError:block:]
-     * 3. -[BSG_KSCrash captureThreads:depth:]
+     * 3. -[BugsnagClient notify:handledState:block:]
+     * 4. -[BSG_KSCrash captureThreads:depth:]
      */
     int depth = (int)(BSGNotifierStackFrameCount);
     NSArray *threads = [[BSG_KSCrash sharedInstance] captureThreads:exception depth:depth];
-    NSArray *errors = @[[self generateError:exception threads:threads]]; // TODO move to notifyinternal, or equivalent?
+    NSArray *errors = @[[self generateError:exception threads:threads]];
 
     NSDictionary *systemInfo = [BSG_KSSystemInfo systemInfo];
     BugsnagEvent *event = [[BugsnagEvent alloc] initWithApp:[self generateAppWithState:systemInfo]
@@ -1518,9 +1541,9 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
 }
 
 - (NSArray *)collectThreads {
-    return @[
-        // TODO implement
-    ];
+    int depth = (int)(BSGNotifierStackFrameCount);
+    NSException *exc = [NSException exceptionWithName:@"Bugsnag" reason:@"" userInfo:nil];
+    return [[BSG_KSCrash sharedInstance] captureThreads:exc depth:depth];
 }
 
 - (void)addRuntimeVersionInfo:(NSString *)info
--- a/Source/BugsnagClient.m
+++ b/Source/BugsnagClient.m
@@ -51,6 +51,7 @@
 #import "BugsnagStateEvent.h"
 #import "BugsnagCollections.h"
 #import "BSG_KSCrashReport.h"
+#import "BSG_KSCrash.h"
 
 #if TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
 #import <UIKit/UIKit.h>
@@ -62,7 +63,7 @@ NSString *const BSTabCrash = @"crash";
 NSString *const BSAttributeDepth = @"depth";
 NSString *const BSEventLowMemoryWarning = @"lowMemoryWarning";
 
-static NSInteger const BSGNotifierStackFrameCount = 6;
+static NSInteger const BSGNotifierStackFrameCount = 3;
 
 struct bugsnag_data_t {
     // Contains the state of the event (handled/unhandled)
@@ -118,10 +119,6 @@ static bool hasRecordedSessions;
 + (BugsnagDeviceWithState *)deviceWithDictionary:(NSDictionary *)event;
 - (NSDictionary *)toDictionary;
 - (void)appendRuntimeInfo:(NSDictionary *)info;
-@end
-
-@interface BugsnagThread ()
-- (instancetype)initWithThread:(NSDictionary *)thread binaryImages:(NSArray *)binaryImages;
 @end
 
 /**
@@ -326,6 +323,10 @@ void BSGWriteSessionCrashData(BugsnagSession *session) {
                      errors:(NSArray<BugsnagError *> *)errors
                     threads:(NSArray<BugsnagThread *> *)threads
                     session:(BugsnagSession *)session;
+@end
+
+@interface BugsnagError ()
+- (instancetype)initWithEvent:(NSDictionary *)event errorReportingThread:(BugsnagThread *)thread;
 @end
 
 @interface BSGOutOfMemoryWatchdog ()
@@ -832,7 +833,13 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
 // MARK: - Notify
 
 - (void)notifyError:(NSError *_Nonnull)error {
-    [self notifyError:error block:nil];
+    BugsnagHandledState *state = [BugsnagHandledState handledStateWithSeverityReason:HandledError
+                                                                            severity:BSGSeverityWarning
+                                                                           attrValue:error.domain];
+    NSException *wrapper = [NSException exceptionWithName:NSStringFromClass([error class])
+                                                   reason:error.localizedDescription
+                                                 userInfo:error.userInfo];
+    [self notify:wrapper handledState:state block:nil];
 }
 
 - (void)notifyError:(NSError *)error
@@ -866,7 +873,9 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
 }
 
 - (void)notify:(NSException *_Nonnull)exception {
-    [self notify:exception block:nil];
+    BugsnagHandledState *state =
+            [BugsnagHandledState handledStateWithSeverityReason:HandledException];
+    [self notify:exception handledState:state block:nil];
 }
 
 - (void)notify:(NSException *)exception 
@@ -907,14 +916,12 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
     NSDictionary *appState = @{@"oom": lastLaunchInfo, @"didOOM": @YES};
     [self.crashSentry reportUserException:BSGOutOfMemoryErrorClass
                                    reason:message
-                        originalException:nil
                              handledState:[handledState toJson]
                                  appState:appState
                         callbackOverrides:@{}
                            eventOverrides:@{}
                                  metadata:@{}
-                                   config:@{}
-                             discardDepth:0];
+                                   config:@{}];
 }
 
 - (void)notify:(NSException *)exception
@@ -939,6 +946,21 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
     event.apiKey = self.configuration.apiKey;
     event.context = self.context;
     event.originalError = exception;
+
+    /**
+     * Stack frames starting from this one are removed by setting the depth.
+     * This helps remove bugsnag frames from showing in NSErrors as their
+     * trace is synthesized.
+     *
+     * 0. +[Bugsnag notifyError:block:]
+     * 1. -[BugsnagClient notifyError:block:]
+     * 2. -[BugsnagClient notifyError:block:]
+     * 3. -[BSG_KSCrash captureThreads:depth:]
+     */
+    int depth = (int)(BSGNotifierStackFrameCount + event.depth);
+    event.threads = [[BSG_KSCrash sharedInstance] captureThreads:event.originalError depth:depth];
+    event.errors = [self generateErrors:event];
+
     [self notifyInternal:event block:block];
 }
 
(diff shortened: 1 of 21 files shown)


Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_User.c
```diff
@@ -147,10 +143,19 @@ void bsg_kscrashsentry_reportUserException(const char *name, const char *reason,
...
     }
 }
+
+void bsg_kscrashsentry_suspend_threads_user() {
+    pthread_mutex_lock(&bsg_suspend_threads_mutex);
+    BSG_KSLOG_DEBUG("Suspending all threads");
+    bsg_kscrashsentry_suspendThreads();
+}
+
+void bsg_kscrashsentry_resume_threads_user() {
```
tomlongridge, line 158: Worth adding a `terminateProgram` parameter so that it can be called from `bsg_kscrashsentry_reportUserException` too?

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1635,3 +1617,59 @@ void bsg_kscrashreport_logCrash(const BSG_KSCrash_Context *const crashContext) {
...
+    strcat(thread_json_data, tmp);
+    return BSG_KSJSON_OK;
+}
+
+char *bsg_kscrw_i_captureThreadTrace(const BSG_KSCrash_Context *crashContext) {
+    if (thread_json_data == NULL) {
+        thread_json_data = malloc(thread_json_size);
+    }
+    memset(thread_json_data, 0, strlen(thread_json_data));
+    thread_json_data[0] = '\0';
```
tomlongridge, line 1643: Can you move this to `bsg_kscrw_i_collectJsonData` so `thread_json_data` is managed in one place?

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_User.c
```diff
@@ -101,11 +101,7 @@ void bsg_kscrashsentry_reportUserException(const char *name, const char *reason,
         int callstackCount = 100;
         uintptr_t callstack[callstackCount];
 
-        if (localContext->suspendThreadsForUserReported) {
```
tomlongridge, line 104: Why are we not honouring `suspendThreadsForUserReported`?

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1635,3 +1617,59 @@ void bsg_kscrashreport_logCrash(const BSG_KSCrash_Context *const crashContext) {
...
 }
+
+size_t thread_json_size = sizeof(char) * 32 * 1024;
+char *thread_json_data;
+
+int bsg_kscrw_i_collectJsonData(const char *const data, const size_t length, void *const userData) {
+    size_t new_size = strlen(thread_json_data) + length;
+    char tmp[length];
+    memcpy(tmp, data, length);
+    tmp[length] = '\0';
```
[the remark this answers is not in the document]
fractalwrench: KSCrash does not pass non-terminated chars by default in its JSON codec. Altering the codec itself to use terminated chars resulted in KSCrash being unable to read the serialized report - but the `NSJSONSerialization` class does seem to require null-terminated char arrays.

fractalwrench:
Updated to fix the serialization of error reporting thread's stacktrace, which is recorded differently compared to all the other threads.

I've converted this to a draft PR in the meanwhile as there are a few more things that need testing out. Once that's done, the changeset should fix the population of both `event.threads` and `event.errors`.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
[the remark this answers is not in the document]
fractalwrench: Action: increase length of `tmp` by 1 to avoid overflow, and rename to `terminatedString`

Source/BugsnagClient.m
```diff
@@ -935,6 +940,7 @@ - (void)notifyInternal:(BugsnagEvent *_Nonnull)event
                  block:(BugsnagOnErrorBlock)block
 {
     [self populateEventData:event];
+    event.threads = [self generateThreads];
```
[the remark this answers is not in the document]
fractalwrench: This is now called from within `notify:handledState:block:` which is where every other field is now poulated.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1635,3 +1617,59 @@ void bsg_kscrashreport_logCrash(const BSG_KSCrash_Context *const crashContext) {
...
+    // Don't write the binary images for user reported crashes to improve performance
+    if (crashContext->crash.writeBinaryImagesForUserReported == true ||
+            crashContext->crash.crashType != BSG_KSCrashTypeUserReported) {
+        bsg_kscrw_i_writeBinaryImages(writer, BSG_KSCrashField_BinaryImages);
+    }
+    writer->beginObject(writer, BSG_KSCrashField_Crash);
+    {
+    // Conditionally write threads depending on user configuration
+    int sendPolicy = crashContext->crash.threadTracingEnabled;
+    bool unhandledCrash = crashContext->crash.crashType != BSG_KSCrashTypeUserReported;
```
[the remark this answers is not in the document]
fractalwrench: Moved to top of function.

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_User.c
fractalwrench: Added parameter and removed duplicate code.

Source/BugsnagClient.m
```diff
@@ -1475,6 +1481,44 @@ - (BugsnagDeviceWithState *)generateDeviceWithState:(NSDictionary *)systemInfo {
...
+                if (error != nil) {
+                    bsg_log_err(@"Failed to deserialize JSON: %@", error);
+                } else {
+                    return decode;
+                }
+            }
+        }
+    }
+    return nil;
+}
```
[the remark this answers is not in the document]
fractalwrench: Moved function to `BSGSerialization`.

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_User.c
fractalwrench: This is always set to `true` and is an internal property of KSCrash which isn't exposed. For now I've reinstated the checks here but it would make sense to fully remove `suspendThreadsForUserReported` at a later date as it's not used.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashC.h
```diff
@@ -182,6 +182,14 @@ void bsg_kscrash_setWriteBinaryImagesForUserReported(
  */
 BSG_KSCrash_Context *crashContext(void);
 
+/**
+ * Captures a thread trace for the current application state, if the user
+ * has configured this functionality.
```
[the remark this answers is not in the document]
fractalwrench: `bsg_kscrw_i_writeAllThreads` takes a parameter `recordAllThreads` which controls whether all threads are captured. If this parameter is false, the threads are still suspended but only the main thread's stacktrace is serialized.

I've updated the function's documentation accordingly

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
fractalwrench: Extracted to `bsg_kscrw_i_resetThreadTraceData` as suggested by in-person discussion.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
```diff
@@ -306,41 +316,64 @@ - (void)deleteAllReports {
...
+- (NSArray<BugsnagThread *> *)captureThreads:(NSException *)exc depth:(int)depth {
+    NSArray *addresses = [exc callStackReturnAddresses];
+    int numFrames = (int) [addresses count];
+    uintptr_t *callstack = malloc(numFrames * sizeof(*callstack));
+
+    for (NSUInteger i = 0; i < numFrames; i++) {
+        callstack[i] = [addresses[i] unsignedLongValue];
+    }
+    if (numFrames > 0) {
+        depth = 0; // reset depth if the stack does not need to be generated
```
tomlongridge, line 328: This took some parsing in my head!

I think it would make the intent clearer if the work to populate callstack from the exception happens inside the `numFrames > 0` path, then we don't need to realloc and some comments can describe the two cases separately.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1635,3 +1619,61 @@ void bsg_kscrashreport_logCrash(const BSG_KSCrash_Context *const crashContext) {
...
+static size_t bsg_g_thread_json_size = sizeof(char) * 32 * 1024;
+static char *bsg_g_thread_json_data;
+
+int bsg_kscrw_i_collectJsonData(const char *const data, const size_t length, void *const userData) {
+    size_t new_size = strlen(bsg_g_thread_json_data) + length;
+    char terminated_str[length + 1];
+    memcpy(terminated_str, data, length);
+    terminated_str[length] = '\0';
+
+    if (new_size > bsg_g_thread_json_size) {
```
tomlongridge, line 1632: I believe the data may or may not have a terminator already (I can see calls from KS code that passes data as a string literal which will be terminated.

Therefore by adding our own and then `strlen`-ing the array, we will be able to pass it to `strlen` to give us the length we need. So I think you can do away with `new_size` and just use:

```suggestion
    if (strlen(bsg_g_thread_json_data) + strlen(terminated_str) > bsg_g_thread_json_size) {
```

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_User.c
```diff
@@ -96,34 +93,12 @@ void bsg_kscrashsentry_reportUserException(const char *name, const char *reason,
         BSG_KSCrash_Context *reportContext = bsg_kscrashsentry_generateReportContext();
         BSG_KSCrash_SentryContext *localContext = &reportContext->crash;
 
-        // We want these variables to persist until the onCrash
-        // call later
-        int callstackCount = 100;
```
tomlongridge, line 101: I'm unclear what this function is now doing now - does it need to suspend threads?

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_User.h
```diff
@@ -54,32 +54,29 @@ void bsg_kscrashsentry_uninstallUserExceptionHandler(void);
...
         const char *eventOverrides,
         const char *metadata,
         const char *appState,
         const char *config,
-        int discardDepth,
         bool terminateProgram);
 
+void bsg_kscrashsentry_suspend_threads_user(void);
+
+void bsg_kscrashsentry_resume_threads_user(bool terminateProgram);
```
tomlongridge, line 78: Doc comments please

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
```diff
@@ -306,41 +316,64 @@ - (void)deleteAllReports {
     [self.crashReportStore deleteAllFiles];
 }
 
+- (NSArray<BugsnagThread *> *)captureThreads:(NSException *)exc depth:(int)depth {
+    NSArray *addresses = [exc callStackReturnAddresses];
+    int numFrames = (int) [addresses count];
```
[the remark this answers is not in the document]
fractalwrench: `[addresses count]` returns an `NSUInteger` which is an unsigned long - this requires casting to an `int` as this is what KSCrash uses for stackframe counts throughout.

Source/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_User.h
fractalwrench: Added docs to the header file.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
fractalwrench: Updated code to use this suggestion.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
fractalwrench: Refactored this so that callstack allocation happens within the `numFrames > 0` path.

Source/BugsnagClient.m
```diff
@@ -976,36 +981,51 @@ - (void)notifyInternal:(BugsnagEvent *_Nonnull)event
...
+    BugsnagThread *errorReportingThread;
+
+    for (BugsnagThread *thread in event.threads) {
+        if (thread.errorReportingThread) {
+            errorReportingThread = thread;
+            break;
+        }
+    }
+
+    BugsnagError *error = [[BugsnagError alloc] initWithEvent:@{} errorReportingThread:errorReportingThread];
```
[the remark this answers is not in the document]
fractalwrench: Created `initWithErrorReportingThread` as a wrapper around this constructor to better signify the event information is not required for this method of instantiation.

Source/BugsnagClient.m
```diff
@@ -929,6 +941,21 @@ - (void)notify:(NSException *)exception
...
+     * trace is synthesized.
+     *
+     * 0. +[Bugsnag notifyError:block:]
+     * 1. -[BugsnagClient notifyError:block:]
+     * 2. -[BugsnagClient notifyError:block:]
+     * 3. -[BSG_KSCrash captureThreads:depth:]
+     */
+    int depth = (int)(BSGNotifierStackFrameCount + event.depth);
+    event.threads = [[BSG_KSCrash sharedInstance] captureThreads:event.originalError depth:depth];
+    event.errors = [self generateErrors:event];
```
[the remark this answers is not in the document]
fractalwrench: Refactored so that the errors/threads are captured before creating a `BugsnagEvent` object, avoiding unnecessary duplication.

Source/BugsnagClient.m
```diff
@@ -822,7 +828,13 @@ - (NSString *)context {
...
 
 - (void)notifyError:(NSError *_Nonnull)error {
-    [self notifyError:error block:nil];
+    BugsnagHandledState *state = [BugsnagHandledState handledStateWithSeverityReason:HandledError
+                                                                            severity:BSGSeverityWarning
+                                                                           attrValue:error.domain];
+    NSException *wrapper = [NSException exceptionWithName:NSStringFromClass([error class])
+                                                   reason:error.localizedDescription
+                                                 userInfo:error.userInfo];
+    [self notify:wrapper handledState:state block:nil];
```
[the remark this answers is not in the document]
fractalwrench: I've reverted this change as I don't believe it's necessary

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
```diff
@@ -306,41 +316,66 @@ - (void)deleteAllReports {
...
+
+        BSG_KSLOG_DEBUG("Fetching call stack.");
+        numFrames = backtrace((void **)callstack, numFrames);
+        if (numFrames <= 0) {
+            BSG_KSLOG_ERROR(@"backtrace() returned call stack length of %d", numFrames);
+            numFrames = 0;
+        }
+    }
+
+    char *trace = bsg_kscrash_captureThreadTrace(depth, numFrames, callstack);
```
fractalwrench, line 345: This method currently returns a global variable which can be mutated on another thread once `bsg_kscrash_captureThreadTrace` has resumed thread execution.

Source/BugsnagClient.m
```diff
@@ -981,36 +1001,55 @@ - (void)notifyInternal:(BugsnagEvent *_Nonnull)event
...
+            errorReportingThread = thread;
+            break;
+        }
+    }
+
+    BugsnagError *error = [[BugsnagError alloc] initWithErrorReportingThread:errorReportingThread];
+    error.errorClass = errorClass;
+    error.errorMessage = errorMessage ?: @"";
+    return error;
+}
```
tomlongridge, line 1051: I don't see any reason we can't use `initWithErrorClass` for this - we basically just want to set class, message and type. Passing in the `errorReportingThread` value doesn't seem to do anything with an empty `event` parameter.

Source/BugsnagEvent.m
```diff
@@ -569,6 +569,11 @@ - (instancetype)initWithUserData:(NSDictionary *)event {
     obj.context = bugsnagPayload[@"context"];
     obj.groupingHash = bugsnagPayload[@"groupingHash"];
     obj.error = [obj getMetadataFromSection:BSGKeyError];
+
+    if ([errors count] > 0) {
+        BugsnagError *bugsnagError = errors[0];
+        obj.customException = BSGParseCustomException(event, bugsnagError.errorClass, bugsnagError.errorMessage);
```
tomlongridge, line 575: Should the two strings be copied like in `initWithKSCrashData`? Also, can you rename `bugsnagError` to something else as this implies to me that it's some sort of internally generated construct.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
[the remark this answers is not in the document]
tomlongridge: It didn't before, which is why I was questioning it. Why is this different?

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1635,3 +1621,65 @@ void bsg_kscrashreport_logCrash(const BSG_KSCrash_Context *const crashContext) {
...
 }
+
+static const size_t bsg_g_buffer_increment = sizeof(char) * 8 * 1024;
+static size_t bsg_g_thread_json_size = 128 * 1024;
+static char *bsg_g_thread_json_data;
+
+int bsg_kscrw_i_collectJsonData(const char *const data, const size_t length, void *const userData) {
+    if (bsg_kscrw_i_exceedsBufferLen(length)) {
+        bsg_g_thread_json_size += bsg_g_buffer_increment;
+        bsg_g_thread_json_data = realloc(bsg_g_thread_json_data, bsg_g_thread_json_size);
```
tomlongridge, line 1632: It seems to be an anti-pattern to assign the result straight back to the original pointer as it doesn't all you to free the memory. I guess in this case it's a single static area, but would it not be cleaner to `free` and return the error code immediately after `realloc` if it fails rather than introduce a second if-clause to check this later?

Source/BugsnagClient.m
fractalwrench: Updated to use existing constructor.

Source/BugsnagClient.m
```diff
@@ -1502,9 +1541,9 @@ - (NSArray *)collectBreadcrumbs {
 }
 
 - (NSArray *)collectThreads {
-    return @[
-        // TODO implement
-    ];
+    int depth = (int)(BSGNotifierStackFrameCount);
+    NSException *exc = [NSException exceptionWithName:@"Bugsnag" reason:@"" userInfo:nil];
+    return [[BSG_KSCrash sharedInstance] captureThreads:exc depth:depth];
 }
```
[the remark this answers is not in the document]
fractalwrench: This will be used by React Native.

Source/BugsnagEvent.m
fractalwrench: I don't see any reason why `initWithKSCrashData` needs to copy the string - I assume this is just done for legacy reasons. Renamed variable.

Source/BugsnagClient.m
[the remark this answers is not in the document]
tomlongridge: Can you some comment docs on this, especially as it's a bit of a quirky thing to do!

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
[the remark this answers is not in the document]
fractalwrench: Previously KSCrash stored the value of `backtrace()` in an `int` as it was calculated in a different file. `backtrace()` returns an `int` which is why this has been changed here:
https://github.com/bugsnag/bugsnag-cocoa/pull/601/files#diff-14847c7e626ff843aeb8c965b155b829L115

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
fractalwrench: Implemented suggestion and returned immediately in the error case.

PR #602, opened 2020-05-15
Sanitize dictionary to prevent NSNull values

fractalwrench:
## Goal

The `BugsnagMetadata` implementation crashes if `NSNull` is present in the dictionary. To protect against invalid input the metadata implementation should be defensive and sanitize when reading information from disk, as otherwise this could lead to a crash when attempting to deliver.

## Changeset

When initializing `BugsnagMetadata` with a dictionary the section and its values are now checked for null, and any null values are removed. Unit test cases have been added to verify this behaviour.

### DIFF
--- a/Source/BugsnagMetadata.m
+++ b/Source/BugsnagMetadata.m
@@ -45,14 +45,25 @@
     if (self = [super init]) {
         // Ensure that the instantiating dictionary is mutable.
         // Saves checks later.
-        self.dictionary = [self sanitizeDictionary:[dict mutableCopy]];
+        self.dictionary = [self sanitizeDictionary:dict];
         self.stateEventBlocks = [NSMutableArray new];
     }
     [self notifyObservers];
     return self;
 }
 
-- (NSMutableDictionary *)sanitizeDictionary:(NSMutableDictionary *)input {
+/**
+ * Sanitizes the given dictionary to prevent [NSNull null] values from being added
+ * to the metadata when deserializing a payload.
+ *
+ * @param dictionary the input dictionary
+ * @return a sanitized dictionary
+ */
+- (NSMutableDictionary *)sanitizeDictionary:(NSDictionary *)dictionary {
+    NSMutableDictionary *input;
+    if (dictionary != nil && [dictionary isKindOfClass:[NSDictionary class]]) {
+        input = [dictionary mutableCopy];
+    }
     for (NSString *key in [input allKeys]) {
         id obj = input[key];
 
@@ -63,6 +74,12 @@
         } else if ([obj isKindOfClass:[NSArray class]]) {
             NSMutableArray *ary = [obj mutableCopy];
             [ary removeObject:[NSNull null]];
+
+            for (NSUInteger k = 0; k < [ary count]; ++k) {
+                if ([ary[k] isKindOfClass:[NSDictionary class]]) {
+                    ary[k] = [self sanitizeDictionary:ary[k]];
+                }
+            }
             input[key] = ary;
         }
     }
--- a/Source/BugsnagMetadata.m
+++ b/Source/BugsnagMetadata.m
@@ -61,7 +61,7 @@
  */
 - (NSMutableDictionary *)sanitizeDictionary:(NSDictionary *)dictionary {
     NSMutableDictionary *input;
-    if (dictionary != nil && [dictionary isKindOfClass:[NSDictionary class]]) {
+    if (dictionary != nil) {
         input = [dictionary mutableCopy];
     }
     for (NSString *key in [input allKeys]) {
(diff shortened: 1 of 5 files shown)


Source/BugsnagMetadata.m
```diff
@@ -45,13 +45,30 @@ - (id)initWithDictionary:(NSDictionary *)dict {
...
+- (NSMutableDictionary *)sanitizeDictionary:(NSMutableDictionary *)input {
+    for (NSString *key in [input allKeys]) {
+        id obj = input[key];
+
+        if (obj == [NSNull null]) {
+            [input removeObjectForKey:key];
+        } else if ([obj isKindOfClass:[NSDictionary class]]) {
+            input[key] = [self sanitizeDictionary:[obj mutableCopy]];
+        } else if ([obj isKindOfClass:[NSArray class]]) {
+            NSMutableArray *ary = [obj mutableCopy];
```
tomlongridge, line 64: Presumably the array could contain dictionaries?

Source/BugsnagMetadata.m
```diff
@@ -45,13 +45,30 @@ - (id)initWithDictionary:(NSDictionary *)dict {
     if (self = [super init]) {
         // Ensure that the instantiating dictionary is mutable.
         // Saves checks later.
-        self.dictionary = [dict mutableCopy];
+        self.dictionary = [self sanitizeDictionary:[dict mutableCopy]];
```
tomlongridge, line 48: Suggest moving the mutable copy inside sanitizeDictionary to better contain the functionalty.

Source/BugsnagMetadata.m
```diff
@@ -45,13 +45,30 @@ - (id)initWithDictionary:(NSDictionary *)dict {
...
         self.stateEventBlocks = [NSMutableArray new];
     }
     [self notifyObservers];
     return self;
 }
 
+- (NSMutableDictionary *)sanitizeDictionary:(NSMutableDictionary *)input {
```
tomlongridge, line 55: Be good to have some comments to inform future readers why this is necessary.

Source/BugsnagMetadata.m
fractalwrench: Handled this case by recursing.

tomlongridge: I think you're still missing the case for an array of arrays:
```json
{
  [
    [
      { 'key': [NSNull null] },
      [NSNull null]
    ]
  ]
}
```

Source/BugsnagMetadata.m
```diff
@@ -45,13 +45,47 @@ - (id)initWithDictionary:(NSDictionary *)dict {
...
+/**
+ * Sanitizes the given dictionary to prevent [NSNull null] values from being added
+ * to the metadata when deserializing a payload.
+ *
+ * @param dictionary the input dictionary
+ * @return a sanitized dictionary
+ */
+- (NSMutableDictionary *)sanitizeDictionary:(NSDictionary *)dictionary {
+    NSMutableDictionary *input;
+    if (dictionary != nil && [dictionary isKindOfClass:[NSDictionary class]]) {
```
tomlongridge, line 64: I believe you need to invert the check to `![dictionary isKindOfClass:[NSMutableDictionary class]]` as `NSDictionary ` extends `NSDictionary` so would always be true.

Source/BugsnagMetadata.m
```diff
@@ -45,13 +45,47 @@ - (id)initWithDictionary:(NSDictionary *)dict {
...
+        input = [dictionary mutableCopy];
+    }
+    for (NSString *key in [input allKeys]) {
+        id obj = input[key];
+
+        if (obj == [NSNull null]) {
+            [input removeObjectForKey:key];
+        } else if ([obj isKindOfClass:[NSDictionary class]]) {
+            input[key] = [self sanitizeDictionary:[obj mutableCopy]];
```
tomlongridge, line 73: `mutableCopy` isn't required as that will be done inside `sanitizeDictionary `.

Tests/BugsnagMetadataTests.m
```diff
@@ -262,4 +262,50 @@ - (void)testMetadataMutability {
...
+    BugsnagMetadata *metadata = [[BugsnagMetadata alloc] initWithDictionary:@{
+            @"foo": @{
+                    @"bar": @YES,
+                    @"custom": @[[NSNull null], @"foo"]
+            }
+    }];
+    XCTAssertEqual(1, [metadata.dictionary count]);
+    XCTAssertEqualObjects(@[@"foo"], [metadata getMetadataFromSection:@"foo" withKey:@"custom"]);
+}
+
```
tomlongridge, line 310: Does this pass?
```
BugsnagMetadata *metadata = [[BugsnagMetadata alloc] initWithDictionary:@{
            @"foo": @{
                    @"bar": @[
                        @[
                            @{ @"custom": [NSNull null] }
                        ]
                    ]
            }
    }];
```

Source/BugsnagMetadata.m
tomlongridge: That's not quite it - added suggested test below.

Tests/BugsnagMetadataTests.m
fractalwrench: I've added this as a test case and created a separate method for sanitising arrays.

PR #606, opened 2020-05-18
V6 uat swift api checks

robinmacharg:
## Goal

Implement a test that verifies that all public interfaces' Swift names are as expected.

## Design

Manual walkthrough of public headers, as noted in the podspec json.

## Tests

See `BugsnagSwiftPublicAPITests.swift`

### Outstanding Questions

  - [X] Final review
  - [ ] Release

  features)


Source/BugsnagConfiguration.m
```diff
@@ -531,7 +532,7 @@ - (NSArray *)enabledReleaseStages {
 - (void)setEnabledReleaseStages:(NSArray *)newReleaseStages;
 {
     @synchronized (self) {
-        NSArray *releaseStagesCopy = [newReleaseStages copy];
+        NSSet<NSString *> *releaseStagesCopy = [newReleaseStages copy];
         _enabledReleaseStages = releaseStagesCopy;
         [self.config addMetadata:releaseStagesCopy
```
fractalwrench, line 537: Will this still be serialized when adding a set? I'm unsure off the top of my head whether BugsnagMetadata handles anything complex other than dictionaries + arrays.

Tests/Swift Tests/BugsnagSwiftPublicAPITests.swift
```diff
@@ -0,0 +1,368 @@
...
+//
+
+import XCTest
+
+/**
+ * Test all public APIs from Swift.  Purely existence tests, no attempt to verify correctness
+ */
+
+class FakePlugin: NSObject, BugsnagPlugin {
+    func load(_ client: BugsnagClient!) {}
```
fractalwrench, line 16: Should this be `BugsnagClient` instead?

Tests/Swift Tests/BugsnagSwiftPublicAPITests.swift
```diff
@@ -0,0 +1,368 @@
...
+        Bugsnag.pauseSession()
+        Bugsnag.resumeSession()
+        
+        Bugsnag.setContext("ctx")
+        let _ = Bugsnag.context()
+        
+        Bugsnag.setUser("me", withEmail: "memail@foo.com", andName: "you")
+        let _ = Bugsnag.user()
+        
+        Bugsnag.addOnSession(block: sessionBlock)
```
fractalwrench, line 74: would be good to verify that trailing block closures are supported too (ditto for config/client):

```
Bugsnag.addOnSession { (block) -> Bool in
    return true;
}
```

Tests/Swift Tests/BugsnagSwiftPublicAPITests.swift
```diff
@@ -0,0 +1,368 @@
...
+        let _ = BugsnagConfiguration.loadConfig()
+        let config = BugsnagConfiguration(apiKey)
+
+        config.apiKey = apiKey
+        config.releaseStage = "stage1"
+        config.enabledReleaseStages = nil
+        config.enabledReleaseStages = ["one", "two", "three"]
+        config.redactedKeys = nil
+        config.redactedKeys = ["1", 2, 3]
+        config.redactedKeys = ["a", "a", "b"]
```
fractalwrench, line 94: `config.redactedKeys` should also allow `NSRegularExpression` objects in the same array

Tests/Swift Tests/BugsnagSwiftPublicAPITests.swift
```diff
@@ -0,0 +1,368 @@
...
+        
+        md.addMetadata(["key" : "secret"], section: "mental")
+        md.addMetadata("spock", key: "kirk", section: "enterprise")
+        md.getMetadata(section: "mental")
+        md.getMetadata(section: "mental", key: "key")
+        md.clearMetadata(section: "enterprise")
+        md.clearMetadata(section: "enterprise", key: "key")
+    }
+    
+    func testBugsnagEventClass() throws {
```
fractalwrench, line 154: `event.severity` could use testing

Tests/Swift Tests/BugsnagSwiftPublicAPITests.swift
```diff
@@ -0,0 +1,368 @@
...
+        event.threads = [BugsnagThread()]
+        event.originalError = nil
+        event.originalError = 123
+        event.originalError = BugsnagError()
+        _ = event.user
+        event.setUser("user", withEmail: "email", andName: "name")
+    }
+    
+    func testBugsnagAppClass() throws {
+        let app = BugsnagApp()
```
fractalwrench, line 179: `BugsnagAppWithState` and `BugsnagDeviceWithState` should be checked instead as they include additional stateful fields

Tests/Swift Tests/BugsnagSwiftPublicAPITests.swift
```diff
@@ -0,0 +1,368 @@
...
+        sf.method = nil
+        sf.method = "method"
+        _ = sf.method
+        sf.machoFile = nil
+        sf.machoFile = "file"
+        _ = sf.machoFile
+        sf.machoUuid = nil
+        sf.machoUuid = "uuid"
+        _ = sf.machoUuid
+        sf.frameAddress = 123
```
fractalwrench, line 335: This can now be nil as of #614, along with additional properties in that changeset

Tests/Swift Tests/BugsnagSwiftPublicAPITests.swift
robinmacharg: This is what was offered by XCode's autofix.  I've updated the protocol definitiion to be explicitly `_Nonnull` for the client arg which fixes things.

Source/BugsnagConfiguration.m
robinmacharg: Good point.  I've reverted all `NSSet` use, and created an internal ticket to address the `NSSet` vs. `NSArray` issue more fully when time permits.

fractalwrench: I think it'd be preferable to use `NSSet` rather than `NSArray` on our public headers, as after the next major we won't be able to alter the type without making a breaking change. If it's going to take a long time to implement serialization of sets in metadata, then altering the implementation so that it continues to use `NSArray` under the hood seems good to me.

robinmacharg: Public API now presents `NSSet`.

PR #607, opened 2020-05-18
Fix BugsnagEvent related mazerunner tests

fractalwrench:
## Goal

Fixes several `BugsnagEvent` related mazerunner tests which are failing as a result of #585. Note that some E2E scenarios will still fail, but will be addressed in #601 as they relate to stacktraces/threads.

## Changeset

- `BugsnagClient` now sets a default autogenerated user ID if no value was set in `BugsnagConfiguration`, which matches the notifier spec behaviour. The default user ID is applied to all events and can be removed by calling `Bugsnag.setUser(nil, nil, nil)`. Related test assertions have been updated.
- The `time` property on `BugsnagDevice` was not populated for handled errors because it was searching for `report.timestamp`, which is only present in a full KSCrash report. The function which collects device system information has now been amended


features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/UserDisabledScenario.swift
```diff
@@ -16,7 +16,6 @@ internal class UserDisabledScenario: Scenario {
     }
 
     override func run() {
-        Bugsnag.setUser(nil, withEmail: nil, andName: nil)
```
tomlongridge, line 19: Would it be better to change expectation on this scenario that all fields are null, but then add an extra scenario for default behaviour?

fractalwrench: Added an extra scenario which verifies the default user information only contains an ID.

PR #608, opened 2020-05-19
Disable collection of KSCrash report fields in handled errors

fractalwrench:
## Goal

Handled errors now save the entire payload as JSON matching the Bugsnag Error API schema. Therefore KSCrash report fields should not be serialized where they duplicate this information, as it increases the size of the report on disk and the time taken to write to file.

## Changeset

- Added `recordKSCrashFields` boolean which defaults to true, and can be disabled for user-reported events
- Gated serialization of KSCrash report fields using flag, with the exception of `bsg_kscrw_i_writeReportInfo` which other deserialization code relies upon when reading a stored report

## Tests
Sent a handled + unhandled error to the dashboard and verified they contain the necessary fields. Inspected the payload stored on disk for each and verified only the unhandled error contains KSCrash report fields.

### DIFF
--- a/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
+++ b/Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
@@ -619,7 +619,7 @@ void bsg_kscrw_i_writeTraceInfo(const BSG_KSCrash_Context *crashContext, const B
 
 bool bsg_kscrw_i_exceedsBufferLen(const size_t length);
 
-void bsg_kscrashreport_writeKSCrashFields(BSG_KSCrash_Context *crashContext, BSG_KSCrashReportWriter *writer, bool recordKSCrashFields);
+void bsg_kscrashreport_writeKSCrashFields(BSG_KSCrash_Context *crashContext, BSG_KSCrashReportWriter *writer);
 
 void bsg_kscrashreport_writeBugsnagPayload(const BSG_KSCrash_Context *crashContext, const BSG_KSCrashReportWriter *writer);
 
@@ -1549,11 +1549,19 @@ void bsg_kscrashreport_writeStandardReport(
     bsg_ksjsonbeginEncode(bsg_getJsonContext(writer), true,
                           bsg_kscrw_i_addJSONData, &fd);
 
+    // KSCrash report fields are not required for handled errors as
+    // they serialize info in the bugsnag payload JSON schema instead.
     bool recordKSCrashFields = crashContext->crash.userException.eventOverrides == NULL;
 
     writer->beginObject(writer, BSG_KSCrashField_Report);
     {
-        bsg_kscrashreport_writeKSCrashFields(crashContext, writer, recordKSCrashFields);
+        bsg_kscrw_i_writeReportInfo(
+                writer, BSG_KSCrashField_Report, BSG_KSCrashReportType_Standard,
+                crashContext->config.crashID, crashContext->config.processName);
+
+        if (recordKSCrashFields) {
+            bsg_kscrashreport_writeKSCrashFields(crashContext, writer);
+        }
 
         if (crashContext->config.onCrashNotify != NULL) {
             // NOTE: The blacklist for BSG_KSCrashField_UserAtCrash children in BugsnagEvent.m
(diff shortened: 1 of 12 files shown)


Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1542,76 +1548,95 @@ void bsg_kscrashreport_writeStandardReport(
                           bsg_kscrw_i_addJSONData, &fd);
 
     writer->beginObject(writer, BSG_KSCrashField_Report);
-    { // TODO avoid collecting these fields for handled errors. they're not required at all.
-        bsg_kscrw_i_writeReportInfo(
+    {
+        bsg_kscrashreport_writeKSCrashFields(crashContext, writer);
```
tomlongridge, line 1552: It would be cleaner to condition on the presence of `eventOverrides` to determine what is output.

The call to `bsg_kscrw_i_writeReportInfo` could be brought up to this level, then the calls to `bsg_kscrashreport_writeKSCrashFields`, `bsg_kscrashreport_writeBugsnagPayload` and `bsg_kscrashreport_writeOverrides` can each be conditioned on `crashContext->crash.userException.eventOverrides`. This would mean we wouldn't need to pass an extra parameter all the way down and it's only this function that needs to care what is written.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1645,7 +1670,7 @@ char *bsg_kscrw_i_captureThreadTrace(const BSG_KSCrash_Context *crashContext) {
     BSG_KSCrashReportWriter concreteWriter;
     BSG_KSCrashReportWriter *writer = &concreteWriter;
     bsg_kscrw_i_prepareReportWriter(writer, &jsonContext);
-    bsg_ksjsonbeginEncode(bsg_getJsonContext(writer), false, bsg_kscrw_i_collectJsonData, 0);
```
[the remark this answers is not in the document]
fractalwrench: Disabled once again.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
fractalwrench: Updated implementation to check for `eventOverrides` rather than passing a boolean parameter. This required changing the OOM notify call to pass in `nil` rather than a dictionary, as OOMs do require the KSCrash report fields.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1543,77 +1549,99 @@ void bsg_kscrashreport_writeStandardReport(
...
+            // should be updated when adding new fields here
+
+            // Write handled exception report info
+            writer->beginObject(writer, BSG_KSCrashField_UserAtCrash);
+            if (crashContext->crash.crashType == BSG_KSCrashTypeUserReported) {
+                bsg_kscrashreport_writeBugsnagPayload(crashContext, writer);
+
+                if (recordKSCrashFields) {
+                    bsg_kscrashreport_writeOverrides(crashContext, writer);
+                }
```
tomlongridge, line 1569:
```suggestion
                if (recordKSCrashFields) {
                    bsg_kscrashreport_writeOverrides(crashContext, writer);
                } else {
                    bsg_kscrashreport_writeBugsnagPayload(crashContext, writer);
                }
```

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1543,77 +1549,99 @@ void bsg_kscrashreport_writeStandardReport(
...
+        writer->addJSONElement(writer, BSG_KSCrashField_Config,
+                crashContext->crash.userException.config);
+    }
+    writer->addIntegerElement(writer, BSG_KSCrashField_DiscardDepth,
+            crashContext->crash.userException.discardDepth);
+}
+
+void bsg_kscrashreport_writeKSCrashFields(BSG_KSCrash_Context *crashContext,
+                                          BSG_KSCrashReportWriter *writer, 
+                                          bool recordKSCrashFields) {
```
tomlongridge, line 1618: If we pull the `writeReportInfo` up a level, there's no need to pass the param down and all the logic around which fields to write is nicely contained in one function.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1543,77 +1549,99 @@ void bsg_kscrashreport_writeStandardReport(
     bsg_ksjsonbeginEncode(bsg_getJsonContext(writer), true,
                           bsg_kscrw_i_addJSONData, &fd);
 
+    bool recordKSCrashFields = crashContext->crash.userException.eventOverrides == NULL;
```
tomlongridge, line 1552: This feels like a good point to add some comments about why different fields are included/excluded.

fractalwrench: Added inline comment.

Source/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
fractalwrench: Amended as suggested.

PR #610, opened 2020-05-21
Alter state event sync to support React Native

fractalwrench:
## Goal

Alters the state syncing implementation to support React Native changes.

## Changeset

- Add a `removeObserverWithBlock` method, and rename `addObserverWithBlock` for consistency
- As `addObserverWithBlock` can be called externally it now only adds observers to `client` and `client.metadata`. Previously it was adding observers for `self.configuration.config` and others, which led to metadata being overwritten in the React Native implementation. Observers for these within `BugsnagClient` are now set separately.

## Tests
Added unit test for verifying block removal, also tested in React Native example app to verify only one message was sent after removing and adding an observer.


Source/BugsnagClient.m
```diff
@@ -435,14 +439,18 @@ - (id)initWithConfiguration:(BugsnagConfiguration *)initConfiguration {
...
-    [self.configuration.metadata addObserverUsingBlock:observer];
-    [self.configuration.config addObserverUsingBlock:observer];
-    [self.state addObserverUsingBlock:observer];
+    [self.metadata addObserverWithBlock:observer];
+}
+
+- (void)removeObserverWithBlock:(BugsnagObserverBlock _Nonnull)observer {
+    [self.stateEventBlocks removeObject:observer];
+
+    // additionally listen for metadata updates
```
tomlongridge, line 452:
```suggestion
    // additionally remove metadata listener
```

PR #612, opened 2020-05-21
Move onCrashHandler data to individual metadata sections

fractalwrench:
## Goal

Rather than adding `onCrashhandler` into a separate section with a key of "custom", the data is now a string:map structure that is merged in to the top level of existing metadata. Any other data elements that are not a dictionary are dropped.

Updated unit tests to verify and tested that the behaviour was as expected when sending an event from an example app to the dashboard.

### DIFF
--- a/Source/BugsnagEvent.m
+++ b/Source/BugsnagEvent.m
@@ -477,7 +477,9 @@ NSDictionary *BSGParseCustomException(NSDictionary *report,
 
     NSMutableDictionary *userAtCrash = [self parseOnCrashData:event];
     if (userAtCrash != nil) {
-        [metadata addMetadata:userAtCrash toSection:@BSG_KSCrashField_OnCrashMetadataSectionName];
+        for (NSString *section in [userAtCrash allKeys]) {
+            [metadata addMetadata:userAtCrash[section] toSection:section];
+        }
     }
     NSString *deviceAppHash = [event valueForKeyPath:@"system.device_app_hash"];
     BugsnagDeviceWithState *device = [BugsnagDeviceWithState deviceWithDictionary:event];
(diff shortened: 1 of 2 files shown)


Source/BugsnagEvent.m
```diff
@@ -601,6 +603,12 @@ - (NSMutableDictionary *)parseOnCrashData:(NSDictionary *)report {
             @"id",
     ];
     [userAtCrash removeObjectsForKeys:blacklistedKeys];
+
+    for (NSString *key in [userAtCrash allKeys]) { // remove any non-dictionary values
+        if (![userAtCrash[key] isKindOfClass:[NSDictionary class]]) {
+            [userAtCrash removeObjectForKey:key];
```
tomlongridge, line 609: Suggest we log a warning here?

Tests/BugsnagOnCrashTest.m
```diff
@@ -82,16 +82,18 @@ - (void)testBlacklistedFields {
...
                                     @"age": @76
                             }
                     }
             }
     }];
-    NSMutableDictionary *data = [event getMetadataFromSection:@"onCrash"];
+    NSMutableDictionary *data = [event getMetadataFromSection:@"customer"];
     XCTAssertNotNil(data);
     XCTAssertEqual(2, [data count]);
     XCTAssertEqualObjects(@"Joe Bloggs", data[@"name"]);
```
tomlongridge, line 99: Seems like a test for removal of a non-conformant key would be simple and worthwhile.

fractalwrench: Added an additional non-conforming value to this test case.

PR #614, opened 2020-05-21
Use NSNumber rather than int for public properties on BugsnagEvent

fractalwrench:
## Goal

Uses `NSNumber` rather than int on properties on `BugsnagEvent`, as `NSNumber` is nullable and allows for redaction of its values. This better matches the notifier specification.

## Tests

Updated existing unit tests to handle new property types.

### DIFF
--- a/Source/BugsnagAppWithState.m
+++ b/Source/BugsnagAppWithState.m
@@ -72,11 +72,11 @@
     NSDictionary *stats = system[@"application_stats"];
 
     // convert from seconds to milliseconds
-    NSUInteger activeTimeSinceLaunch = (NSUInteger) ([stats[@"active_time_since_launch"] longValue] * 1000);
-    NSUInteger backgroundTimeSinceLaunch = (NSUInteger) ([stats[@"background_time_since_launch"] longValue] * 1000);
+    NSNumber *activeTimeSinceLaunch = @([stats[@"active_time_since_launch"] longValue] * 1000);
+    NSNumber *backgroundTimeSinceLaunch = @([stats[@"background_time_since_launch"] longValue] * 1000);
 
-    app.durationInForeground = @(activeTimeSinceLaunch);
-    app.duration = @(activeTimeSinceLaunch + backgroundTimeSinceLaunch);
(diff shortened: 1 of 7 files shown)


Source/BugsnagAppWithState.m
```diff
@@ -75,8 +75,8 @@ + (BugsnagAppWithState *)appWithDictionary:(NSDictionary *)event
     NSUInteger activeTimeSinceLaunch = (NSUInteger) ([stats[@"active_time_since_launch"] longValue] * 1000);
     NSUInteger backgroundTimeSinceLaunch = (NSUInteger) ([stats[@"background_time_since_launch"] longValue] * 1000);
```
tomlongridge, line 76: Neater to use `NSNumber.numberWithLong` here rather than box it later?

Source/BugsnagStackframe.m
```diff
@@ -71,17 +71,22 @@ - (NSDictionary *)toDictionary {
...
+        BSGDictSetSafeObject(dict, symbolAddr, BSGKeySymbolAddr);
+    }
+    if (self.machoLoadAddress) {
+        NSString *imageAddr = [NSString stringWithFormat:BSGKeyFrameAddrFormat, [self.machoLoadAddress unsignedLongValue]];
+        BSGDictSetSafeObject(dict, imageAddr, BSGKeyMachoLoadAddr);
+    }
+    if (self.machoVmAddress) {
+        NSString *vmAddr = [NSString stringWithFormat:BSGKeyFrameAddrFormat, [self.machoVmAddress unsignedLongValue]];
+        BSGDictSetSafeObject(dict, vmAddr, BSGKeyMachoVMAddress);
+    }
```
tomlongridge, line 89: Seems somewhat convoluted now to convert to long just so it can be formatted into a string - can we just get the string value and change the `BSGKeyFrameAddrFormat`?

fractalwrench: It seems a big long-winded but my understanding is that the API field relies on having hexadecimal addresses as a string. I'd be wary of changing the serialized response, unless I'm misunderstanding the review comment

PR #615, opened 2020-05-21
UAT: check App and Device fields

robinmacharg:
## Goal

e2e tests for App and Device fields.


features/app_and_device_attributes.feature
```diff
@@ -0,0 +1,45 @@
...
+    
+    # Device
+    
+    And the payload field "events.0.device.osName" equals "iOS"
+    And the payload field "events.0.device.jailbroken" is false
+    And the payload field "events.0.device.osVersion" matches the regex "\d+\.\d+"
+    And the payload field "events.0.device.manufacturer" equals "Apple"
+    And the payload field "events.0.device.locale" is not null
+    And the payload field "events.0.device.id" is not null
+    And the payload field "events.0.device.model" equals "iPhone10,4"
```
twometresteve, line 17: This will only work for certain iPhone models and indeed the Buildkite iOS 11 test fails because it's "iPhone10,1".  Having searched existing features files for the model field, step `the payload field "device.model" matches the test device model` looks like what you want.

features/app_and_device_attributes.feature
```diff
@@ -0,0 +1,45 @@
...
+    And the payload field "events.0.device.modelNumber" is not null
+    And the payload field "events.0.device.runtimeVersions.osBuild" is not null
+    And the payload field "events.0.device.runtimeVersions.clangVersion" is not null 
+    And the payload field "events.0.device.totalMemory" is an integer
+
+    # DeviceWithState
+
+    And the payload field "events.0.device.freeDisk" is an integer
+    And the payload field "events.0.device.freeMemory" is an integer
+    #And the payload field "events.0.device.orientation" equals "portrait"
```
twometresteve, line 27: Don't leave lines commented out.  If a check fails and it can't be easily resolved, remove it and raise an internal ticket to address the failure and reinstate the check.

features/app_and_device_attributes.feature
```diff
@@ -0,0 +1,45 @@
...
+    And the payload field "events.0.app.id" equals "com.bugsnag.iOSTestApp"
+    And the payload field "events.0.app.releaseStage" equals "development"
+    And the payload field "events.0.app.type" equals "iOS"
+    And the payload field "events.0.app.version" equals "1.0.3"
+    
+    # AppWithState
+    
+    And the payload field "events.0.app.duration" is a number
+    And the payload field "events.0.app.durationInForeground" is a number
+    And the payload field "events.0.app.inForeground" is not null
```
twometresteve, line 45: Is this not predictably "true" in this test?

features/app_and_device_attributes.feature
```diff
@@ -0,0 +1,45 @@
...
+    And the payload field "events.0.device.runtimeVersions.osBuild" is not null
+    And the payload field "events.0.device.runtimeVersions.clangVersion" is not null 
+    And the payload field "events.0.device.totalMemory" is an integer
+
+    # DeviceWithState
+
+    And the payload field "events.0.device.freeDisk" is an integer
+    And the payload field "events.0.device.freeMemory" is an integer
+    #And the payload field "events.0.device.orientation" equals "portrait"
+    And the payload field "events.0.device.time" is not null
```
twometresteve, line 28: Seems unnecessary given the next line.

features/app_and_device_attributes.feature
robinmacharg: When I initially tested this it came back false so the test was changed to a presence-confirmation.  Running it again I see it's true.  Will leave it as true until/unless it proves to be inconsistent.

twometresteve: I may well be inconsistent, to be fair.

features/app_and_device_attributes.feature
robinmacharg: Ticket raised, line left commented in for context, on the assumption that this needs addressed before UAT is considered complete.

features/steps/ios_steps.rb
```diff
@@ -101,6 +101,12 @@
   end
 end
 
+When("I bring the app to the foreground") do
```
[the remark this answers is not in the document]
robinmacharg: I've removed the new-session-behaviour test from the session_tracking feature that made use of this and created a ticket to re-add it later.

features/steps/v6uatsteps.rb
```diff
@@ -0,0 +1,53 @@
+require 'date'
```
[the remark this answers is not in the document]
twometresteve: What were your thoughts on those?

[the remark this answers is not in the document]
robinmacharg: Re #609 comments: 
1) Thanks!
2) Date comparison/parsing is addressed.  The comment I made there: "The timestamp matchers either target header or query parameters, or test numeric timestamps, AFAICT. There is no step that explicitly matches an iso datetime for a request field. I'd say ISO8601 is reasonably standard and unlikely to change. I agree that guarding the parse is a good idea."
3) Negations ("is not an X") have been removed.

features/steps/v6uatsteps.rb
```diff
@@ -0,0 +1,30 @@
...
+
+# Number
+
+Then(/^the payload field "(.+)" is a number$/) do |field_path|
+    value = read_key_path(Server.current_request[:body], field_path)
+    assert_kind_of Numeric, value
```
twometresteve, line 7: The change to `assert_kind_of` should have been applied to the other blocks as well, but I can take care of that when I move it into MR.

robinmacharg: NP, I've made the change.

PR #620, opened 2020-05-22
Add E2E scenarios for metadata redaction

fractalwrench:
## Goal

Adds mazerunner scenarios to verify that `event.metadata` values can be redacted by a string or regex set in `config.redactedKeys`. Verified the following:

- Values with key 'password' are removed by default
- Redaction takes place after an `OnError` callback
- Nested values are redacted


twometresteve:
LGTM and relevant test results in Buildkite checked.

PR #621, opened 2020-05-22
Add E2E tests for enabledReleaseStages

fractalwrench:
## Goal

Adds E2E tests to verify that `config.enabledReleaseStages` controls whether errors/sessions are delivered depending on the value of `config.releaseStage`. If the release stage is not present in `config.enabledReleaseStages` no errors or sessions should be sent.

## Changeset

- Fixed a bug in the session tracking implementation where session capture did not respect the `config.enabledReleaseStages` property
- Added E2E tests to verify that both manual and automatic sessions respect `enabledReleaseStages`
- Renamed scenarios in existing Event tests to use the `enabledReleaseStage` terminology rather than `notifyReleaseStage`
- Removed seemingly unnecessary wait from existing release stages scenarios


Source/BugsnagSessionTracker.m
```diff
@@ -138,6 +138,10 @@ - (void)startNewSessionIfAutoCaptureEnabled {
 }
 
 - (void)startNewSessionWithAutoCaptureValue:(BOOL)isAutoCaptured {
+    NSArray *releaseStages = self.config.enabledReleaseStages;
+    if (releaseStages != nil && ![releaseStages containsObject:self.config.releaseStage]) {
+        return;
+    }
```
twometresteve, line 144: I'm inclined to say that any code changes should be on a separate PR, as we could easily miss an explanation for changed behaviour (when the title just says "Add tests").  I'd also want to make sure it receives sufficient attention from a Cocoa expert.  E.g. are there cases we still need to consider here (`releaseStages` being `nil`, perhaps)?

fractalwrench:
Split changes to the code out into a separate PR which will be reviewed independently.

PR #622, opened 2020-05-26
Add E2E tests for verifying user info

fractalwrench:
## Goal

Adds E2E tests to verify that user information is added to errors/sessions correctly.

The addition of E2E tests revealed a bug in the session implementation which have also been addressed here.

## Changeset

- `BugsnagSessionTracker` now uses the user from `client` rather than `config` as this is the one which is updated after initialisation
- Added E2E tests for verifying event/session callbacks can override the user
- Removed some user test coverage from unit tests as this is now covered by E2E tests
- Added E2E tests for verifying the user can be set from config/client for both events and sessions

### DIFF
--- a/features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/UserFromConfigEventScenario.swift
+++ b/features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/UserFromConfigEventScenario.swift
@@ -1,0 +1,32 @@
+//
+//  UserFromConfigEventScenario.swift
+//  iOSTestApp
+//
+//  Created by Jamie Lynch on 26/05/2020.
+//  Copyright © 2020 Bugsnag. All rights reserved.
+//
+
+import Foundation
+import Bugsnag
+
+/**
+ * Sends an event  to Bugsnag which contains a user set from Configuration
+ */
+internal class UserFromConfigEventScenario: Scenario {
+
+    override func startBugsnag() {
+        self.config.autoTrackSessions = false;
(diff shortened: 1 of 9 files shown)


features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/UserFromConfigEventScenario.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    override func startBugsnag() {
+        self.config.autoTrackSessions = false;
+        self.config.setUser("abc", withEmail: "fake@gmail.com", andName: "Fay K")
+        super.startBugsnag()
+    }
+
+    override func run() {
+        let user = Bugsnag.user()
+        Bugsnag.addMetadata(user.id, key: "id", section: "clientUserValue")
```
bengourley, line 25: It's unclear to me what this is doing and could maybe do with a comment?
My best guess is that it is taking the value from `Bugsnag.user()` and putting it in metadata so you can assert that the correct values were set by passing in config. Is that right?

fractalwrench: There is a default user information which has been temporarily removed while working around an issue with the buildkite pipeline - a ticket exists to enable this again once that issue has been resolved.

This line is achieving the following test requirement, I'll add a comment to clarify this for future readers:

>Setting the user in the Configuration and verifying its presence using getUser on the Client, Event, and Session (Suggest acquiring the user and setting metadata to the same value e.g. receivedClientUser)

PR #624, opened 2020-05-26
Add E2E scenarios for unhandled signals

fractalwrench:
## Goal

Adds E2E scenarios to verify that an error is captured for each different signal that Bugsnag sets a signal handler for. This changeset additionally moves existing unhandled scenarios from `crashprobe.feature` into a separate file to make test tracing more obvious.

## Changeset

- Added E2E scenarios to verify an error report is captured for all the signals Bugsnag sets a handler for
- Moved existing scenarios for NSException, mach exception, and C++ exception

### DIFF
--- a/features/unhandled_mach_exception.feature
+++ b/features/unhandled_mach_exception.feature
@@ -1,0 +1,7 @@
+Feature: Bugsnag captures an unhandled mach exception
+
+Scenario: Trigger a mach exception
+    When I run "UnhandledMachExceptionScenario" and relaunch the app
+    And I configure Bugsnag for "UnhandledMachExceptionScenario"
+    And I wait to receive a request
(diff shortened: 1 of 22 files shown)


features/unhandled_mach_exception.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Bugsnag captures an unhandled mach exception
+
+Scenario: Trigger a mach exception
+    When I run "UnhandledMachExceptionScenario" and relaunch the app
+    And I configure Bugsnag for "UnhandledMachExceptionScenario"
+    And I wait to receive a request
```
bengourley, line 6: is an assertion you could make that would confirm that this the expected mach exception, rather than some other error?

such as the error class matching `UnhandledMachExceptionScenario`?

fractalwrench: Updated with an extra assertion to confirm that the error message matches the scenario.

PR #632, opened 2020-05-27
Add tests to check the default config values

imjoehaines:
## Goal

Currently we're only testing the default values for a couple of configuration options

This PR adds a unit test for each configuration option, ensuring the default value is what we say it is

### DIFF
--- a/Tests/BugsnagConfigurationTests.m
+++ b/Tests/BugsnagConfigurationTests.m
@@ -294,15 +294,6 @@
     XCTAssertFalse([config shouldSendReports]);
 }
 
-- (void)testDefaultReleaseStage {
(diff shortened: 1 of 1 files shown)


Tests/BugsnagConfigurationTests.m
```diff
@@ -583,6 +574,113 @@ - (void)testMaxBreadcrumb {
...
+#else
+    XCTAssertEqualObjects(@"production", config.releaseStage);
+#endif
+}
+
+- (void)testDefaultSendThreads {
+    BugsnagConfiguration *config = [[BugsnagConfiguration alloc] initWithApiKey:DUMMY_APIKEY_32CHAR_1];
+
+    XCTAssertEqual(BSGThreadSendPolicyAlways, config.sendThreads);
+}
```
robinmacharg: I think most - if not all - of these can be lumped into a single test.  There shouldn't be any interaction between values and modifying them should be side-effect free.

PR #634, opened 2020-05-27
Address bug in enabledBreadcrumbTypes record logic

fractalwrench:
## Goal

If `config.enabledBreadcrumbTypes` was set to `BSGEnabledBreadcrumbTypeNone`, the `shouldRecordBreadcrumbType` method returned true for all breadcrumb types. This changeset makes it return false in this condition and adds unit test coverage to verify the change.

E2E tests added in #635 depend on this change.

### DIFF
--- a/Source/BugsnagConfiguration.m
+++ b/Source/BugsnagConfiguration.m
@@ -475,8 +475,8 @@ NSString * const kBugsnagUserUserId = @"BugsnagUserUserId";
  */
 - (BOOL)shouldRecordBreadcrumbType:(BSGBreadcrumbType)type {
     // enabledBreadcrumbTypes is BSGEnabledBreadcrumbTypeNone
-    if (!self.enabledBreadcrumbTypes) {
-        return YES;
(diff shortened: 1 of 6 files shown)


Source/BugsnagConfiguration.m
```diff
@@ -475,8 +475,8 @@ - (void)setMaxBreadcrumbs:(NSUInteger)capacity {
  */
 - (BOOL)shouldRecordBreadcrumbType:(BSGBreadcrumbType)type {
     // enabledBreadcrumbTypes is BSGEnabledBreadcrumbTypeNone
-    if (!self.enabledBreadcrumbTypes) {
-        return YES;
+    if (self.enabledBreadcrumbTypes == 0 && type != BSGBreadcrumbTypeManual) {
```
robinmacharg, line 478: A comparison to `BSGEnabledBreadcrumbTypeNone` would be more explicit here.

fractalwrench: Updated comparison to use `BSGEnabledBreadcrumbTypeNone`

PR #645, opened 2020-05-28
Validate file arg

twometresteve:
Check that the feature file given exists before attempting to run Maze Runner.


twometresteve:
> Happy enough with this as a short-term fix - it'd be better to address this in the bugsnag-mazerunner repository if possible so that it applies to all the repos.

I'm sure we can do something better than what we currently have, it just wasn't totally trivial to do in five minutes.  W'll look at the internal ticket you raised in due course though, for sure.

PR #646, opened 2020-05-29
Assert the API Key is not nil or empty on init and accept otherwise invalid API keys

imjoehaines:
## Goal

Currently we convert `nil` and API keys that aren't 32 hex characters to an empty string

The expected behaviour is to reject `nil` and empty string API keys outright and to warn about API keys that aren't 32 hex characters but use them anyway

## Changeset

- `setApiKey` now checks for a missing key (`nil` or empty) and throws
- `setApiKey` then checks if it's a valid 32 hex char string and warns if not but still uses the given string
- `initWithApiKey` now calls `setApiKey` rather than setting it itself, so it inherits the above logic
- `isValidApiKey` doesn't throw itself, so that it's always safe to call when we check the validity of the key

## Tests

- Existing tests that pass a now-invalid key have been changed to assert it throws
- Existing tests that pass a 16 character string have been updated to reflect that this now uses the given string, instead of throwing or using an empty string (depending on if `setApiKey` or `initWithApiKey` is used)

### DIFF
--- a/Source/BugsnagConfiguration.m
+++ b/Source/BugsnagConfiguration.m
@@ -48,7 +48,7 @@
 static NSString *const kHeaderApiPayloadVersion = @"Bugsnag-Payload-Version";
 static NSString *const kHeaderApiKey = @"Bugsnag-Api-Key";
 static NSString *const kHeaderApiSentAt = @"Bugsnag-Sent-At";
-static NSString *const BSGApiKeyError = @"apiKey must be a 32-digit hexadecimal value.";
+static NSString *const BSGApiKeyMissingError = @"No Bugsnag API Key set";
 static NSString *const BSGInitError = @"Init is unavailable.  Use [[BugsnagConfiguration alloc] initWithApiKey:] instead.";
 static const int BSGApiKeyLength = 32;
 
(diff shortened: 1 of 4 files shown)


Tests/Swift Tests/BugsnagSwiftPublicAPITests.swift
```diff
@@ -46,7 +46,6 @@ class BugsnagSwiftPublicAPITests: XCTestCase {
     let onBreadcrumbBlock: BugsnagOnBreadcrumbBlock = { (breadcrumb) -> Bool in return false }
     
     func testBugsnagClass() throws {
-        Bugsnag.start(withApiKey: "")
```
imjoehaines, line 49: This had to go as it now throws an Objective C exception, which can't be caught here :(

Tests/Swift Tests/BugsnagSwiftPublicAPITests.swift
```diff
@@ -89,9 +88,8 @@ class BugsnagSwiftPublicAPITests: XCTestCase {
         }
         Bugsnag.removeOnBreadcrumb(block: onBreadcrumbBlock)
     }
-    
+
     func testBugsnagConfigurationClass() throws {
-        let _ = BugsnagConfiguration.loadConfig()
```
imjoehaines, line 94: Same as above — there's no API key to load here so it throws too

Source/BugsnagConfiguration.m
```diff
@@ -170,13 +170,28 @@ - (nonnull id)copyWithZone:(nullable NSZone *)zone {
...
     return isHex && [apiKey length] == BSGApiKeyLength;
 }
 
+/**
+ * Check if the API Key is missing, i.e. it's nil, not a string or is empty
+ *
+ * This is distinct from 'isValidApiKey' to allow throwing an exception in this case, but not
+ * if the API Key is in the wrong format
+ */
++ (BOOL)isApiKeyMissing:(NSString *)apiKey {
```
robinmacharg, line 191: This method would benefit from @param/@return documentation.  Also, is the `isKindOfClass` comparison necessary?  Users are warned about incorrect assignment and incorrect type assignment is not possible in Swift.  I wonder if it's overkill.

imjoehaines: The `[apiKey length]` call caused one or two tests to throw an exception because they passed the wrong type, when they were expecting `isValidApiKey` to return false safely. This just moved from where `isHex` is set rather than being a new check, but I can remove it if you'd prefer it to throw when given an invalid type — `setApiKey` will throw either way, but `isValidApiKey` is used in a few other places too

robinmacharg: I went through the rest of the code and revised my original opinion on type checking; I think there is a case for it.  I edited the comment, I suspect while you were replying.  The doc nit still stands.

PR #647, opened 2020-05-29
Fix enabled error type test flake

fractalwrench:
## Goal

Fixes a flaky E2E test that checks enabled error types.

The root cause of this was that the `EnabledErrorTypesCxxScenario` creates a handled error, then immediately creates an unhandled error. This leads to a race where the handled error is delivered on a background thread and might not be sent to the mock API before the unhandled error triggers.

The fix for this is to delay the unhandled error for 2 seconds - note that is potentially applicable to other flaky scenarios which have been written in the same way. Verified by running this scenario 20 times on Buildkite without any failures.

### DIFF
--- a/features/enabled_error_types.feature
+++ b/features/enabled_error_types.feature
@@ -27,15 +27,13 @@ Scenario: NSException Crash Reporting is disabled
     And the event "unhandled" is false
     And the payload field "events.0.exceptions.0.message" equals "DisableNSExceptionScenario - Handled"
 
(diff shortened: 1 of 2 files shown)


twometresteve:
The failing we were seeing was an unhandled error being reported, which we wouldn't expect at all.  I'd be concerned that we are now passing the test having received the handled error, with the test harness closing down before it gets to chance to see an erroneous request (but I'd need to check the behaviour of the test harness there).

fractalwrench:
Converting to draft as requested for now

PR #651, opened 2020-05-31
Test fixture enhancements

twometresteve:
## Goal

Adds a button to the test fixture allowing automated tests to clear user data between scenarios (to following in a subsequent PR).

Also provides the ability to run the test fixture in "Manual mode".  If an API key is provided, the fixture will run scenarios in the same way, but send any generated reports to the Bugsnag dashboard.

## Tests

Passing CI run verifies that the change does not affect existing automated tests.  New API key field tested manually together with basic testing of the clear data button, although full testing of that will follow in the subsequent PR that integrates its use into automated tests.


twometresteve:
> LGTM.
> 
> Before merging is it worth adding the function/step to the features file to use the reset? (Even if it's not present in the scenarios yet)

If I wasn't taking a day's leave I would carry it on an problem finish the whole lot, but as it is I'd like to get it is as it may be useful in the interim.

PR #654, opened 2020-06-01
Rationalise preprocessor macros into header file

fractalwrench:
## Goal

Moves preprocessor macros into a new header file `BugsnagPlatformConditional`. This ensures that macros used to control behaviour by the target (iOS/OSX/tvOS) are used consistently and defined in one place.

Much of this platform-specific code should be split out into platform-specific files at a future date - an additional ticket has been raised to address this.

## Changeset

- Added `BugsnagPlatformConditional` header which defines three macros to determine which platform is in use
- Imported Foundation header in `BugsnagPlatformConditional` to ensure that the target is defined properly
- Removed use of `TARGET_OS_*` macros within codebase to prefer `BugsnagPlatformConditional`, which should always be imported first


Source/KSCrash/Source/KSCrash/Recording/BSG_KSSystemInfo.h
```diff
@@ -57,7 +57,9 @@
 #define BSG_KSSystemField_BuildType "build_type"
 
 #import <Foundation/Foundation.h>
```
tomlongridge, line 59: Should we remove the `Foundation/Foundation.h` include?

Source/BugsnagClient.m
```diff
@@ -196,7 +198,9 @@ void BSSerializeDataCrashHandler(const BSG_KSCrashReportWriter *writer, int type
  *
  * @returns A string representing the device orientation or nil if there's no equivalent
  */
-#if TARGET_OS_IOS
+
+#if BSG_PLATFORM_TVOS
```
tomlongridge, line 202: Why the empty test for tvOS? If it's to show that iOS only is intentional, we should probably also include an empty macOS block.

Source/BugsnagPlatformConditional.h
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Defined as true if this is the iOS platform.
+ */
+#define BSG_PLATFORM_IOS TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
+
+/**
+* Defined as true if this is the OSX platform.
+*/
+#define BSG_PLATFORM_OSX TARGET_OS_MAC
```
tomlongridge, line 26: I believe we should use `TARGET_OS_OSX` here: http://sealiesoftware.com/blog/archive/2010/8/16/TargetConditionalsh.html

Source/BugsnagPlatformConditional.h
```diff
@@ -0,0 +1,33 @@
...
+
+#ifndef BugsnagPlatformConditional_h
+#define BugsnagPlatformConditional_h
+
+#import <Foundation/Foundation.h>
+
+/**
+ * Defined as true if this is the iOS platform.
+ */
+#define BSG_PLATFORM_IOS TARGET_IPHONE_SIMULATOR || TARGET_OS_IPHONE
```
tomlongridge, line 21: I don't believe we should be using `TARGET_IPHONE_SIMULATOR` unless we're trying to determine the difference between a real device and a simulator. This will be true for mobile, TV or watch simulators.

Source/BugsnagClient.m
```diff
@@ -625,7 +630,7 @@ - (void)addTerminationObserver:(NSString *)name {
 
 - (void)computeDidCrashLastLaunch {
     const BSG_KSCrash_State *crashState = bsg_kscrashstate_currentState();
-#if TARGET_OS_TV || TARGET_OS_IPHONE
```
tomlongridge, line 628: Deliberately not simulator?

Source/BSGConnectivity.m
```diff
@@ -41,7 +43,7 @@
  * @return YES if the connectivity change should be reported
  */
 BOOL BSGConnectivityShouldReportChange(SCNetworkReachabilityFlags flags) {
-    #if TARGET_OS_TV || TARGET_OS_IPHONE
```
tomlongridge, line 44: Deliberately not simulator?

Source/BSGConnectivity.m
```diff
@@ -69,7 +71,7 @@ BOOL BSGConnectivityShouldReportChange(SCNetworkReachabilityFlags flags) {
  */
 NSString *BSGConnectivityFlagRepresentation(SCNetworkReachabilityFlags flags) {
     BOOL connected = (flags & kSCNetworkReachabilityFlagsReachable);
-    #if TARGET_OS_TV || TARGET_OS_IPHONE
```
tomlongridge, line 72: Deliberately not simulator?

Source/BugsnagPlatformConditional.h
tomlongridge, line 21: Also, should we use `TARGET_OS_IOS`?

Source/KSCrash/Source/KSCrash/Recording/BSG_KSSystemInfo.h
fractalwrench: This is an import which means Obj-C will guard against the header being included more than once. It's also more explicit that `BSG_KSSystemInfo` requires this header and guards against the internals of `BugsnagPlatformConditional` removing `Foundation.h` at some later date and then having to add `Foundation.h` back in again.

Source/BSGConnectivity.m
fractalwrench: I can't see any reason why this code shouldn't apply to iOS/tvOS simulators as the `kSCNetworkReachabilityFlagsIsWWAN` flag is only missing on macOS.

Source/BugsnagPlatformConditional.h
fractalwrench: Updated to use `TARGET_OS_OSX`

Source/BugsnagPlatformConditional.h
fractalwrench: changed the define to use `TARGET_OS_IOS`

Source/BugsnagClient.m
fractalwrench: Removed change here and in `BSGOutOfMemoryWatchdog` as this appears to be deliberate

Source/BugsnagClient.m
fractalwrench: This was required when `BSG_PLATFORM_IOS` was defined via the simulator macro - this is no longer required so has been amended.

PR #655, opened 2020-06-01
Update V6 objective-c-ios example application

Cawllec:
## Goal
 Updates the `objective-c-ios` example app to using the V6 syntax and functionality.  Covers a broad range of common functionality and explains some of the Bugsnag usage. 

For swift I'm looking to basically reproduce this using Swift as an OSX app, so we'll cover both iOS and OSX and both languages.  However merging this may be useful for manual testing in the short-term.


examples/objective-c-ios/Bugsnag Test App/ViewController.m
```diff
@@ -16,24 +16,33 @@ @implementation ViewController
...
+ */
+- (IBAction)crashMemoryPressure:(id)sender {
+    OutOfMemoryController *controller = [OutOfMemoryController new];
     [self.navigationController pushViewController:controller animated:YES];
 }
 
-- (IBAction)generateException:(id)sender {
-    [Bugsnag leaveBreadcrumbWithMessage:@"Generating exception by non-existent selector"];
+/**
+ This generates an exception that won't be caught by any error handling.  If Bugsnag is set up correctly upon reopening the app the error will be reported to your Bugsnag dashboard.
```
twometresteve, line 37: nit

```suggestion
 This generates an exception that won't be caught by any error handling.  If Bugsnag is set up correctly, upon reopening the app the error will be reported to your Bugsnag dashboard.
```

examples/objective-c-ios/Bugsnag Test App/AppDelegate.m
```diff
@@ -13,9 +13,51 @@ @implementation AppDelegate
...
     
+    /**
+     This is the minimum amount of setup required for Bugsnag to work.  Simply input your apiKey and the application will deliver all error and session notifications to the appropriate dashboard.
+     */
     NSString *apiKey = @"<YOUR_APIKEY_HERE>";
-    BugsnagConfiguration *config = [[BugsnagConfiguration alloc] initWithApiKey:apiKey];
-    [Bugsnag startWithConfiguration:config];
+    [Bugsnag startWithApiKey:apiKey];
+    
+    /**
```
twometresteve, line 22: I wonder if we are making a rod for our own backs here by duplicating information that is available in the docs.  I personally find that commented out code can be distracting and would rather copy/paste it for myself from the docs as required.  It's also one less thing for us to maintain should config properties change over time.

PR #666, opened 2020-06-02
Avoid calling C++ termination handler for NSException

fractalwrench:
## Goal

Avoids calling the original C++ termination handler for uncaught `NSException` objects. If `enabledErrorType.nsexceptions` is set to false, then the `BSG_KSCrashSentry_NSException` handler will not be installed, and any uncaught `NSException` will propagate through to the C++ handler instead. Calling the original handler ultimately results in a SIGABRT with the bugsnag method as the error context:

Screenshot 2020-06-02 at 15 38 31

## Changeset

This changeset only calls the original termination handler for C++ exceptions that are not NSException, thus allowing for the correct context and other information to be gathered on the SIGABRT.

Verified by testing an NSException before and after the fix, with NSException capture both enabled/disabled.


fractalwrench:
>Can you confirm that tests exist that verify this behaviour?

This specific scenario has been manually tested - in addition there are E2E tests to confirm that NSException + C++ exception are captured correctly.

PR #669, opened 2020-06-03
Deliver each event in a separate request to avoid exceeding payload size limit

fractalwrench:
## Goal

The payload size limit can be exceeded if multiple events are batched together into one request, meaning events are dropped by our API. This changeset removes the batching functionality and sends each event in a separate request to reduce the likelihood of this occurring. It is based off the obsolete #424

## Changeset

- Sent each event in a separate request within `sendAllReportsWithBlock`.
- Logged the filename when a request succeeds/fails
- Passed through 1 event at a time to `BugsnagApiClient` for errors (sessions are still batched together in multiple payloads)
- Changed completion callback to send the filename rather than the count of reports sent
- Removed `deleteBehaviorAfterSendAll` from `BSG_KSCrash` as the property is not exposed externally and the behaviour is not resepected
- Updated KSCrash to always assume that `BugsnagErrorReportSink` should be called when delivering error reports, rather than a `BSG_KSCrashReportFilter`. Deleted `BSG_KSCrashReportFilter` as the error sink is the only place this is used.
- Renamed 'completion' to 'block' for general consistency, raised in previous PR review comments
- Reworked file deletion to allow deleting one file at a time, rather than all in a batch

## Tests

- Updated existing unit test coverage to work with changeset
- Updated mazerunner scenario assertions where they assumed requests were batched


Bugsnag/BugsnagErrorReportSink.m
```diff
@@ -0,0 +1,180 @@
...
+        if ([event shouldBeSent] && [self runOnSendBlocks:configuration event:event]) {
+            storedEvents[fileKey] = event;
+        } else { // delete the report as the user has discarded it
+            [self finishActiveRequest:fileKey success:true error:nil block:block];
+        }
+    }
+    [self deliverStoredEvents:storedEvents configuration:configuration block:block];
+}
+
+- (void)deliverStoredEvents:(NSMutableDictionary<NSString *, BugsnagEvent *> *)storedEvents
```
tomlongridge, line 134: We discuss IRL whether this should be done in separate phases - can you comment the function with some rationale?

Bugsnag/Delivery/BugsnagErrorReportApiClient.m
```diff
@@ -30,16 +30,16 @@ - (void)main {
...
                             BOOL completed, NSError *error) {
                         if (error) {
-                            bsg_log_warn(@"Failed to send reports: %@", error);
-                        } else if (sentReportCount > 0) {
-                            bsg_log_info(@"Reports sent.");
+                            bsg_log_warn(@"Failed to send error report: %@ %@", filename, error);
+                        } else if (filename) {
+                            bsg_log_info(@"Delivered error report '%@'", filename);
                         }
                     }];
```
tomlongridge, line 40: Do we need this block? It's just logging the outcome which is also done by `BSG_KSCrash.sendAllReportsWithBlock`

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
```diff
@@ -288,33 +287,26 @@ - (void)listenForLoadedBinaries {
...
-
-- (void)deleteAllReports {
-    [self.crashReportStore deleteAllFiles];
+            withBlock:^(NSString *filename, BOOL completed,
+                    NSError *error) {
+                BSG_KSLOG_DEBUG(@"Process finished with completion: %d", completed);
+                if (error != nil) {
+                    BSG_KSLOG_ERROR(@"Failed to send reports: %@", error);
+                }
+                if (completed) {
```
tomlongridge, line 304: && filename != nil?

CHANGELOG.md
```diff
@@ -463,6 +463,9 @@ names are deprecated but still supported until the next major release.
 This release disables reporting out-of-memory events in debug mode, to reduce
 false positives.
 
+* Deliver each event in a separate request to avoid exceeding payload size limit
```
twometresteve, line 466: Should there also be an entry in the v6 section for this PR?

features/steps/ios_steps.rb
```diff
@@ -87,25 +87,54 @@
...
+  assert_true(request_fields_are_equal(key, index_a, index_b))
+end
+
+Then("the payload field {string} is not equal for request {int} and request {int}") do |key, index_a, index_b|
+  assert_false(request_fields_are_equal(key, index_a, index_b))
+end
+
+def request_fields_are_equal(key, index_a, index_b)
+  requests = Server.stored_requests.to_a
+  request_a = requests[index_a][:body]
```
twometresteve, line 133: These will fall over if enough requests haven't been received.  Consider adding something like:
```
assert_true(requests.length > index_a, "Not enough requests received to access index #{index_a}")
assert_true(requests.length > index_b, "Not enough requests received to access index #{index_b}")
```

Bugsnag/BugsnagErrorReportSink.m
```diff
@@ -0,0 +1,180 @@
...
+              configuration:(BugsnagConfiguration *)configuration
+                      block:(BSGOnErrorSentBlock)block {
+    for (NSString *filename in storedEvents) {
+        BugsnagEvent *event = storedEvents[filename];
+        event.redactedKeys = configuration.redactedKeys;
```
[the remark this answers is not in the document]
fractalwrench: Moved next to `codeBundleId` assignment

Bugsnag/Delivery/BugsnagErrorReportApiClient.m
fractalwrench: This doesn't seem to be necessary so I've removed it. In addition this means `BSG_KSCrashCallCompletion` is no longer required so I've removed these files from the project.

Bugsnag/BugsnagErrorReportSink.m
```diff
@@ -0,0 +1,180 @@
...
+    // 3. If so, ignore them and perform no action
+    // 4. When a request has completed and deleted the file, remove the files from the dictionary
+    NSArray<NSString *> *keys = [self prepareNewRequests:[ksCrashReports allKeys]];
+    NSMutableDictionary<NSString *, BugsnagEvent *>* storedEvents = [NSMutableDictionary new];
+    BugsnagConfiguration *configuration = [Bugsnag configuration];
+
+    for (NSString *fileKey in keys) {
+        NSDictionary *report = ksCrashReports[fileKey];
+        BugsnagEvent *event = [[BugsnagEvent alloc] initWithKSReport:report];
+        event.codeBundleId = [Bugsnag client].codeBundleId;
```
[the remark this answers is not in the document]
fractalwrench: This should not be required as the property holds a reference to an immutable string object

CHANGELOG.md
fractalwrench: Added a note to the upgrade guide and moved this to v6 - the original PR was targeted at v5.

features/steps/ios_steps.rb
```diff
@@ -87,25 +87,54 @@
...
+  request_count = requests.count()
+  match_count = 0
+
+  # iterate through each row in the table. exactly 1 request should match each row.
+  table.hashes.each do |row|
+    requests.each do |request|
+      event = read_key_path(request[:body], "events.0")
+
+      if request_matches_row(event, row)
+        puts("Found a match")
```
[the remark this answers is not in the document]
fractalwrench: Updated to use one line.

features/steps/ios_steps.rb
fractalwrench: Updated to handle the case where not enough requests are received

features/steps/ios_steps.rb
```diff
@@ -87,25 +87,54 @@
...
   end
+  assert_equal(request_count, match_count, "Unexpected number of requests matched the received payloads")
+end
+
+def request_matches_row(body, row)
+  request_matches = true
+
+  row.all? do |key, expected_value|
+    obs_val = read_key_path(body, key)
+    equal_values = expected_value.to_s.eql? obs_val.to_s
```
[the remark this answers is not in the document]
fractalwrench: Addressed by nil checking here.

features/steps/ios_steps.rb
```diff
@@ -87,25 +87,59 @@
...
+  assert_equal(request_count, match_count, "Unexpected number of requests matched the received payloads")
+end
+
+def request_matches_row(body, row)
+  request_matches = true
+
+  row.all? do |key, expected_value|
+    obs_val = read_key_path(body, key)
+    match = false
+    if obs_val.nil?
```
twometresteve, line 112: I don't think this is right - expected will always match if observed is nil, even if it isn't itself.  I'm looking at the tests myself as well now though, so can look at handling this.

PR #670, opened 2020-06-03
V6 rebuild project

robinmacharg:
## Goal

[Previous work to combine three separate Xcode projects](https://github.com/bugsnag/bugsnag-cocoa/pull/633) didn't go far enough and left a folder structure that didn't follow Apple best-practice and, further, wasn't amenable to consistent install instructions.  This work rebuilds the entire project structure to address these issues.

## Design

- A new Xcode iOS framework project was created.
- Framework targets for each platform were added, along with tests and a static target.
- Build settings from the previous project merge were copied across
- Source files were moved from existing locations to preserve history, with additional directory structure added.  Tests were also re-added.  A single redundant Swift test was elided.

Commits were granular and show the progression towards a stable state.  Pushes were frequent to ensure that CI issues were addressed.  Builds were attempted frequently locally. Popular open source frameworks were consulted to ensure best (common) practice was followed as much as possible.

## Changeset

- Project file added at top level, source files moved; this generates a large changeset that may be better reviewed as a checkout.  v6 changes were merged in to ensure that the two branches did not diverge too greatly.
- Makefile adjusted for new source path

## Tests

Manual install, Cocoapods, Carthage both local and remote.


fractalwrench:
Updated PR to address the review points raised above - I've tested out on an example app using Cocoapods only. @tomlongridge would be good to get another pair of eyes reviewing this.

fractalwrench:
`Configurations/Config.xcconfig` does not appear to be necessary - I've deleted this and rebased the PR to resolve conflicts from the deadlock fix merged into `v6`.

PR #674, opened 2020-06-04
fix: the DYLD cache locking mechanism now determines the lock mechani…

robinmacharg:
## Goal

The previous fix for a DYLD deadlock issue has compile-time checks that determine the locking mechanism to use.  This can cause apps with low deployment targets to fail IRL due to being built and tested on more modern OSs.

## Design

The previous lock determination mechanism was purely compile-time `#ifdef`-based.  This fix uses a combination of checks: 

- Compile-time to determine the OS-type without resorting to weakly supported hybrid checks
- Run-time to determine the version of the OS in a cross-platform way.

The bulk of the logic was moved from setting `#define`s in the header to the combination approach in the main `.m` file.

The fix was targeted at v6 due to ongoing UAT but could also easily be back-ported to the v5 release.

## Changeset

`BSG_KSMachHeaders` and `BSG_KSCrash`

## Tests

Run on both real iOS 9.3 iPad and in v13.X simulator.

### DIFF
--- a/Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
+++ b/Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
@@ -11,6 +11,118 @@
 #import <Foundation/Foundation.h>
 #import "BSG_KSDynamicLinker.h"
 #import "BSG_KSMachHeaders.h"
+#import "BugsnagPlatformConditional.h"
+
+// MARK: - Locking
+
+// Pragma's hide unavoidable (and expected) deprecation/unavailable warnings
+_Pragma("clang diagnostic push")
+_Pragma("clang diagnostic ignored \"-Wunguarded-availability\"")
+static os_unfair_lock bsg_mach_binary_images_access_lock_unfair = OS_UNFAIR_LOCK_INIT;
+_Pragma("clang diagnostic pop")
+
+_Pragma("clang diagnostic push")
+_Pragma("clang diagnostic ignored \"-Wdeprecated-declarations\"")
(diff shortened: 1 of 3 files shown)


Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -11,6 +11,118 @@
...
+void bsg_unfair_unlock() {
+    _Pragma("clang diagnostic push")
+    _Pragma("clang diagnostic ignored \"-Wunguarded-availability\"")
+    os_unfair_lock_unlock(&bsg_mach_binary_images_access_lock_unfair);
+    _Pragma("clang diagnostic pop")
+}
+
+// Lock and unlock sections of code
+
+void bsg_dyld_cache_lock() {
```
fractalwrench, line 64: This function is a candidate for the use of [isOperatingSystemAtLeastVersion](https://developer.apple.com/documentation/foundation/nsprocessinfo/1414876-isoperatingsystematleastversion?language=objc), which would simplify the code and minimise the number of lock calls/return statements:

```objc
- (BOOL)spinLockSupported {
    NSOperatingSystemVersion minSdk;

    #if BSG_PLATFORM_IOS
        minSdk.majorVersion = 10;
        minSdk.minorVersion = 12;
        minSdk.minorVersion = 0;
    #elif
        // ... handle other platforms
    #endif
    return [NSProcessInfo isOperatingSystemAtLeastVersion:minSdk];
}

void bsg_dyld_cache_lock() {
    if (spinLockSupported) {
        bsg_spin_lock();
    } else {
        bsg_unfair_lock();
    }
}
```

fractalwrench:
@tomlongridge I've updated the PR with changes from my review feedback. Would be good to get another review + some testing on a real device that uses iOS <10.

PR #675, opened 2020-06-04
Fix DYLD lock mechanism preventing compilation on iOS <10

fractalwrench:
This changeset is a port of #674 for bugsnag-cocoa v5 - original PR description copied below:

## Goal

The previous fix for a DYLD deadlock issue has compile-time checks that determine the locking mechanism to use.  This can cause apps with low deployment targets to fail IRL due to being built and tested on more modern OSs.

## Design

The previous lock determination mechanism was purely compile-time `#ifdef`-based.  This fix uses a combination of checks: 

- Compile-time to determine the OS-type without resorting to weakly supported hybrid checks
- Run-time to determine the version of the OS in a cross-platform way.

The bulk of the logic was moved from setting `#define`s in the header to the combination approach in the main `.m` file.

The fix was targeted at v6 due to ongoing UAT but could also easily be back-ported to the v5 release.

## Changeset

`BSG_KSMachHeaders` and `BSG_KSCrash`

## Tests

Run on both real iOS 9.3 iPad and in v13.X simulator.


Source/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -12,6 +12,88 @@
...
+
+void bsg_dyld_cache_unlock() {
+    if (bsg_unfair_lock_supported) {
+        bsg_unfair_unlock();
+    } else {
+        bsg_spin_unlock();
+    }
+}
+
+BOOL IsUnfairLockSupported(NSProcessInfo *processInfo) {
```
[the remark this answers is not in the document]
fractalwrench: Updated to include `BSG` prefix.

Bugsnag.podspec.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Bugsnag",
-  "version": "5.23.2",
+  "version": "5.22.3",
```
twometresteve, line 3: Shouldn't these be `5.23.3`?
```suggestion
  "version": "5.23.3",
```

Bugsnag.podspec.json
```diff
@@ -9,7 +9,7 @@
   },
   "source": {
     "git": "https://github.com/bugsnag/bugsnag-cocoa.git",
-    "tag": "v5.23.2"
+    "tag": "v5.22.3"
```
twometresteve, line 12:
```suggestion
    "tag": "v5.23.3"
```

Source/BugsnagNotifier.m
```diff
@@ -47,7 +47,7 @@
 #import <AppKit/AppKit.h>
 #endif
 
-NSString *const NOTIFIER_VERSION = @"5.23.2";
+NSString *const NOTIFIER_VERSION = @"5.22.3";
```
twometresteve, line 50:
```suggestion
NSString *const NOTIFIER_VERSION = @"5.23.3";
```

VERSION
```diff
@@ -1 +1 @@
-5.23.2
+5.22.3
```
twometresteve, line 1:
```suggestion
5.23.3
```

twometresteve, line 1: I think the two `Podfile.lock`s for the examples should also be updated in line with this.

Bugsnag.podspec.json
twometresteve: I've provided Github suggestions for these, but actually it would be better to change them all in a single commit.

fractalwrench: Good catch - I've updated these in a single commit.

twometresteve:
In terms of testing I'm content that the our CI pipeline is generally sufficient for this to be merged, but we will need to manually test for regression of the original fix on an appropriate selection of platforms/OS versions before releasing.

PR #677, opened 2020-06-05
test: Add scenario that starts Bugsnag with no arguments

bengourley:
Spiking out automated tests for #676 – starting Bugsnag with no params/arguments, loading all config from the plist. Once the approach has been ratified I can add more assertions.

@twometresteve unfortunately this breaks your dynamic manual/automatic mode:

https://github.com/bugsnag/bugsnag-cocoa/blob/8ad1c32cb658fa50fafbc5dd59d0bb0a78947fe9/features/fixtures/ios-swift-cocoapods/iOSTestApp/ViewController.swift#L51-L63

Since there's no place to change the values at runtime, I had to hardcode the bs local url directly in the plist. I'm not sure of a good way to resolve this.


Cawllec:
Based on my knowledge of how the Android configuration loading worked, hard coding a URL into the plist shouldn't effect any configurations _except_ those that load it explicitly or call `Bugsnag.start` with no configuration.

Is this not the case?

Cawllec:
This should now be ready for a review

PR #678, opened 2020-06-05
Added plist loading to examples

tomlongridge:
## Goal

With the change of emphasis from configuration in code to plist, the examples have been amended to read from plist file, with commented out examples for how to override it in code.

PR also contains some general tidying of readme and project structure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 Changelog
 =========
 
-## 6.0.0 (TBD)
+## TBD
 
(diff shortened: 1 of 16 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 Changelog
 =========
 
-## 6.0.0 (TBD)
+## TBD
```
tomlongridge, line 4: This shouldn't have changed before - it's controlled by the makefile.

PR #683, opened 2020-06-08
Respect altered apiKey field on BugsnagEvent

fractalwrench:
## Goal

Use the value of `event.apiKey` when sending error reports, so that the correct key is used when making requests to the Bugsnag API.

This relies on changes in #669 which ensure events are not batched into a single request.

## Changeset

- Updated `BugsnagErrorReportSink` to use the `apiKey` set on the `BugsnagEvent`, after the point where `OnSendError` callbacks are executed
- Altered the constructor for `initWithUserData` as it was erroneously creating two objects when invoked. This seems to have been introduced in a merge - the redundant code has been removed and this is now more inline with the other two constructors.

## Tests
Manually verified that a handled and unhandled error use the correct API key.

### DIFF
--- a/Bugsnag/Configuration/BugsnagConfiguration.m
+++ b/Bugsnag/Configuration/BugsnagConfiguration.m
@@ -44,6 +44,7 @@
 #import "BugsnagStateEvent.h"
 #import "BugsnagCollections.h"
 #import "BugsnagMetadataInternal.h"
+#import "BugsnagKeys.h"
 
 static NSString *const kHeaderApiPayloadVersion = @"Bugsnag-Payload-Version";
 static NSString *const kHeaderApiKey = @"Bugsnag-Api-Key";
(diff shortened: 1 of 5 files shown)


Bugsnag/Configuration/BugsnagConfiguration.m
```diff
@@ -343,7 +344,7 @@ - (void)removeOnBreadcrumbBlock:(BugsnagOnBreadcrumbBlock _Nonnull)block {
 - (NSDictionary *)errorApiHeaders {
     return @{
              kHeaderApiPayloadVersion: @"4.0",
-             kHeaderApiKey: self.apiKey,
```
tomlongridge, line 346: This should either stay as-is or we should move the other constants to `BugsnagKeys.h`. I don't see any wider use to make this worthwhile moving.

fractalwrench: Moved constants to `BugsnagKeys`.

PR #694, opened 2020-06-10
Reformatted Upgrade Guide to better match Android

tomlongridge:
## Goal

The v6 upgrade guide has just been appended to during v6 development - this PR attempts to format it for easier consumption.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -52,30 +52,20 @@ The following functions have been added to the `Bugsnag` client:
 | Property/Method                                                    | Usage                                                             |
 | ------------------------------------------------------------------ | ----------------------------------------------------------------- |
 | `addOnBreadcrumb` / `removeOnBreadcrumb`<br />`addOnBreadcrumbBlock` / `removeOnBreadcrumbBlock` | Add/remove callbacks to modify or discard breadcrumbs before they are recorded. (See docs: [iOS](https://docs.bugsnag.com/platforms/ios/customizing-breadcrumbs/#discarding-and-amending-breadcrumbs) \| [macOS](https://docs.bugsnag.com/platforms/macos/customizing-breadcrumbs/#discarding-and-amending-breadcrumbs) \| [tvOS](https://docs.bugsnag.com/platforms/tvos/customizing-breadcrumbs/#discarding-and-amending-breadcrumbs)).
-| `addMetadata:metadata:toSection` / `addMetadata(metadata:section)` | Adds a dictionary of metadata to a section/tab. (See docs: [iOS](https://docs.bugsnag.com/platforms/ios/customizing-error-reports/#global-metadata) \| [macOS](https://docs.bugsnag.com/platforms/macos/customizing-error-reports/#global-metadata) \| [tvOS](https://docs.bugsnag.com/platforms/tvos/customizing-error-reports/#global-metadata)).
-| `getMetadata` / `getMetadataFromSection`                           | Retrieves previously set metadata from a section/tab. (See docs: [iOS](https://docs.bugsnag.com/platforms/ios/customizing-error-reports/#global-metadata) \| [macOS](https://docs.bugsnag.com/platforms/macos/customizing-error-reports/#global-metadata) \| [tvOS](https://docs.bugsnag.com/platforms/tvos/customizing-error-reports/#global-metadata)).
+| `addMetadata:metadata:toSection` / `addMetadata(metadata:section)` | Adds a dictionary of metadata to a section, shown as a tab on the Bugsnag dashboard. (See docs: [iOS](https://docs.bugsnag.com/platforms/ios/customizing-error-reports/#global-metadata) \| [macOS](https://docs.bugsnag.com/platforms/macos/customizing-error-reports/#global-metadata) \| [tvOS](https://docs.bugsnag.com/platforms/tvos/customizing-error-reports/#global-metadata)).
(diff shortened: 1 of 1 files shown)


UPGRADING.md
```diff
@@ -1,316 +1,277 @@
...
+### Bugsnag Client
+
+#### Starting Bugsnag
+
+You can now start Bugsnag using your application's `plist` file with configuration values, including your API key. The simplest start-up code is therefore:
 
 ```objc
-NSError *error;
-BugsnagConfiguration *config = [[BugsnagConfiguration alloc] initWithApiKey:@"YOUR API KEY HERE"];
+[Bugsnag start];
```
fractalwrench, line 26: IMO it'd be ok to just pick one language for examples - iOS developers have to read both languages so they'll be able to understand Swift, and it simplifies the flow.

UPGRADING.md
```diff
@@ -1,316 +1,277 @@
...
+| ------------------------------------------------------------------ | ----------------------------------------------------------------- |
+| `addOnBreadcrumb` / `removeOnBreadcrumb`<br />`addOnBreadcrumbBlock` / `removeOnBreadcrumbBlock` | Add/remove callbacks to modify or discard breadcrumbs before they are recorded. (See docs: [iOS](https://docs.bugsnag.com/platforms/ios/customizing-breadcrumbs/#discarding-and-amending-breadcrumbs) \| [macOS](https://docs.bugsnag.com/platforms/macos/customizing-breadcrumbs/#discarding-and-amending-breadcrumbs) \| [tvOS](https://docs.bugsnag.com/platforms/tvos/customizing-breadcrumbs/#discarding-and-amending-breadcrumbs)).
+| `addMetadata:metadata:toSection` / `addMetadata(metadata:section)` | Adds a dictionary of metadata to a section/tab. (See docs: [iOS](https://docs.bugsnag.com/platforms/ios/customizing-error-reports/#global-metadata) \| [macOS](https://docs.bugsnag.com/platforms/macos/customizing-error-reports/#global-metadata) \| [tvOS](https://docs.bugsnag.com/platforms/tvos/customizing-error-reports/#global-metadata)).
```
fractalwrench, line 55: section/tab -> section

UPGRADING.md
```diff
@@ -1,316 +1,277 @@
...
+| `clearTabWithName`                                                 | `clearMetadataFromSection`                                        |
+| `startBugsnagWithApiKey`                                           | `startWithApiKey`                                                 |
+| `startBugsnagWithConfiguration`                                    | `startWithConfiguration`                                          |
+| `stopSession`                                                      | `pauseSession`                                                    |
 
-### `Bugsnag` class
+##### Swift
 
-#### Removals
+| v5.x API                                                           | v6.x API                                                          |
```
fractalwrench, line 75: perhaps it would make sense to add this as an extra column onto the first table to reduce duplication?

UPGRADING.md
```diff
@@ -1,316 +1,277 @@
...
-#### Renames
+### Events (`BugsnagCrashReport`)
 
-```diff
-ObjC: 
+`BugsnagCrashReport` is now called `BugsnagEvent` and contains all the data representing the error that's been captured for access and/or mutation in callbacks.
 
-- [BugsnagMetadata clearTabWithName:]
-+ [BugsnagMetadata clearMetadataInSection:]
+#### Additions
```
fractalwrench, line 183: This does not render as a section title (when viewed via GH)

UPGRADING.md
```diff
@@ -1,316 +1,277 @@
...
+| `user` / `setUser:withEmail:andName` / `setUser(_:email:name)`     | The user of the app when the event occurred. (See docs: [iOS](https://docs.bugsnag.com/platforms/ios/customizing-error-reports/#setuser) \| [macOS](https://docs.bugsnag.com/platforms/macos/customizing-error-reports/#setuser) \| [tvOS](https://docs.bugsnag.com/platforms/tvos/customizing-error-reports/#setuser)).
 
-Swift:
+#### Changes
 
-- BugsnagMetadata.clearTab(name:)
-+ BugsnagMetadata.clearMetadata(section:)
+The `device`, `app`, `errors` and `threads` fields on the `BugsnagEvent` class are now structured classes, rather than `NSDictionary` types. (See docs: [iOS](https://docs.bugsnag.com/platforms/ios/customizing-error-reports/#the-bugsnagevent-class) \| [macOS](https://docs.bugsnag.com/platforms/macos/customizing-error-reports/#the-bugsnagevent-class) \| [tvOS](https://docs.bugsnag.com/platforms/tvos/customizing-error-reports/#the-bugsnagevent-class)).
```
fractalwrench, line 195: I'm not 100% sure these were available in `v5`

UPGRADING.md
tomlongridge: It was referred to as a tab in v5, so wanted to keep something of the old naming. Reworded now for clarity.

UPGRADING.md
tomlongridge: Reworded so this is no longer implied.

UPGRADING.md
tomlongridge: Merged tables using the style I've used further down.

UPGRADING.md
tomlongridge: @duncanhewett - do you have thoughts on whether we use just Obj-C or Swift? For the most part the changes are equivalent, but I don't think it loses much in readability to have both.

duncanhewett: I'm happy either way but it doesn't seem too distracting to have both.

PR #698, opened 2020-06-11
Verify that large breadcrumbs are no longer dropped

fractalwrench:
## Goal

Verifies that breadcrumbs with a serialized size > 4kb are not dropped when added to the Client. This behaviour changed in #433.

### DIFF
--- a/Tests/BugsnagClientTests.m
+++ b/Tests/BugsnagClientTests.m
@@ -241,10 +241,12 @@ NSString *BSGFormatSeverity(BSGSeverity severity);
 
     // large breadcrumb is also left without issue
     __block NSUInteger crumbSize = 0;
+    __block BugsnagBreadcrumb *crumb;
 
     [client addOnBreadcrumbBlock:^BOOL(BugsnagBreadcrumb *breadcrumb) {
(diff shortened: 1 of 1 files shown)


Tests/BugsnagClientTests.m
```diff
@@ -218,4 +222,49 @@ - (void)testChangesToConfigurationAreIgnoredAfterCallingStart {
...
+        crumbSize = data.length;
+        return true;
+    }];
+
+    NSDictionary *largeMetadata = [self generateLargeMetadata];
+    [client leaveBreadcrumbWithMessage:@"Hello World"
+                              metadata:largeMetadata
+                               andType:BSGBreadcrumbTypeManual];
+    XCTAssertTrue(crumbSize > 4096); // previous 4kb limit
+    XCTAssertEqual(2, [breadcrumbs count]);
```
twometresteve, line 256: Is it possible to check that the breadcrumb message/metadata hasn't been truncated, despite a breadcrumb entry having been successfully added?

fractalwrench: Updated to verify that the metadata + message are equal to the original values after a breadcrumb has been recorded by the Client.

PR #699, opened 2020-06-11
Improve callback crash E2E tests

fractalwrench:
## Goal

Improves E2E tests so that they verify callbacks do not crash the app. Additionally, existing scenarios have been updated so that they verify if one callback crashes, subsequent callbacks are also executed.

### DIFF
--- a/features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/SessionCallbackCrashScenario.swift
+++ b/features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/SessionCallbackCrashScenario.swift
@@ -27,7 +27,7 @@ class SessionCallbackCrashScenario : Scenario {
 
         // overwrite app ID set in first callback
         self.config.addOnSession { (session) -> Bool in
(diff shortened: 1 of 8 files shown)


features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/SessionCallbackCrashScenario.swift
```diff
@@ -27,7 +27,7 @@ class SessionCallbackCrashScenario : Scenario {
 
         // overwrite app ID set in first callback
```
twometresteve, line 28: I started by noting that this comment was out of date, but then I realised that the change means losing the test that  values overwritten in the second callback take precedent over those set in the first.

Perhaps add setting of the device id here, instead of replacing setting of the app id - then we keep both tests.  And update the comment of course.

fractalwrench: Updated the test so that it covers this scenario too.

PR #701, opened 2020-06-11
V6 examples update

Cawllec:
## Goal
Updates the swift example app to use `import Bugsnag` as stated in the integration guide.
In addition it upgrades the minimum supported iOS version of the `objective-c-ios` example app to 9.3,


Cawllec:
I've updated the app, removing the Custom-keyboard target and associated files.
I've also updated the Swift version to Swift 4, so it shouldn't complain upon opening.

@tomlongridge would you be able to re-review and quickly run the `objective-c-ios` to make sure it still builds by following the readme?

examples/swift-ios/bugsnag-example.xcodeproj/xcshareddata/xcschemes/bugsnag-example.xcscheme
```diff
@@ -0,0 +1,88 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
+   LastUpgradeVersion = "1150"
+   version = "1.3">
+   <BuildAction
```
fractalwrench, line 5: it looks like the schemes have been duplicated when opening the project as there are two objective-c-osx targets

examples/swift-ios/bugsnag-example.xcodeproj/xcshareddata/xcschemes/bugsnag-example.xcscheme
```diff
@@ -0,0 +1,88 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
```
fractalwrench, line 2: The swift target seems to have debug executable enabled by default

examples/objective-c-ios/Bugsnag Test App.xcodeproj/project.pbxproj
```diff
@@ -123,16 +111,6 @@
 /* End PBXFrameworksBuildPhase section */
 
 /* Begin PBXGroup section */
-		8ACFFA0A1D895D4D00357B5E /* custom-keyboard */ = {
```
fractalwrench, line 126: This seems to have been removed from the example but is still present as a scheme in XCode?

Cawllec:
I've updated based on the feedback, removing tests where necessary.
Regarding the other issues, I can't see any duplicate targets/schemes when opening the app, and the `debug executable` setting is disabled by default when I open the app fresh.
If this still isn't the case, it may mean settings aren't being saved correctly.

PR #703, opened 2020-06-12
Ensure connectivity check listens for gain of connection

fractalwrench:
## Goal

Sets the default value for `bsg_current_reachability_state` to -1. This prevents the connectivity listener from not firing when a connection is regained, as 0 is a valid value for the bitmask cached later in the method.

### DIFF
--- a/Bugsnag/Delivery/BSGConnectivity.m
+++ b/Bugsnag/Delivery/BSGConnectivity.m
@@ -51,7 +51,9 @@ BOOL BSGConnectivityShouldReportChange(SCNetworkReachabilityFlags flags) {
     #endif
     __block BOOL shouldReport = YES;
     // Check if the reported state is different from the last known state (if any)
-    if ((flags & importantFlags) != (bsg_current_reachability_state & importantFlags)) {
+    SCNetworkReachabilityFlags newFlags = flags & importantFlags;
+    SCNetworkReachabilityFlags oldFlags = bsg_current_reachability_state & importantFlags;
+    if (newFlags != oldFlags) {
         // When first subscribing to be notified of changes, the callback is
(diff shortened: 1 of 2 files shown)


Bugsnag/Delivery/BSGConnectivity.m
```diff
@@ -51,11 +51,13 @@ BOOL BSGConnectivityShouldReportChange(SCNetworkReachabilityFlags flags) {
...
     // Check if the reported state is different from the last known state (if any)
-    if ((flags & importantFlags) != (bsg_current_reachability_state & importantFlags)) {
+    SCNetworkReachabilityFlags newFlags = flags & importantFlags;
+    SCNetworkReachabilityFlags oldFlags = bsg_current_reachability_state & importantFlags;
+    if (newFlags != oldFlags) {
         // When first subscribing to be notified of changes, the callback is
         // invoked immmediately even if nothing has changed. So this block
         // ignores the very first check, reporting all others.
-        if (bsg_current_reachability_state == 0) {
+        if (bsg_current_reachability_state == -1) {
```
tomlongridge, line 60: For this to work, don't we need to initialize `bsg_current_reachability_state` to -1 to catch the first call after registering?

fractalwrench: Yes - this was a local change that didn't make its way into this changeset, I've updated the PR.

fractalwrench:
Additionally removed one unit test that is inherently flaky on CI due to the use of global mutable state in the implementation.

Cawllec:
Manually verified on a device that both the Connectivity Breadcrumb is appearing when WI-FI is enabled, and that delivery occurs when the connectivity is re-enabled.

PR #712, opened 2020-06-18
Ensure client metadata/user is serialized for unhandled errors

fractalwrench:
## Goal

Metadata added via the client and the user is not serialized for unhandled errors. This occurs because the `BugsnagClient` now makes a copy of the metadata on `BugsnagConfiguration`, but is still attempting to serialize the `BugsnagConfiguration` metadata.

This updates the `BugsnagClient` so that Client's metadata is serialized into JSON whenever it changes, so that it is ready to be added to any unhandled error that occurs.

## Tests

Manually verified that an unhandled error has custom data/user information added via the client.


twometresteve:
It would be good to reinforce this by following up with an additional e2e test.  I'm happy if we just raise an internal ticket for that at this stage.

PR #713, opened 2020-06-19
fix(client): remove client operations for adding/removing callbacks

tomlongridge:
## Goal

Decision made that modifying delivery-time callbacks on the client doesn't seem very useful (as it's on the subsequent launch) and likely to cause confusion, so removing them from the API.

## Changeset

* Removed from code and headers
* Removed tests for the methods specifically
* Updated expectations on e2e tests

### DIFF
--- a/Bugsnag/Bugsnag.h
+++ b/Bugsnag/Bugsnag.h
@@ -272,27 +272,6 @@
     NS_SWIFT_NAME(removeOnSession(block:));
 
 // =============================================================================
-// MARK: - onSend
-// =============================================================================
-
-/**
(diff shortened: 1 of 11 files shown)


twometresteve:
Nit: The changelog entry includes the following.  Technically true, but potentially confusing (suggest removing "`Bugsnag` and")
```
* Added `addOnSendBlock:`, `removeOnSendBlock:` and `clearOnSendBlocks` methods to `Bugsnag` 
  and `BugsnagConfiguration`.
  [#485](https://github.com/bugsnag/bugsnag-cocoa/pull/485)
```

PR #714, opened 2020-06-19
fix(swift example app): Modernise the iOS example Swift app

xander-jones:
## Goal
Modernise the Swift iOS example app:
* Expand the hit area of buttons
* Port the Out of Memory function from the Objective C example application.

## Changeset
* Added `LaunchScreen.storyboard` in favour of `LaunchScreen.xib`
* Updated `Main.storyboard` to use `UITableView` directly rather than buttons on top
* Updated the `ViewController.swift` to use the `UITableView` rather than linking `@IBAction`s
* Ported OOM Objective C code from `BigHonkinWebViewController.m/.h`

## Tests
Tested on iPhone simulators: iPhone SE (2nd Gen) iOS 13.4; iPhone 11 Pro iOS 13.4


examples/swift-ios/bugsnag-example/BigHonkinWebViewController.h
```diff
@@ -0,0 +1,4 @@
+#import <UIKit/UIKit.h>
+@interface BigHonkinWebViewController : UIViewController
```
twometresteve, line 2: I'm fairly sure the the big honker was removed in the past because we weren't keen on the frivolous naming.  Could you rename the files to indicate their actual purpose?

xander-jones: Name changed to `OutOfMemoryController.m/h`, references all updated.

PR #715, opened 2020-06-19
Ensure app/device metadata is added to handled error reports

fractalwrench:
## Goal

Certain App/Device metadata fields were missing from handled error reports due to a refactor in how these are serialized on disk. This PR reinstates them.

## Changeset

Reinstated the following fields for handled errors:

```
app.dsymUUIDs - missing due to bug in deserialization function
device.orientation - not added to `BugsnagDeviceWithState`
metaData.app.name - now populated in initial self.metadata
metaData.device.batteryLevel - populated from self.metadata
metaData.device.charging - populated from self.metadata
metaData.device.simulator - now populated in initial self.metadata
metaData.device.timezone - now populated in initial self.metadata
metaData.device.wordSize - now populated in initial self.metadata
```

## Tests

Updated unit test coverage. Manually verified that a handled + unhandled error contain the expected information.

### DIFF
--- a/Bugsnag/Client/BugsnagClient.m
+++ b/Bugsnag/Client/BugsnagClient.m
@@ -96,6 +96,9 @@ static NSUInteger handledCount;
 static NSUInteger unhandledCount;
 static bool hasRecordedSessions;
 
+NSDictionary *BSGParseAppMetadata(NSDictionary *event);
+NSDictionary *BSGParseDeviceMetadata(NSDictionary *event);
+
 @interface NSDictionary (BSGKSMerge)
 - (NSDictionary *)BSG_mergedInto:(NSDictionary *)dest;
 @end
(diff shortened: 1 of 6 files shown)


Bugsnag/Client/BugsnagClient.m
```diff
@@ -617,6 +618,11 @@ - (void)start {
     // notification not received in time on initial startup, so trigger manually
     [self willEnterForeground:self];
     [self.pluginClient loadPlugins];
+
+    // add metadata about app/device
+    NSDictionary *systemInfo = [BSG_KSSystemInfo systemInfo];
+    [self.metadata addMetadata:BSGParseAppMetadata(@{@"system": systemInfo}) toSection:BSGKeyApp];
+    [self.metadata addMetadata:BSGParseDeviceMetadata(systemInfo) toSection:BSGKeyDevice];
```
tomlongridge, line 625: Seems to make sense to wrap this in a "system" key to match App and then you don't need to check in two places in `BSGParseDeviceMetadata `.

```suggestion
    [self.metadata addMetadata:BSGParseDeviceMetadata(@{@"system": systemInfo}) toSection:BSGKeyDevice];
```

PR #728, opened 2020-06-28
refactor(mach headers): refactor to linked list to avoid threading issues

tomlongridge:
## Goal

The current implementation of the Mach Headers cache using an expandable array is not easy to make work in a thread-safe way, so wouldn't be usable if threads were not suspended before accessing. 

## Design

Refactored to a linked list so that there is no `realloc` and less change when new elements are added. Also changed removal so the library is simply marked as removed rather than deleted so that locking is not required for this operation.

## Changeset

* `BSG_KSMachHeaders` - amended implementation to a linked list; renamed functions for internal consistency; amended removal to simply mark the library as removed rather than mutate the list and so removed the lock around the operation; refactored lock functions to work in a cleaner way.
* `BSG_KSCrash` - moved initialisation code to `BSG_KSCrashC`, closer to other initialisation code and moved the registering of changes into the initialisation function of the mach header code.
* `BSG_KSDynamicLinker` - moved several functions to `BSG_KSMachHeaders` to avoid passing indexes around.

## Tests

* Manually tested the reports remain symbolicated
* Amended existing unit tests accordingly - including added a `free` call during initialization so that the list of headers doesn't grow between each test.


Tests/KSCrash/KSDynamicLinker_Tests.m
```diff
@@ -28,13 +28,18 @@
 #import <XCTest/XCTest.h>
 
 #import "BSG_KSDynamicLinker.h"
+#import "BSG_KSMachHeaders.h"
 
 @interface KSDynamicLinker_Tests : XCTestCase @end
 
 @implementation KSDynamicLinker_Tests
 
+// TODO:TL Fix test
```
fractalwrench, line 37: is this comment resolved?

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
```diff
@@ -271,21 +267,6 @@ - (BOOL)install {
     return true;
 }
 
-/**
- * Set up listeners for un/loaded frameworks.  Maintaining our own list of framework Mach
- * headers means that we avoid potential deadlock situations where we try and suspend
- * lock-holding threads prior to loading mach headers as part of our normal event handling
- * behaviour.
- */
```
fractalwrench, line 279: nit: would be good to retain this comment at the first invocation of `bsg_mach_headers_initialize`

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSDynamicLinker.c
```diff
@@ -32,37 +32,13 @@
...
-        if (iImg != UINT32_MAX) {
-            const struct mach_header *header =  bsg_dyld_get_image_header(iImg);
-            if (header != NULL) {
-                uintptr_t cmdPtr = bsg_ksdlfirstCmdAfterHeader(header);
+        BSG_Mach_Header_Info *img = bsg_mach_headers_image_named(imageName, exactMatch);
+        if (img != NULL) {
+            uintptr_t cmdPtr = bsg_mach_headers_first_cmd_after_header(img->header);
                 if (cmdPtr != 0) {
-                    for (uint32_t iCmd = 0; iCmd < header->ncmds; iCmd++) {
+                for (uint32_t iCmd = 0; iCmd < img->header->ncmds; iCmd++) {
```
fractalwrench, line 41: nit: this for block has an inconsistent indentation

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -88,120 +68,58 @@ BOOL IsUnfairLockSupported(NSProcessInfo *processInfo) {
...
-uint32_t bsg_dyld_image_count(void) {
-    return bsg_mach_binary_images.used;
-}
+static void (*bsg_mach_headers_cache_lock_func)(void) = NULL;
+static void (*bsg_mach_headers_cache_unlock_func)(void) = NULL;
 
-const struct mach_header* bsg_dyld_get_image_header(uint32_t imageIndex) {
-    if (imageIndex < bsg_mach_binary_images.used) {
-        return bsg_mach_binary_images.contents[imageIndex].header;
+void bsg_mach_headers_cache_lock() {
```
fractalwrench, line 77: The docs for `[NSProcessInfo processInfo]` states the return value is cached after the first invocation. Therefore I think the null check and assignment to static variable can be removed here as it will be clearer if `bsg_unfair_lock` and `bsg_spin_lock` are just invoked directly

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -88,120 +68,58 @@ BOOL IsUnfairLockSupported(NSProcessInfo *processInfo) {
...
-            // Note: removal of the last (ith) item involves a redundant copy from last->last.
-            if (bsg_mach_binary_images.used >= 2) {
-                bsg_mach_binary_images.contents[i] = bsg_mach_binary_images.contents[--bsg_mach_binary_images.used];
-            }
-            else {
-                bsg_mach_binary_images.used = 0;
-            }
-            break; // an image can only be loaded singularly; exit loop once found
-        }
+    // Free any existing data (this may be present in tests)
```
fractalwrench, line 110: This should be removed if it is only required for the test condition

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
```diff
@@ -272,35 +190,145 @@ bool bsg_populate_mach_image_info(const struct mach_header *header, intptr_t sli
...
-    // Convert header into an info struct
-    BSG_Mach_Binary_Image_Info info = { 0 };
-    if (bsg_populate_mach_image_info(header, slide, &info)) {
-        bsg_remove_mach_binary_image(info.imageVmAddr);
+void bsg_mach_headers_remove_image(const struct mach_header *header, intptr_t slide) {
+    BSG_Mach_Header_Info existingImage = { 0 };
+    if (bsg_mach_headers_populate_info(header, slide, &existingImage)) {
+        for (BSG_Mach_Header_Info *img = bsg_mach_headers_get_images(); img != NULL; img = img->next) {
+            if (img->imageVmAddr == existingImage.imageVmAddr) {
+                img->removed = true;
```
fractalwrench, line 225: Is freeing not a possibility here? I suspect it's not due to the need for thread safety when traversing the linked list but it'd be good to confirm that, and add some docs around why this is the case for future travellers

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
tomlongridge: It was a bit detailed for a comment outside the mach-headers source, but I've added it to the header and made a more general comment here.

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
tomlongridge: That would certainly be cleaner! It seems that is an efficient call, so I'll make the change.

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMachHeaders.m
tomlongridge: As discussed IRL, by making this non-destructive we're avoiding threading issues. Therefore the memory is only freed if the library is reinstalled. Adding some comments to clarify that.

PR #729, opened 2020-06-29
Support swift package manager

### LINKED ISSUE #407: Add support for Swift Package Manager
### Description
I think pretty self explanatory. Not sure if there is anything blocking this, if wanted I can take a look at this.

imjoehaines:
## Goal

This PR adds support for Swift Package Manager

This involved moving the header files to a new `include` directory to match the expected convention for C projects in SPM. This isn't absolutely necessary, for example we could use symlinks instead, but this felt like the cleanest solution

Resolves #407 

## Changeset

Most of the changes are moving header files into `include/Bugsnag`. The `podspec` and Xcode project file have been updated with these new paths too

I've removed some unnecessary imports in `BugsnagConfiguration.m`. This was part of an effort to split up the package into sup-dependencies which was ultimately abandoned, but I figured this was a useful cleanup anyway. Similarly, `BugsnagPlugin` was missing an import which causes issues if it's built before the header it relies on is included, so I added that too

## Tests

I've tested this in our Swift example app and new Swift projects

Carthage and Cocoapods are tested on CI and I've manually tested them in the example projects & static build. I've also tested that our manual install instructions still work


Package.swift
```diff
@@ -0,0 +1,63 @@
...
+                "Delivery",
+                "Helpers",
+                "KSCrash",
+                "Metadata",
+                "Payload",
+                "Plugins",
+                "KSCrash/Source/KSCrash/Reporting/Filters",
+                "KSCrash/Source/KSCrash/Recording/Tools",
+                "KSCrash/Source/KSCrash/Recording",
+                "KSCrash/Source/KSCrash/Recording/Sentry",
```
tomlongridge, line 34: Can you check whether this list is necessary? It's missing `Storage` currently, which suggests it might not be.

Package.swift
```diff
@@ -0,0 +1,63 @@
...
+                .headerSearchPath("Helpers"),
+                .headerSearchPath("KSCrash"),
+                .headerSearchPath("Metadata"),
+                .headerSearchPath("Payload"),
+                .headerSearchPath("Plugins"),
+                .headerSearchPath("Storage"),
+                .headerSearchPath("KSCrash/Source/KSCrash/Reporting/Filters"),
+                .headerSearchPath("KSCrash/Source/KSCrash/Recording/Tools"),
+                .headerSearchPath("KSCrash/Source/KSCrash/Recording"),
+                .headerSearchPath("KSCrash/Source/KSCrash/Recording/Sentry"),
```
tomlongridge, line 53: Alphabetise these to help with maintenance.

Package.swift
imjoehaines: Removed in 5537bb87dc686fb920dc7fbd4a397e7dec186535

.travis.yml
```diff
@@ -93,6 +93,12 @@ jobs:
     script:
     - make build_carthage   # Build example carthage project
     - make build_ios_static # Build static framework target
+    
+  - osx_image: xcode11
+    name: Building with Swift Package Manager
+    stage: integration
+    script:
+    - swift build
```
tomlongridge, line 101: Can we add this to the makefile and run under the same step?

tomlongridge:
The `doc` target on the Makefile uses the `public_header_files` entry to generate docs for each entry. I don't think this will now work with the wildcard - can you check and find an alternative? (It's definitely better with the wildcard, so we should fix the step not the podspec file.)

imjoehaines:
`make doc` works fine with the wildcard as it uses `Dir.glob` 🙂 Someone must have seen this coming!

examples/swift-package-manager/swift-package-manager.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,372 @@
...
+			defaultConfigurationIsVisible = 0;
+			defaultConfigurationName = Release;
+		};
+/* End XCConfigurationList section */
+
+/* Begin XCRemoteSwiftPackageReference section */
+		3A837DD324AA4A0E00A54700 /* XCRemoteSwiftPackageReference "bugsnag-cocoa" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/bugsnag/bugsnag-cocoa.git";
+			requirement = {
```
twometresteve, line 356: Can we change this to "Version Up to next major 6.1.0 < 7" or similar before merging - so that it's correct after release.

PR #732, opened 2020-06-30
Reinstated suspend threads at point of tracing

tomlongridge:
## Goal

The only area of code we need to protect with suspending threads for handled errors is when/if the user wants to capture all thread data (using the `sendThreads` configuration option). This is the opposite to #723, so I've reinstated the suspend call and removed it from the report generation.

## Design

To make it clearer, I've removed the `suspendThreadsForUserReported` flag entirely and made everything conditional on `threadTracingEnabled` and `crashType`.

## Tests

Checked the correct behaviour manually when setting the `sendThreads` option.


features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/ManyConcurrentNotifyNoBackgroundThreads.m
```diff
@@ -2,7 +2,6 @@
 
 @interface BSG_KSCrash
 + (instancetype)sharedInstance;
-- (void)setSuspendThreadsForUserReported:(BOOL)suspend;
 @end
 
 @interface ManyConcurrentNotifyNoBackgroundThreads ()
```
fractalwrench, line 7: This scenario can be removed as it's now identical to `ManyConcurrentNotifyScenario`

Issue #737, opened 2020-07-02
Memory leak on BugsnagConfig


jbelkins:
### Description
Went looking for memory leaks in my own code today and ran across this on latest Bugsnag iOS SDK.

### Issue
Found that `BugsnagConfiguration` leaks due to a reference cycle between it and `BugsnagBreadcrumbs`.

The culprit is [here](https://github.com/bugsnag/bugsnag-cocoa/blob/master/Bugsnag/Configuration/BugsnagConfiguration.m#L223):
```
    _breadcrumbs = [[BugsnagBreadcrumbs alloc] initWithConfiguration:self];
```
`BugsnagBreadcrumbs` keeps a strong reference to the `self` you pass in, causing a reference cycle.  It's not major, but it's leaked memory.  How to fix is up to you.

### Environment
Library versions:
* Bugsnag version: 6.0.1 from Cocoapods

xander-jones:
Hi @jbelkins, thanks for reporting this. We're going to be taking a look into this as a priority.

johnkiely1:
This has now been fixed in [V6.1.1](https://github.com/bugsnag/bugsnag-cocoa/releases/tag/v6.1.1)

PR #743, opened 2020-07-04
fix(breadcrumbs): guard read access to breadcrumbs count

### LINKED ISSUE #199: Are breadcrumbs not thread-safe?

### LINKED ISSUE #616: Breadcrumb API is not thread-safe
### Description

While running our app under the thread sanitiser, it frequently hits a race condition on the Breadcrumb API as seen here:

Screenshot 2020-05-22 at 11 09 20
Screenshot 2020-05-22 at 11 09 41

### Environment

Library versions:

* Bugsnag version (from your Podfile, Podfile.lock, or elsewhere):
* CocoaPods version (if any) (`pod -v`):
* Carthage version (if any):
* iOS/tvOS/macOS version(s)
* debug mode or production?:
* How are you initializing Bugsnag? Can you share a snippet of your
  `AppDelegate` where Bugsnag is configured?

### Example code snippet

```swift
import Bugsnag

// (Insert code sample to reproduce the problem)
```

<details><summary>Error messages:</summary>

```

```
</details>

tomlongridge:
## Goal

Fixes #616 

## Design

Refactored access to breadcrumb list to avoid the need for the `readWriteQueue` guard in most cases.

## Changeset

* BugsnagBreadcrumbs - removed `capacity`, `count`, and `objectAtIndexedSubscript` as no longer needed
* BugsnagBreadcrumbs:arrayValue - removed `readWriteQueue` guard and used a copy of the list instead; also made signature nonnull to return empty array if there are no elements.
* BugsnagConfiguration / BugsnagClient - moved the `breadcrumbs` field from configuration to client and added fields to configuration for `maxBreadcrumbs` and `enabledBreadcrumbTypes`. These are then used in the `BugsnagBreadcrumbs` constructor. This prevents any need for resizing the array of breadcrumbs outside the class.

## Tests

* Enabled Thread Sanitizer and added breadcrumbs in a tight loop to check for issues.
* Amended unit tests accordingly


tomlongridge:
I've removed internal access to the `count` property. Now we have a clearer distinction between internal and external headers, can we remove it?

Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
```diff
@@ -137,52 +137,56 @@ - (void)setCapacity:(NSUInteger)capacity {
...
 - (NSUInteger)count {
-    return self.breadcrumbs.count;
+    __block NSUInteger count;
+    dispatch_barrier_sync(self.readWriteQueue, ^{
+        count = self.breadcrumbs.count;
+    });
+    return count;
 }
 
 - (BugsnagBreadcrumb *)objectAtIndexedSubscript:(NSUInteger)index {
```
fractalwrench, line 147: This method appears to be unused so can be removed

Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
```diff
@@ -137,52 +137,56 @@ - (void)setCapacity:(NSUInteger)capacity {
...
-                              @"JSON object");
-                  continue;
-              }
-              [contents addObject:objectValue];
-          } @catch (NSException *exception) {
-              bsg_log_err(@"Unable to serialize breadcrumb: %@", error);
-          }
-      }
+        NSUInteger numBreadcrumbs = self.breadcrumbs.count;
+        if (numBreadcrumbs > 0) {
```
fractalwrench, line 161: nit: removing this check would mean that the method never needs to return `nil` and could be annotated as nonnull - which feels a better behaviour for an empty collection.

Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
```diff
@@ -137,52 +137,56 @@ - (void)setCapacity:(NSUInteger)capacity {
...
 }
 
 - (NSArray *)arrayValue {
-    if ([self count] == 0) {
-        return nil;
-    }
-    __block NSMutableArray *contents =
-        [[NSMutableArray alloc] initWithCapacity:[self count]];
+    __block NSMutableArray *contents = nil;
     dispatch_barrier_sync(self.readWriteQueue, ^{
```
fractalwrench, line 159: I'm not convinced that a dispatch barrier is necessary - being defensive and serializing a `mutableCopy` of `self.breadcrumbs` should be sufficient IMO.

tomlongridge:
> This touches more on the overall class design, but it's worth noting that using `dispatch_barrier_sync` for a method that can be called from the main thread is not great. This is because it will block until IO completes.
> 
> Fortunately I think it should be possible to replace this with `dispatch_barrier_async` because the only remaining method after implementing these suggestions would be `cachedBreadcrumbs`. This is always called before breadcrumb collection happens, so could be extracted out to `BugsnagClient/BSGOutOfMemoryWatchdog`. This is potentially a bit out of the scope of this changeset so it'd be acceptable to address this when the breadcrumb disk serialization is revisited in general.

I haven't addressed this as there are wider issues trying to run this asynchronously. I'll raise this separately for a future change as this should be a good efficiency gain.

Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
```diff
@@ -59,14 +55,16 @@ - (void)addBreadcrumb:(NSString *)breadcrumbMessage {
...
-    if (self.capacity == 0) {
+    if (self.config.maxBreadcrumbs == 0) {
         return;
     }
     BugsnagBreadcrumb *crumb = [BugsnagBreadcrumb breadcrumbWithBlock:block];
 
     if (crumb != nil && [self shouldSendBreadcrumb:crumb]) {
-        [self resizeToFitCapacity:self.capacity - 1];
         dispatch_barrier_sync(self.readWriteQueue, ^{
+            while ([self.breadcrumbs count] >= self.config.maxBreadcrumbs) {
```
fractalwrench, line 65: This will crash if `maxBreadcrumbs == 0`

Bugsnag/Configuration/BugsnagConfiguration.m
```diff
@@ -455,16 +450,23 @@ -(void)deletePersistedUserData {
 // MARK: - Properties: Getters and Setters
 // -----------------------------------------------------------------------------
 
+@synthesize maxBreadcrumbs = _maxBreadcrumbs;
+
 - (NSUInteger)maxBreadcrumbs {
-    return self.breadcrumbs.capacity;
+    @synchronized (self) {
```
fractalwrench, line 456: I don't see why this needs to take out a lock - I'm aware this is done elsewhere for other properties but think this could be removed for the majority if not all of those custom accessors, unless there's a specific reason. Possibly something to raise in another ticket.

Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
tomlongridge: This is excluded above (L58).

Bugsnag/Configuration/BugsnagConfiguration.m
tomlongridge: I'd rather stick with the pattern used already - will raise a ticket to consider this in future.

Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
tomlongridge: In fact, it makes me nervous in case the count is zero, so I've removed the loop (it will never be more than the max) and added a check for count>0.

tomlongridge:
This PR also fixes #737

raylillywhite:
I'm using commit 9044a26a6f3e8909209b598d6ecc12d4d20adb23 and still getting crashes

```
EXC_BAD_ACCESS · Attempted to dereference garbage pointer 0xd590dd120.
/usr/lib/libobjc.A.dylib_objc_msgSend	
Bugsnag/Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m:122:37-[BugsnagBreadcrumbs arrayValue]	
Bugsnag/Bugsnag/Client/BugsnagClient.m:1096:29-[BugsnagClient serializeBreadcrumbs]	
Bugsnag/Bugsnag/Client/BugsnagClient.m:781:5-[BugsnagClient leaveBreadcrumbWithMessage:metadata:andType:]	
Bugsnag/Bugsnag/Bugsnag.m:187:9+[Bugsnag leaveBreadcrumbWithMessage:metadata:andType:]
```

Link [here](https://app.bugsnag.com/spruce-health/care-messenger-ios/errors/5f075415a85fb200171927ab?filters%5Bevent.since%5D%5B%5D%5Btype%5D=eq&filters%5Bevent.since%5D%5B%5D%5Bvalue%5D=30d&filters%5Berror.status%5D%5B%5D%5Btype%5D=eq&filters%5Berror.status%5D%5B%5D%5Bvalue%5D=open&filters%5Bapp.release_stage%5D%5B%5D%5Bvalue%5D=staging&filters%5Bapp.release_stage%5D%5B%5D%5Btype%5D=eq&event_id=5f07fe58005b8c49c1270000) in case you have access to all bugsnag data

tomlongridge:
Thanks @raylillywhite - I'm investigating now.

tomlongridge:
@raylillywhite - I haven't been able to reproduce this, but it does seem that we're somehow holding on to a freed reference, even though the mutable copy should have preserved the reference.

Are you able to readily reproduce this in a non-live environment? I have pushed a change (808b8bce8d0c99e2c16a2c2f89ae0a8129b38fa8) that reinstates the barrier guard for `arrayValue` so that the operation is protected as it was before #743. I was wondering if you'd be able to see if this resolves the issue for you?

raylillywhite:
I can't reproduce it on demand, but we saw it quite a few times (we log a lot of breadcrumbs). And we're not shipping for a few weeks so I'll use that commit in our next test build today and see if it still happens.

Issue #747, opened 2020-07-08
BugsnagBreadcrumbs.arrayValue returns NSArray<NSDictionary*> * not NSArray<NSString*> * as documented


bolsinga:
https://github.com/bugsnag/bugsnag-cocoa/blob/2fd7c67e65bfc5523ecea4705066f9591c831aca/Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m#L170

We are going to pull the breadcrumbs into our internal bug reports. I found the types are incorrectly documented. I'm not sure how, but pre-6.0 the JSON data was decoded, but the dictionary was added to the array. In 6.0 the JSON is not decoded, but the dictionary is still in the array. it's still checked for JSON compatibility.

It should be documented correctly. I also think it would be good to have strings in the array, but that is an open question.

phillipsam:
Hi @bolsinga - yes this has changed in v6, as you have suggested. Thanks for bringing this to our attention, we will look to fix the header docs in the next release.

This is not actually on a public method. The current recommended way of exposing breadcrumbs is by using an OnBreadcrumbCallback method: https://docs.bugsnag.com/platforms/android/configuration-options/#addonbreadcrumb

bolsinga:
@phillipsam Thanks. Since our app is what is leaving the breadcrumbs, having a callback to know what breadcrumb was just sent seems redundant. If there is not a way to query the current breadcrumbs, then it would be best for us to cache the breadcrumbs being sent to Bugsnag and not implement a callback to get the same data. 

If possible, please re-consider having a way to request the current breadcrumbs. W e're working on making our internal bug reports and the data found by Bugsnag "line-up". Thanks.

phillipsam:
Hi @bolsinga - thanks for the clarification. One option available is to use an `OnSendError` callback to get the breadcrumbs list in an event:
https://github.com/bugsnag/bugsnag-cocoa/blob/master/Bugsnag/include/Bugsnag/BugsnagEvent.h#L60

bolsinga:
Hi @phillipsam I haven't been able to update to 6.0 yet because when I last checked, it isn't integrated into BugsnagReactNative / iOS yet. But when I search for "BugsnagEvent alloc" it seems it is created once a crash is detected.

[...]

PR #748, opened 2020-07-08
docs(breadcrumbs): updated outdated docs for BugsnagBreadcrumbs.arrayValue

tomlongridge:
## Goal

As per #747, the docs on BugsnagBreadcrumbs.arrayValue were out of date as they now return an array of dictionaries, not strings.


snmaynard:
@tomlongridge should we not document the keys?

tomlongridge:
This is an internal method, so I didn't think it was necessary as it will change with the structure of the `BugsnagBreadcrumb` class.

PR #755, opened 2020-07-16
Ensure foreground stats are recorded for handled errors

fractalwrench:
## Goal

Ensures that foreground stats are recorded correctly for handled errors. This fixes a bug where `app.inForeground`, `app.durationInForeground`, and `app.duration` were captured for unhandled errors but not for handled errors. This is due to the two errors using a different mechanism for capture.

## Changeset

- Added `[BSG_KSCrash captureApplicationStatsInfo]` to `BSG_KSSystemInfo` to ensure that application stats are always captured for handled errors
- Returned the value of `bsg_kscrw_i_writeAppStats` as a JSON string when capturing application stats
- Renamed `BSG_ThreadDataBuffer` to `BSG_JSONDataBuffer` as the struct can be used as a buffer for both threads + app stats

## Tests

Verified locally in an example app that the fields are now populated.


Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.h
```diff
@@ -119,6 +119,11 @@
  */
 - (NSArray<BugsnagThread *> *)captureThreads:(NSException *)exc depth:(int)depth;
 
+/**
+ * Collects information about the application's foreground state (duration in foreground/background)
+ */
+- (NSDictionary *)captureApplicationStatsInfo;
```
tomlongridge, line 125:
```suggestion
/**
 * Collects information about the application's state (e.g. duration in foreground/background)
 */
- (NSDictionary *)captureAppStats;
```

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.h
```diff
@@ -74,6 +74,13 @@ void bsg_kscrashreport_logCrash(const BSG_KSCrash_Context *const crashContext);
  */
 char *bsg_kscrw_i_captureThreadTrace(const BSG_KSCrash_Context *crashContext);
 
+/**
+ * Captures app foreground stats for use by the Objective-C layer to append to handled errors.
+ *
+ * @return the stats encoded as a JSON string
+ */
+char *bsg_kscrw_i_captureAppForegroundStats(const BSG_KSCrash_Context *crashContext);
```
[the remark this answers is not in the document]
fractalwrench: Renamed all methods as specified

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1700,6 +1700,18 @@ char *bsg_kscrw_i_captureThreadTrace(const BSG_KSCrash_Context *crashContext) {
     return userData.data;
 }
 
+char *bsg_kscrw_i_captureAppForegroundStats(const BSG_KSCrash_Context *crashContext) {
```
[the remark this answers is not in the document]
fractalwrench: I observed a warning about `bsg_kscrw_i_writeAppStats` not declaring its 3rd parameter as `const` so have addressed that - I'm assuming that's the same as what you observed.

fractalwrench:
Updated approach to avoid serialization by getting the information from the KSCrash state directly. This necessitates a small refactor of the logic for updating the duration/durationInForeground fields, which need to be recalculated whenever an error is captured.

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashState.m
```diff
@@ -72,6 +72,8 @@
 // Avoiding static functions due to linker issues.
 
 // ============================================================================
+void bsg_kscrashstate_updateForegroundStats(BSG_KSCrash_State *const state);
```
tomlongridge, line 75: Declaration can be removed.

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashState.h
```diff
@@ -121,6 +121,14 @@ void bsg_kscrashstate_notifyAppCrash(BSG_KSCrashType type);
  */
 const BSG_KSCrash_State *bsg_kscrashstate_currentState(void);
 
+/**
+ * Updates the stats for duration in foreground/background. This needs to
+ * be updated whenever an error report is captured.
+ *
+ * @param state the kscrash state
+ */
+void bsg_kscrashstate_updateForegroundStats(BSG_KSCrash_State *const state);
```
tomlongridge, line 130:
```suggestion
void bsg_kscrashstate_updateDurationStats(BSG_KSCrash_State *const state);
```

PR #763, opened 2020-07-23
Populate app/device metadata fields for RN errors

fractalwrench:
## Goal

Fixes the population of the following app/device metadata fields in RN errors:

```
app.name
device.batteryLevel
device.charging
device.simulator
device.timezone
device.wordsize
```

This changeset comes in two parts:

1. Performing an initial sync of metadata whenever a state event observer is added. This allows the JS layer on React Native to pass down the stateless fields in metadata.
2. Always adding stateful device metadata fields when `notifyInternal` is called.

## Tests

Verified in a local example app that the fields are populated.

### DIFF
--- a/Bugsnag/Client/BugsnagClient.m
+++ b/Bugsnag/Client/BugsnagClient.m
@@ -458,6 +458,7 @@ NSString *_lastOrientation = nil;
 
     // additionally listen for metadata updates
     [self.metadata addObserverWithBlock:observer];
+    [self.metadata notifyObservers];
 }
 
 - (void)removeObserverWithBlock:(BugsnagObserverBlock _Nonnull)observer {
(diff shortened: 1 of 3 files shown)


Bugsnag/Client/BugsnagClient.m
```diff
@@ -454,6 +458,7 @@ - (void)addObserverWithBlock:(BugsnagObserverBlock _Nonnull)observer {
 
     // additionally listen for metadata updates
     [self.metadata addObserverWithBlock:observer];
+    [self.metadata notifyObservers];
```
tomlongridge, line 461: I believe this would cause an extra 2 writes to disk of existing metadata as Client would already be registered and would get triggered when RN loads.

As a general pattern, triggering all the observers when another gets added is a little unexpected. How about adding the logic to BugsnagMetadata.notifyObservers and have it call the new observer only with a new state event?

fractalwrench: Updated to invoke the `observer` parameter only.

PR #765, opened 2020-07-27
Use strcmp() instead of ==  to compare C strings correctly.

croberts22:
## Goal

In Xcode 12 beta 3, compiling the Bugsnag iOS target causes a compiler error to occur:

```
Result of comparison against a string literal is unspecified (use an explicit string comparison function instead)
```

This pull request relieves that error.

## Design

`strcmp()` is used to compare the contents of two C strings, where as `==` compares the position of two pointers. I assume that in this file, we are wanting to compare of the contents of `kThreadSecondary`.

## Changeset

`==` was replaced with `strcmp()`.

## Tests

I ran the tests locally and they all passed.


croberts22:
Looks like my fork is actually pretty outdated and omits this code entirely. Going to close out. 🙂

PR #766, opened 2020-07-29
[PLAT-4707] Record thread information for unhandled JS errors

fractalwrench:
## Goal

Records thread information for unhandled JS errors. The thread information was not captured for unhandled JS errors when `BSGThreadSendPolicyUnhandledOnly` was set because the native layer used the presence of `BSG_KSCrashTypeUserReported` as the `crashType` to determine whether an error was unhandled instead.

This change alters the `captureThreadTrace` method to calculate whether threads should be captured or not, and sets `threadTracingEnabled` appropriately. By default this value is set when installing KSCrash - for handled errors this value is overwritten.

Additionally the crash context has been set to a local struct, which avoids the scenario where multiple notify calls overwrite a global struct.

Note: sending a handled error will overwrite the crash context currently which means the thread send behaviour is incorrect for unhandled errors sent after handled errors. This will be addressed in a separate PR which avoids reusing the crash context.

## Tests
Covered by existing E2E test coverage, additionally verified in an Objective C example app against handled + unhandled errors.

## References

* PLAT-4707


Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1034,6 +1034,7 @@ void bsg_kscrw_i_writeThread(const BSG_KSCrashReportWriter *const writer,
  * @param writeNotableAddresses whether notable addresses should be written
  * so additional information about the error can be extracted
  * only the main thread's stacktrace is serialized.
+ * @param recordAllThreads whether all threads should be recorded, or just the current one
```
tomlongridge, line 1037: Remove param doc

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1702,17 +1699,17 @@ char *bsg_kscrw_i_captureThreadTrace(const BSG_KSCrash_Context *crashContext) {
 
 void bsg_kscrw_i_writeTraceInfo(const BSG_KSCrash_Context *crashContext,
                                 const BSG_KSCrashReportWriter *writer) {
-    bool unhandledCrash = crashContext->crash.crashType != BSG_KSCrashTypeUserReported;
+    const BSG_KSCrash_SentryContext *crash = &crashContext->crash;
 
     // Don't write the binary images for user reported crashes to improve performance
-    if (crashContext->crash.writeBinaryImagesForUserReported == true || unhandledCrash) {
+    if (crash->writeBinaryImagesForUserReported == true || crashContext->crash.crashType != BSG_KSCrashTypeUserReported) {
```
tomlongridge, line 1705:
```suggestion
    if (crash->writeBinaryImagesForUserReported == true || crash.crashType != BSG_KSCrashTypeUserReported) {
```

Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry.h
```diff
@@ -60,7 +60,7 @@ typedef struct BSG_KSCrash_SentryContext {
      * The methodology used for tracing threads.
      * The value will be equal to an enum value from BSGThreadSendPolicy
```
tomlongridge, line 61:
```suggestion
     * If true, will capture traces for all running threads
```

Bugsnag/BugsnagCrashSentry.m
```diff
@@ -30,8 +30,7 @@ - (void)install:(BugsnagConfiguration *)config
     ksCrash.introspectMemory = YES;
     ksCrash.onCrash = onCrash;
     ksCrash.maxStoredReports = BSG_MAX_STORED_REPORTS;
-    ksCrash.threadTracingEnabled = (int) config.sendThreads;
-    
+
```
tomlongridge, line 33: It looks like default values are used inside KSCrash's `initWithReportFilesDirectory` and they are overridden with Bugsnag config values here.

Can the changes made to `BSG_KSCrash.m` therefore be made here instead?

```suggestion
ksCrash.threadTracingEnabled = config.sendThreads != BSGThreadSendPolicyNever;
```

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.h
```diff
@@ -117,7 +117,10 @@
  * @param depth the number of frames to discard from the main thread's stacktrace
  * @return an array of BugsnagThread
  */
-- (NSArray<BugsnagThread *> *)captureThreads:(NSException *)exc depth:(int)depth;
+- (NSArray<BugsnagThread *> *)captureThreads:(NSException *)exc
+                                       depth:(int)depth
+                                   unhandled:(BOOL)unhandled
+                                 sendThreads:(BSGThreadSendPolicy)sendThreads;
```
tomlongridge, line 123: If we make the callers in BugsnagClient compute recordAllThreads and make this the only extra parameter to captureThreads, it avoids KSCrash needing to know anything about BSGThreadSendPolicy values and simplifies the logic as the `notifyInternal` route will always be `sendThreads != never` and the calculation in `collectThreads` is easier to understand in the context of RN.

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashC.c
```diff
@@ -218,7 +218,9 @@ void bsg_kscrash_setThreadTracingEnabled(bool threadTracingEnabled) {
 }
 
 char *bsg_kscrash_captureThreadTrace(int discardDepth, int frameCount, uintptr_t *callstack, const bool recordAllThreads) {
-    BSG_KSCrash_Context *context = crashContext();
+    BSG_KSCrash_Context *globalContext = crashContext();
+    BSG_KSCrash_Context *context = malloc(sizeof(BSG_KSCrash_Context));
```
tomlongridge, line 222: This requires a `free` before returning and a guard for `malloc` returning null (we just return null in that case).

Bugsnag/Client/BugsnagClient.m
```diff
@@ -546,6 +546,11 @@ - (void)start {
     [self.crashSentry install:self.configuration
                     apiClient:self.errorReportApiClient
                       onCrash:&BSSerializeDataCrashHandler];
+    // overridden elsewhere for handled errors, so we can assume that this only
+    // applies to unhandled errors
+    BSGThreadSendPolicy sendThreads = self.configuration.sendThreads;
+    [BSG_KSCrash sharedInstance].threadTracingEnabled = sendThreads == BSGThreadSendPolicyAlways
+                                                     || sendThreads == BSGThreadSendPolicyUnhandledOnly;
```
tomlongridge, line 553: As per previous comment, can this be done in `BSG_KSCrash.m` with all the other values that are overridden by our own configuration?

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashC.c
fractalwrench: Updated this to create a `localContext` variable on the stack instead, as this avoids the need for `malloc` entirely

Bugsnag/BugsnagCrashSentry.m
fractalwrench: Updated to set `threadTracingEnabled` here

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.h
```diff
@@ -117,7 +117,9 @@
  * @param depth the number of frames to discard from the main thread's stacktrace
  * @return an array of BugsnagThread
```
tomlongridge, line 118: Add param to docs

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
```diff
@@ -38,9 +38,10 @@
 #import "BSG_KSLogger.h"
 #import "BugsnagThread.h"
 #import "BSGSerialization.h"
-#import "BugsnagErrorReportSink.h"
+#import "Bugsnag.h"
 #import "BugsnagCollections.h"
 #import "BSG_KSCrashReportFields.h"
+#import "Private.h"
```
tomlongridge, line 45: Can/should these header changes be reverted now?

fractalwrench: Removed unnecessary header

PR #779, opened 2020-08-17
Next release

fractalwrench:


CHANGELOG.md
```diff
@@ -1,10 +1,29 @@
...
+
+* Call react native observer with initial user/context values
+  [#768](https://github.com/bugsnag/bugsnag-cocoa/pull/768)
+
+* Respect bundle version set from config
+  [#762](https://github.com/bugsnag/bugsnag-cocoa/pull/762)
+
+* Populate app/device metadata fields for RN errors
+  [#763](https://github.com/bugsnag/bugsnag-cocoa/pull/763)
+
```
tomlongridge, line 26: These have been added to 6.1.2 section - move up?

CHANGELOG.md
```diff
@@ -1,10 +1,29 @@
 Changelog
 =========
 
+## 6.1.3 (2020-08-17)
+
+### Bug fixes
+
+* Persist user when set on client
```
tomlongridge, line 8: I believe these changes only affect React Native? If so, we should prefix them (e.g. "(React Native) Persist user when set on client") to make it clear the issue doesn't affect vanilla Cocoa.

CHANGELOG.md
```diff
@@ -1,10 +1,29 @@
...
+  [#770](https://github.com/bugsnag/bugsnag-cocoa/pull/770)
+
 ## 6.1.2 (2020-07-21)
 
 ### Bug fixes
 
+* Record thread information for unhandled JS errors
```
tomlongridge, line 15: This PR does affect vanilla Cocoa - the sharing of the crash context is probably worth highlighting as it's not obvious from the PR name.

CHANGELOG.md
fractalwrench: That's not the case for this particular change but I've updated the ones which are RN specific as requested.

CHANGELOG.md
fractalwrench: Reworded description

PR #781, opened 2020-08-18
[PLAT-4799] Allow `EXC_BREAKCRUMB` for `errorClass` in swift assertion test

Cawllec:
## Goal
This will stop the flake caused by an unexpected change in response for a failed assertion in swift with the a later minor release of iOS.

### DIFF
--- a/features/crashprobe.feature
+++ b/features/crashprobe.feature
@@ -113,7 +113,9 @@ Feature: Reporting crash events
     And I configure Bugsnag for "SwiftAssertion"
     And I wait to receive a request
     Then the request is valid for the error reporting API version "4.0" for the "iOS Bugsnag Notifier" notifier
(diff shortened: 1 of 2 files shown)


features/crashprobe.feature
```diff
@@ -113,7 +113,9 @@ Feature: Reporting crash events
     And I configure Bugsnag for "SwiftAssertion"
     And I wait to receive a request
     Then the request is valid for the error reporting API version "4.0" for the "iOS Bugsnag Notifier" notifier
-    And the exception "errorClass" equals "Fatal error"
+    And the exception "errorClass" equals one of:
```
fractalwrench, line 116: It'd be good to include a comment referencing the ticket number which will ultimately provide a more long-term fix for this

PR #783, opened 2020-08-20
Make presence of space optional

twometresteve:
## Goal

Fix a test failure seen on iOS 10

## Design

I'm not sure if the optional space is strictly needed, as I don't know if we ever see the value with a space in tests.  But it seemed safest to make it that way in terms of further failure - and harmless enough in any case.

## Testing

The current instability on iOS 10 may mean that we don't see the fix in action right away, but I have tested the concept in an `irb` session:
```
require 'test/unit'
include Test::Unit::Assertions
regex = Regexp.new(" ?match")
assert_match(regex, "match")
=> nil
assert_match(regex, " match")
=> nil
```
(`nil` if good - an exception results when it fails)


Cawllec (approved):
Pending CI passing

PR #784, opened 2020-08-21
Avoid crash reporting OOMs if nil provided as codeBundleId

fractalwrench:
## Goal

Avoids a crash reporting OOMs if nil provided as codeBundleId. This was caused by attempting to add nil to an `NSDictionary`.


tomlongridge (approved):

ichina:
please bump the framework version with this fix

PR #785, opened 2020-08-25
[PLAT-3907] Add timeout to reset to avoid duplicate payloads

Cawllec:
## Goal
Attempts to address duplicate payload issues using a timeout before each application reset.
This has already been implemented in Maze-runner, this change uses the newly added method.


twometresteve (approved):
LGTM

PR #791, opened 2020-09-04
JSON codec wrapper to handle exceptions

kstenerud:
## Goal

NSJSONSerialization can sometimes throw an exception, even though the API has an NSError field. We should have a unified way to handle JSON without having to remember to put in try/catch blocks, or manually go through our data to ensure it conforms before attempting to encode, or worry about bad data causing the decoder to freak out.

## Design

BSGJSONSerialization wraps all NSJSONSerialization calls in a try/catch block, turning any resulting exceptions into an NSError.

## Changeset

* Added BSGJSONSerialization
* Updated instances of NSJSONSerialization to use BSGJSONSerialization instead.

## Testing

* Added tests for all components that (de)serialize JSON to ensure they work as expected.

### DIFF
--- a/Tests/BugsnagClientTests.m
+++ b/Tests/BugsnagClientTests.m
@@ -43,6 +43,8 @@
 
 NSString *BSGFormatSeverity(BSGSeverity severity);
 
+void BSSerializeJSONDictionary(NSDictionary *dictionary, char **destination);
+
 @implementation BugsnagClientTests
 
 /**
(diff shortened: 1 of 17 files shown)


Tests/BugsnagClientTests.m
```diff
@@ -289,4 +291,9 @@ - (NSDictionary *)generateLargeMetadata {
     return dict;
 }
 
+- (void)testBSSerializeJSONDictionary {
+    char* dest = NULL;
+    BSSerializeJSONDictionary(@{@1: @"a"}, &dest);
```
kattrali, line 296: This test doesn't assert anything?

kstenerud: It's just making sure no exceptions bubble up.

kattrali: ah! makes sense. that could probably use a comment for future travelers.

kattrali (approved):

PR #794, opened 2020-09-08
Exposing the BSG_KSCrash type

gpambrozio:
## Goal

Bugsnag collects stack traces automatically when an issue is reported. For some situations however clients might want to supply their own stack traces. Even though the `BugsnagEvent` exposes a `threads` variable that can be populated, there's no easy way to create the proper array.

This PR just re-organizes some headers to make the `BSG_KSCrash` type, making it possible for clients to use something like `BSG_KSCrash.sharedInstance().captureThreads(exception, depth: BSGNotifierStackFrameCount, recordAllThreads: true)` to create the appropriate stack trace.

## Design

No special design was necessary

## Changeset

Exposing the `BSG_KSCrash` type to make it possible to capture the appropriate `BugsnagThread` array

## Testing

Using my branch in my pod file I added a call to `BSG_KSCrash.sharedInstance().captureThreads(exception, depth: Int32(BSGNotifierStackFrameCount), recordAllThreads: recordAllThreads)` to my project and used the value returned by it in my `BugsnagEvent`'s `threads`


nickdowell:
Hi @gpambrozio thanks for the PR.

Would you mind giving a bit of context around why you want to populate the `threads` property yourself? It's not a use-case we had in mind, perhaps there is a better way we can address the problem.

gpambrozio:
@nickdowell In a gist, what we do in our project to log errors and warnings is something like this:

```
    func log(type: OSLogType, message: @autoclosure @escaping () -> String) {
        DispatchQueue.global(qos: .background).async {
            Bugsnag.notify(NSException(name: NSExceptionName.genericException, reason: message(), userInfo: [:])) { event in
                // set our own thread here
                return true
            }
        }
    }
```

One of the reasons why we dispatch async is that the closure that creates `message` can get pretty expensive for us in some cases. But we have more reasons too, the example above is a very simplified version of what we actually have.

Since Bugsnag would capture threads at the point where we call `Bugsnag.notify` the stack trace we get would be irrelevant to the log itself since it happens in the context of the async call, not where the log is actually called.

So with the change in this PR I can do this:

```
    func log(type: OSLogType, message: @autoclosure @escaping () -> String) {
        let exception = NSException(name: NSExceptionName.genericException, reason: "", userInfo: [:])
        let threads = BSG_KSCrash.sharedInstance().captureThreads(exception, depth: BSGNotifierStackFrameCount, recordAllThreads: false) ?? []

        DispatchQueue.global(qos: .background).async {
            Bugsnag.notify(NSException(name: NSExceptionName.genericException, reason: message(), userInfo: [:])) { event in
                event.threads = threads
                return true
            }
        }
    }
```

This captures the threads at the appropriate moment and saves it until the async call happens. It would be better to have `captureThreads` maybe be a simpler function, maybe a static in `Bugsnag`, ideally only with a `recordAllThreads` parameter but this change is the minimal to make it work for me.

Thanks.

nickdowell:
Thanks for the additional info @gpambrozio

Within Bugsnag we will discuss the possibility of adding a formal API for this, and update this PR in due course.

It's unlikely we will opt to expose `BSG_KSCrash` as it's an internal implementation detail and not something we would like to maintain as part of our public API.

gpambrozio:
@nickdowell any updates or ideas on how we can solve this? We had to update our fork to use the latest version of the library and now we're having issues with some cases not capturing the stack trace.

nickdowell:
Hi @gpambrozio sorry for going so long without a response.

There is way to solve your original problem (setting the message asynchronously) without needing to use any of our internal APIs or capture thread states yourself.

`BugsnagEvent` has an `orignalError` property that normally contains the `NSError` or `NSException` object passed to `notify()`, but since it's of type `id` / `Any?` it can be used to attach arbitrary objects (including closures!) to then use in an `OnSendError` callback.

You can then call `notify()`  synchronously (without computing the message up front) and the correct stack traces will be captured.

For example:

```swift
configuration.addOnSendError { event in
    if let message = event.originalError as? () -> String {
        event.errors.first?.errorMessage = message()
    }
    return true
}

func log(type: OSLogType, message: @autoclosure @escaping () -> String) {
    Bugsnag.notify(NSException(name: NSExceptionName.genericException, reason: "")) { event in
        event.originalError = message
        return true
    }
}
```

Hope this helps!

gpambrozio:
@nickdowell I got an email with a suggestion but I see you removed it. Any reason? Would your suggestion still work? I was gonna share that with a co-worker that might work on the issue. Thanks.

nickdowell:
@gpambrozio sorry about that, I removed it once I realised the suggested approach won't work for events where delivery needs to be retried. This is because we store the original state of the event (before OnSend) when persisting to disk and cannot persist originalError since it can be of any type.

gpambrozio:
@nickdowell would you then reconsider what we're suggesting or maybe suggest a different approach. At the moment we have to maintain our own fork because of this limitation.

nickdowell:
Hi @gpambrozio a simpler alternative to maintaining a branch of our repo would be to expose our internal threads API in an Objective-C bridging header:

```Objective-C
// AwesomeApp-Bridging-Header.h

#import <Bugsnag/Bugsnag.h>

NS_ASSUME_NONNULL_BEGIN

@interface BugsnagThread ()

+ (NSArray<BugsnagThread *> *)allThreads:(BOOL)allThreads
                callStackReturnAddresses:(NSArray<NSNumber *> *)callStackReturnAddresses;

@end

NS_ASSUME_NONNULL_END
```

Allowing your Swift code to capture the threads before calling `notify()` (as you're already doing):

```Swift
let threads = BugsnagThread.allThreads(true, callStackReturnAddresses: Thread.callStackReturnAddresses)

// ...

DispatchQueue.global(qos: .background).async {
    Bugsnag.notify(NSException(name: NSExceptionName.genericException, reason: message(), userInfo: [:])) { event in
        event.threads = threads
        return true
    }
}
```

This does of course run the risk of breaking your build if we remove that API in a future release.

PR #796, opened 2020-09-09
Replace the entire observers list when adding/removing observers in o…

kstenerud:
## Goal

Fixes https://bugsnag.atlassian.net/browse/PLAT-5007

## Design

Instead of adding/removing observers from the observer list directly, make a copy of the list, modify the copy, then replace the original. This makes concurrency issues (where an observer gets added/removed while the observer list is being walked) impossible.

So long as the list remains small (100 or so entries), the CPU cost is negligible. We don't expect to ever see more than around 10 observers here no matter what future features we add.

### DIFF
--- a/Bugsnag/Metadata/BugsnagMetadata.m
+++ b/Bugsnag/Metadata/BugsnagMetadata.m
@@ -31,7 +31,7 @@
 #import "BugsnagStateEvent.h"
 
 @interface BugsnagMetadata ()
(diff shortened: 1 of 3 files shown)


kattrali (commented):
Is there a test we can add here to demonstrate the problem in the previous implementation?

kstenerud:
Added a test that deliberately accesses on two threads.

fractalwrench (approved):
LGTM

kattrali (approved):

PR #801, opened 2020-09-14
Add a high performance, async-safe key-value store for use in time sensitive code

kstenerud:
The idea is to minimize the chances of a KV write operation getting aborted if the app happens to crash or shut down, by making use of syscalls for the heavy lifting (syscalls can't be interrupted). The majority of the work (and thus CPU time) happens in the kernel, giving a very good chance that the write will always succeed.

Code is async-safe, idempotent, 100% C and syscalls.

Tests are also included.

Note: This code is 100% standalone. I'll add integration with the OOM system in a later PR.

Fixes https://bugsnag.atlassian.net/browse/PLAT-5003


Bugsnag/Helpers/BugsnagKVStore.c
```diff
@@ -0,0 +1,143 @@
...
+    *err = 0;
+}
+
+void bsgkv_setBytes(const char* key, const uint8_t* value, int length, int* err) {
+    int fd = openKeyWrite(key);
+    if(fd < 0) {
+        *err = errno;
+        return;
+    }
+    if(write(fd, value, length) < 0) {
```
kattrali, line 88: Shouldn't this check if the number of bytes written is equal to `length`? It seems possible that bytes written could be less than the intended amount.

Tests/BugsnagKVStoreTest.m
```diff
@@ -0,0 +1,167 @@
...
+    
+    bsgkv_delete(name, &err);
+    XCTAssertEqual(err, 0);
+
+    bsgkv_setString(name, expected, &err);
+    XCTAssertEqual(err, 0);
+
+    bsgkv_getString(name, actual, sizeof(actual), &err);
+    XCTAssertEqual(err, 0);
+    XCTAssertEqual(strcmp(actual, expected), 0);
```
kattrali, line 95: It would be helpful for future travelers (and potential CI breakage) to print something more descriptive about the failure, like the expected and actual values, particularly since `strcmp()` will fall over in the event of `actual` being NULL.

Maybe an assertion wrapper? Something like:

```c
#define AssertEqualChars(actual, expected) \
    actual == NULL \
    ? XCTFail("Expected '%s', got NULL", expected) \
    : XCTAssertEqual(strcmp(actual, expected), 0, "Expected '%s', got '%s'", expected, actual)
```

Bugsnag/Helpers/BugsnagKVStore.c
kstenerud: The only situations that would cause less bytes to be written are:
- Error (returns -1)
- The device ran out of space (so we're screwed no matter what and this write won't matter in the grand scheme of things)
- The call got interrupted by a signal handler. 4.2BSD introduced automatic restart, which will cause certain syscalls (such as `write`) to be re-sent automatically. Darwin is based on 4.4BSD so it has this functionality.

But then again, if we wanted to use this in other operating systems, it would make sense to put in the EINTR handler code anyway...

kattrali (approved):

PR #806, opened 2020-09-16
[PLAT-3596] Fix reporting of mach exception code and subcode

nickdowell:
## Goal

This change brings in some improvements to mach exception handling that were made in KSCrash

* https://github.com/kstenerud/KSCrash/pull/349
* https://github.com/kstenerud/KSCrash/pull/350

These changes fix the parsing of mach exception `code` and `subcode` values.

Prior to these changes, for a `EXC_BAD_ACCESS` exception, the `subcode` did not contain the expected memory address. The exception handler was expecting the kernel to send 64-bit values, but 32-bit values were being sent instead because the `MACH_EXCEPTION_CODES` flag was not specified when registering the exception handler. There was also a struct padding issue (fixed by adding `#pragma pack(4)`) that was causing the exception handler to read values from the wrong offset, resulting in the memory address coming through as `code` and `subcode` to always contain the value `0x8`.

## Testing

I verified manually, using the example app, that the `EXC_BAD_ACCESS` mach exception is reported to the dashboard. The dashboard is now correctly identifying and displaying the memory address that the app tried to access.

Screenshot 2020-09-18 at 08 48 54

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
@@ -143,8 +143,7 @@ void bsg_kscrw_i_addIntegerElement(const BSG_KSCrashReportWriter *const writer,
 void bsg_kscrw_i_addUIntegerElement(const BSG_KSCrashReportWriter *const writer,
                                     const char *const key,
                                     const unsigned long long value) {
-    bsg_ksjsonaddIntegerElement(bsg_getJsonContext(writer), key,
-                                (long long)value);
+    bsg_ksjsonaddUIntegerElement(bsg_getJsonContext(writer), key, value);
 }
 
 void bsg_kscrw_i_addStringElement(const BSG_KSCrashReportWriter *const writer,
@@ -1181,8 +1180,8 @@ void bsg_kscrw_i_writeError(const BSG_KSCrashReportWriter *const writer,
                             const char *const key,
                             const BSG_KSCrash_SentryContext *const crash) {
     int machExceptionType = 0;
-    kern_return_t machCode = 0;
-    kern_return_t machSubCode = 0;
+    int64_t machCode = 0;
+    int64_t machSubCode = 0;
     int sigNum = 0;
     int sigCode = 0;
     const char *exceptionName = NULL;
@@ -1192,14 +1191,14 @@ void bsg_kscrw_i_writeError(const BSG_KSCrashReportWriter *const writer,
     switch (crash->crashType) {
     case BSG_KSCrashTypeMachException:
         machExceptionType = crash->mach.type;
-        machCode = (kern_return_t)crash->mach.code;
+        machCode = crash->mach.code;
         if (machCode == KERN_PROTECTION_FAILURE && crash->isStackOverflow) {
             // A stack overflow should return KERN_INVALID_ADDRESS, but
             // when a stack blasts through the guard pages at the top of the
             // stack, it generates KERN_PROTECTION_FAILURE. Correct for this.
             machCode = KERN_INVALID_ADDRESS;
         }
-        machSubCode = (kern_return_t)crash->mach.subcode;
+        machSubCode = crash->mach.subcode;
 
         sigNum =
             bsg_kssignal_signalForMachException(machExceptionType, machCode);
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_MachException.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_MachException.c
@@ -45,12 +45,19 @@
 #define kThreadPrimary "KSCrash Exception Handler (Primary)"
 #define kThreadSecondary "KSCrash Exception Handler (Secondary)"
 
+#if __LP64__
+    #define MACH_ERROR_CODE_MASK 0xFFFFFFFFFFFFFFFF
+#else
+    #define MACH_ERROR_CODE_MASK 0xFFFFFFFF
+#endif
+
 // ============================================================================
 #pragma mark - Types -
 // ============================================================================
 
 /** A mach exception message (according to ux_exception.c, xnu-1699.22.81).
  */
+#pragma pack(4)
 typedef struct {
     /** Mach header. */
     mach_msg_header_t header;
@@ -87,9 +94,11 @@ typedef struct {
     /** Padding to avoid RCV_TOO_LARGE. */
     char padding[512];
 } MachExceptionMessage;
+#pragma pack()
 
 /** A mach reply message (according to ux_exception.c, xnu-1699.22.81).
  */
+#pragma pack(4)
 typedef struct {
     /** Mach header. */
     mach_msg_header_t header;
@@ -100,6 +109,7 @@ typedef struct {
     /** Return code. */
     kern_return_t returnCode;
 } MachReplyMessage;
+#pragma pack()
 
 // ============================================================================
 #pragma mark - Globals -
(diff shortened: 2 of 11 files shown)


Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -143,8 +143,7 @@ void bsg_kscrw_i_addIntegerElement(const BSG_KSCrashReportWriter *const writer,
 void bsg_kscrw_i_addUIntegerElement(const BSG_KSCrashReportWriter *const writer,
                                     const char *const key,
                                     const unsigned long long value) {
-    bsg_ksjsonaddIntegerElement(bsg_getJsonContext(writer), key,
-                                (long long)value);
+    bsg_ksjsonaddUIntegerElement(bsg_getJsonContext(writer), key, value);
```
kstenerud, line 146: This was originally cast to signed because JSON technically doesn't support 64-bit unsigned values. But this was written before I learned that technically JSON doesn't even support 63-bit signed values (it's actually 54 bits, the coefficient size in 64-bit ieee754 floats) :P.

Regardless, we need to be sure that upstream processors (with each language's different JSON implementations) are able to handle this change. I believe that some of them always assume signed ints.

Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_MachException.c
```diff
@@ -45,12 +45,19 @@
 #define kThreadPrimary "KSCrash Exception Handler (Primary)"
 #define kThreadSecondary "KSCrash Exception Handler (Secondary)"
 
+#if __LP64__
```
kstenerud, line 48: We probably don't even need this kind of thing anymore. Are there any 32-bit targets that are supported by the latest Xcode?

kstenerud (changes requested):
Please make sure the unsigned ints are safe for other languages to ingest (in particular, with the 64th bit set).

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
nickdowell: Based on my reading of [the spec](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf), I don't think JSON imposes any size limits on what may be stored.

> JSON is agnostic about the semantics of numbers. In any programming language, there can be a variety of number types of various capacities and complements, fixed or floating, binary or decimal. That can make interchange between different programming languages difficult. JSON instead offers only the representation of numbers that humans use: a sequence of digits. All programming languages know how to make sense of digit sequences even if they disagree on internal representations. That is enough to allow interchange.

That said some parsers will inevitably get this wrong and impose their own limits.

If it is the case that our upstream processors don't support 64-bit unsigned values, then it doesn't make sense to incorporate these changes from KSCrash.

Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_MachException.c
nickdowell: The latest Xcode does not support 32-bit targets, but our [README](https://github.com/bugsnag/bugsnag-cocoa/blob/master/README.md) states we support iOS 9.3+, macOS 10.11+ and tvOS 9.2+, which still include 32-bit support 🤔

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
kstenerud: I'd actually be fine with the changes, so long as we can verify that our pipeline can handle it upstream.

Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_MachException.c
kstenerud: But if Xcode doesn't support 32-bit, then nobody can compile 32-bit applications anymore for submission to the Apple stores (they only accept submissions compiled on the latest Xcode). Maybe we should update our supported targets list?

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
kstenerud: Also, we need to check if our upstream code expects signed ints and tries to convert to unsigned, considering that's how the source JSON was formatted until now...

Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_MachException.c
nickdowell: Agreed, all iOS developers must be using the latest Xcode to get on the App Store. It's possible there could be a few developers using old versions of Xcode to build with old macOS SDKs.

Perhaps it's worth us documenting which Xcode versions we support building with?

nickdowell: Resolving this as it's been raised as a separate issue ROAD-1117

examples/objective-c-ios/Bugsnag Test App/ViewController.m
```diff
@@ -63,7 +63,8 @@ - (IBAction)generateSignal:(id)sender {
  This method causes a low-level exception from the operating system to terminate the app.  Upon reopening the app this signal should be notified to your Bugsnag dashboard.
  */
 - (IBAction)generateMachException:(id)sender {
-    void (*ptr)(void) = NULL;
+    // This should result in an EXC_BAD_ACCESS mach exception with code = KERN_INVALID_ADDRESS and subcode = 0xDEADBEEF
+    void (* ptr)(void) = (void *)0xDEADBEEF;
```
kattrali, line 67: The test case for mach exceptions should be updated to verify this behavior as well and assert that the recorded code and subcode equal the same values.

* [test](https://github.com/bugsnag/bugsnag-cocoa/blob/9660b880a97a57fbacb77b10a4b23b00b3104d56/features/unhandled_mach_exception.feature)
* [fixture](https://github.com/bugsnag/bugsnag-cocoa/blob/9660b880a97a57fbacb77b10a4b23b00b3104d56/features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/UnhandledMachExceptionScenario.m)

If the test server doesn't handle the unsigned 64-bit values correctly, then we should look at stringifying them before merging this down.

Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_MachException.c
kstenerud: New info: Nick discovered that the watches still use armv7 chips (32-bit). So we can't drop 32-bit support yet, especially given the newer watches with networking support.

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
nickdowell: Opened PLAT-5084 to cover the pipeline / dashboard handling

examples/objective-c-ios/Bugsnag Test App/ViewController.m
nickdowell: The unhandled mach exception test case has been updated to verify the code and subcode, and has now passed successfully 👍

kattrali (approved):

kstenerud (approved):

PR #810, opened 2020-09-17
docs(upgrading): Depth info moved to docs

xander-jones:
## Goal
Initially raised in https://github.com/bugsnag/bugsnag-cocoa/pull/808. This information is better suited to the main site docs.bugsnag.com, so will be moved here.

This PR is not to be merged until docs PR is merged.

## Design
N/A

## Changeset
* `UPGRADING.md`

## Testing
N/A


kattrali:
keeping this change since the upgrade guide should show how to migrate this functionality when migrating from v5 -> v6

PR #812, opened 2020-09-21
[PLAT-5000] React Native promise rejection stack trace processing

nickdowell:
## Goal

This change allows BugsnagReactNative to convert the native promise rejection stack trace provided by React Native into an array of BugsnagStackframe objects, so that it can be captured as a BugsnagError.

## Design

See ROAD-801 for details of the design and specification.

## Changeset

`[BugsnagStackframe stackframesWithCallStackSymbols:]` was added, which will be invoked by BugsnagReactNative.

## Testing

Unit test cases have been added to verify the processing of passed in call stacks into BugsnagStackframe objects.

End-to-end automated tests will be covered in bugsnag-js

### DIFF
--- a/Bugsnag/Payload/BugsnagStackframe.m
+++ b/Bugsnag/Payload/BugsnagStackframe.m
@@ -7,8 +7,12 @@
 //
 
 #import "BugsnagStackframe.h"
-#import "BugsnagKeys.h"
+
+#import "BSG_KSBacktrace.h"
+#import "BSG_KSDynamicLinker.h"
 #import "BugsnagCollections.h"
+#import "BugsnagKeys.h"
+#import "BugsnagLogger.h"
 
 @implementation BugsnagStackframe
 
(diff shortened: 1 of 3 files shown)


Bugsnag/Payload/BugsnagStackframe.m
```diff
@@ -66,6 +70,90 @@ + (BugsnagStackframe *)frameFromDict:(NSDictionary *)dict
...
+            continue;
+        }
+        NSString *frameNumber = [string substringWithRange:[match rangeAtIndex:1]];
+        NSString *imageName = [string substringWithRange:[match rangeAtIndex:2]];
+        NSString *frameAddress = [string substringWithRange:[match rangeAtIndex:3]];
+        NSRange symbolNameRange = [match rangeAtIndex:5];
+        NSString *symbolName = nil;
+        if (symbolNameRange.location != NSNotFound) {
+            symbolName = [string substringWithRange:symbolNameRange];
+        }
```
kattrali, line 109: `symbolName` should fall back to `frameAddress` when there is no match. It can also be asserted in the dummy symbols test.

Tests/BugsnagStackframeTest.m
```diff
@@ -93,4 +95,65 @@ - (void)testInvalidFrame {
...
+    stackframes = [BugsnagStackframe stackframesWithCallStackSymbols:@[
+        @"0   ReactNativeTest                     0x000000010fda7f1b RCTJSErrorFromCodeMessageAndNSError + 79",
+        @"1   ReactNativeTest                     0x000000010fd76897 __41-[RCTModuleMethod processMethodSignature]_block_invoke_2.103 + 97",
+        @"2   ReactNativeTest                     0x000000010fccd9c3 -[BenCrash asyncReject:rejecter:] + 106",
+        @"3   CoreFoundation                      0x00007fff23e44dec __invoking___ + 140",
+        @"4   CoreFoundation                      0x00007fff23e41fd1 -[NSInvocation invoke] + 321",
+        @"5   CoreFoundation                      0x00007fff23e422a4 -[NSInvocation invokeWithTarget:] + 68",
+        @"6  ReactNativeTest                     0x000000010fd76eae -[RCTModuleMethod invokeWithBridge:module:arguments:] + 578",
+        @"7 ReactNativeTest                     0x000000010fd79138 _ZN8facebook5reactL11invokeInnerEP9RCTBridgeP13RCTModuleDatajRKN5folly7dynamicE + 246"]];
+    XCTAssertEqual(stackframes.count, 8);
```
kattrali, line 121: This test should additionally assert that the frame contents are the expected values. In the case that the behavior regresses in some way in the future, it seems much more likely that a mistake would be made to accidentally change a value rather than remove or add a frame.

kattrali (changes requested):

Bugsnag/Payload/BugsnagStackframe.m
nickdowell: Fixed in 3cd17800795c4be1d1b9938f5c81ede997cfd3df

Tests/BugsnagStackframeTest.m
nickdowell: Fixed in e0e1b570751eb98970fa1ebacc76b5b2e9047f00

kattrali (approved):

Bugsnag/Payload/BugsnagStackframe.m
```diff
@@ -66,6 +70,90 @@ + (BugsnagStackframe *)frameFromDict:(NSDictionary *)dict
     }
 }
 
++ (NSArray<BugsnagStackframe *> *)stackframesWithCallStackSymbols:(NSArray<NSString *> *)callStackSymbols {
+    NSString *pattern = (@"^(\\d+)"             // Capture the leading frame number
+                         @" +"                  // Skip whitespace
+                         @"(\\S+)"              // Image name
```
JoeLago, line 76: Hi, sorry if it's inappropriate to dig up an old pull request like this. I was trying to use this function to rebuild a stacktrace with some `callStackSymbols` I cached for an unhandled error and I believe I'm losing my apps frames because there is a space in our apps display name. For Example:

```
"0   ReactNative Test                     0x000000010fda7f1b RCTJSErrorFromCodeMessageAndNSError + 79"
```

nickdowell: Thanks for reporting this - we are working on a fix: https://github.com/bugsnag/bugsnag-cocoa/pull/1036

PR #821, opened 2020-09-24
[PLAT-5129 ] Fix reporting of user events with multiple errors

nickdowell:
## Goal

This change is necessary for the correct reporting of native module promise rejections from react native.

Although the crash reports received by the Cocoa notifier and stored to disk contained the native stack trace, the reports sent to the back-end were missing the native error.

## Changeset

`-[BugsnagEvent toJson]` was ignoring the errors array if `user.overrides.customStacktraceFrames` and `user.overrides.customStacktraceType` were set in the payload, which is the case for RN native promise rejections.

This was not a problem when events were only expected to have a single error, but RN native promise rejections may have a second error that represents the native stack.

This change ensures the customException only overrides the first error, leaving the native one alone.

## Testing

I pushed a version of this code to bugsnag-js in order to see the effect on the e2e tests. This change allowed `Scenario: Handled JS error with native stacktrace` to pass the `And the payload field "events.0.exceptions" is an array with 2 elements` check and several others, although changes to the scenario will be required to cater for the platform-specific format of Cocoa crash reports  - https://buildkite.com/bugsnag/at-bugsnag-react-native/builds/796#36f69338-2aaf-4436-a5f1-563642c3a4f1/458

A unit test case was added to this project, which fails without the change to `-[BugsnagEvent toJson]` but passes with it.


kstenerud (approved):

fractalwrench (approved):
LGTM

PR #829, opened 2020-10-02
Remove extraneous severity info from unit test

kstenerud:
## Goal

"user.state.crash.severity" is neither created nor used in the cocoa client, so it shouldn't be in the test.

## Design

Removed "user.state.crash.severity" from dictionary in `testHandledReportSeverity`

## Testing

Re-ran unit tests on ios, macos, tvos.

### DIFF
--- a/Tests/BugsnagEventTests.m
+++ b/Tests/BugsnagEventTests.m
@@ -389,7 +389,7 @@
 - (void)testHandledReportSeverity {
     // handled reports should use the serialised depth
     BugsnagHandledState *state = [BugsnagHandledState handledStateWithSeverityReason:HandledException];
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## TBD
+
+### Enhancements
+
+* Removed extraneous data from severity unit test.
```
kattrali, line 8: This doesn't need a changelog entry since there's no change in library behavior

kattrali (changes requested):
the test fix looks good to me

kattrali (approved):

PR #835, opened 2020-10-07
Release v6.2.0

kstenerud:


CHANGELOG.md
```diff
@@ -1,6 +1,23 @@
 Changelog
 =========
 
+## 6.1.8 (2020-10-07)
+
+### Enhancements
+
+* An invalid or missing API key now causes an NSException to be thrown upon
+  calling [BugsnagClient start].
```
tomlongridge, line 9:
```suggestion
  calling [Bugsnag start].
```

tomlongridge: Also, I think we should probably say a bit more about the change so it's easier for people to find details – i.e. the check is moved from creation the configuration object.

CHANGELOG.md
```diff
@@ -1,6 +1,23 @@
 Changelog
 =========
 
+## 6.1.8 (2020-10-07)
+
+### Enhancements
```
tomlongridge, line 6: Are these more bug fixes than enhancements? If we classify them as enhancements, we should make this 6.2.0.

tomlongridge (approved):

PR #836, opened 2020-10-08
[PLAT-5200] Fix intermittent test failures in features/config_from_plist.feature

nickdowell:
## Goal

The test scenarios in `features/config_from_plist.feature` were intermittently failing because the OOM reports were not always sent before the crash reports.

These tests were never intended to be run with OOMs enabled, but the recent changes to allow OOM detection in Debug builds started these being reported after the app is killed by Appium / Browserstack (the e2e tests are run against a [Debug](https://github.com/bugsnag/bugsnag-cocoa/blob/master/features/scripts/export_ios_app.sh#L16) build of the app.)

The test app's `Info.plist` contained keys that were attempting to disable OOMs, but these keys are not actually supported by the notifier. This had gone unnoticed because the notifier was not logging any warnings for unsupported keys, and OOMs were previously disabled in Debug builds.

## Changeset

**OOMs are now prevented**

by clearing out the caches directory (where the previous session state is stored) when a scenario invokes "Given I clear all UserDefaults data" - maybe this should be renamed accordingly?

**Info.plist config keys are now checked.**

A warning will now be logged if there are any unknown config keys in the Bugsnag dictionary, to avoid unexpected behaviour in the future.

## Testing

The OOM disabling mechanism was locally tested by starting and stopping the app from Xcode.

The modified code is covered by e2e tests.

### DIFF
--- a/features/fixtures/ios-swift-cocoapods/iOSTestApp/ViewController.swift
+++ b/features/fixtures/ios-swift-cocoapods/iOSTestApp/ViewController.swift
@@ -44,8 +44,12 @@ class ViewController: UIViewController {
         UserDefaults.standard.removePersistentDomain(forName: Bundle.main.bundleIdentifier!)
         do {
             let cachesDirectory = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask)[0]
-            for url in try FileManager.default.contentsOfDirectory(at: cachesDirectory, includingPropertiesForKeys: []) {
-                try FileManager.default.removeItem(at: url)
+            try FileManager.default.contentsOfDirectory(at: cachesDirectory, includingPropertiesForKeys: []).forEach {
+                do {
+                    try FileManager.default.removeItem(at: $0)
+                } catch {
+                    NSLog("%@", String(describing: error))
+                }
             }
         } catch {
             NSLog("%@", String(describing: error))
--- a/features/fixtures/ios-swift-cocoapods/iOSTestApp/ViewController.swift
+++ b/features/fixtures/ios-swift-cocoapods/iOSTestApp/ViewController.swift
@@ -39,10 +39,10 @@ class ViewController: UIViewController {
         scenario?.startBugsnag()
     }
     
-    @IBAction func clearUserData(_ sender: Any) {
-        NSLog("Clear user defaults")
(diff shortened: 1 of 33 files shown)


kstenerud:
Will we still need OOM-specific code since we can clear cache data in the tests?

nickdowell:
> Will we still need OOM-specific code since we can clear cache data in the tests?

If the cache clearing approach works, this OOM-specific code can be removed 👍

features/fixtures/ios-swift-cocoapods/iOSTestApp/ViewController.swift
```diff
@@ -42,6 +42,18 @@ class ViewController: UIViewController {
     @IBAction func clearUserData(_ sender: Any) {
         NSLog("Clear user defaults")
         UserDefaults.standard.removePersistentDomain(forName: Bundle.main.bundleIdentifier!)
+        do {
```
twometresteve, line 45: I think this is comment of a single line comment: "Delete any existing ..?..".  OOM files, error files, session files, all of the above?

twometresteve (commented):
Will this also stop error/session requests from bleeding in from a previous scenario?  If so (and in any case) this looks like nice neat solution.

features/fixtures/ios-swift-cocoapods/iOSTestApp/ViewController.swift
```diff
@@ -42,6 +42,18 @@ class ViewController: UIViewController {
     @IBAction func clearUserData(_ sender: Any) {
```
twometresteve, line 42: Rename to `clearApplicationState`?

twometresteve, line 42: Rename to `clearApplicationState`?

twometresteve:
> "Given I clear all UserDefaults data" - maybe this should be renamed accordingly?

Yes, please.  "Given I clear all application data" ?  I'm afraid this will involve updating all of the existing `.feature` files, but it would be clear (enough) from looking at them as to what is actually happening.

twometresteve:
LGTM on the tests side.  Happy for this to be merged, but I'll let a Cocoa person give the approval given the smaller change to the notifier itself.

kstenerud (approved):

Issue #832, opened 2020-10-05
Crash in BSG_KSMachHeaders


sethfri:
### Describe the bug
App crashes in `BSG_KSMachHeaders`'s `bsg_mach_headers_cache_unlock` function.

### Steps to reproduce
Unclear, but including stack trace below

### Environment
* Bugsnag version: 6.1.3
* CocoaPods version: 1.9.3
* Carthage version (if any): n/a
* iOS/tvOS/macOS version(s): 12.1.2
* Simulator or physical device: iPhone 6s
* XCode version: 11.2.1
* Swift version (if applicable): 5.1

### Stack trace

<details><summary>Stack trace:</summary>

```
Crashed: com.apple.main-thread
0  ???                            0x23b90df0c (Missing)
1  ???                            0x23b90f4f4 (Missing)
2  MyApp                           0x1056ba354 bsg_mach_headers_cache_unlock + 66 (BSG_KSMachHeaders.m:66)
3  ???                            0x1078443e0 (Missing)
4  ???                            0x23b75176c (Missing)
5  MyApp                           0x1056ba3d4 bsg_mach_headers_register_for_changes + 114 (BSG_KSMachHeaders.m:114)
6  MyApp                           0x1056b02ec bsg_kscrash_install + 107 (BSG_KSCrashC.c:107)
7  MyApp                           0x1056af1e4 -[BSG_KSCrash install] + 230 (BSG_KSCrash.m:230)
8  MyApp                           0x1056cf1cc -[BugsnagCrashSentry install:apiClient:onCrash:] + 50 (BugsnagCrashSentry.m:50)
9  MyApp                           0x1056c6b74 -[BugsnagClient start] + 551 (BugsnagClient.m:551)
10 MyApp                           0x1056c0308 +[Bugsnag startWithConfiguration:] + 90 (Bugsnag.m:90)
11 MyApp                           0x104d9728c -[SCDualCrashReporter setUpDualCrashHandler] + 39 (SCDualCrashReporter.m:39)
12 MyApp                           0x105177cdc AppDelegate.setUp(_:) + 4344249564 (<compiler-generated>:4344249564)
13 MyApp                           0x10517c110 AppDelegate.application(_:willFinishLaunchingWithOptions:) + 647 (AppDelegate.swift:647)
14 MyApp                           0x10517d654 @objc AppDelegate.application(_:willFinishLaunchingWithOptions:) + 4344272468
15 ???                            0x268e91d34 (Missing)
[...]

PR #837, opened 2020-10-09
Change os_unfair_lock calls to use a dispatch queue instead.

kstenerud:
## Goal

https://github.com/bugsnag/bugsnag-cocoa/issues/832 shows a crash in `os_unfair_lock_unlock` that couldn't be traced back to anything user-side. Since we can't repro the issue, we need another way to defend against this.

## Design

`os_unfair_lock` is a relatively new API, so it's not inconceivable that there could still be bugs in it. Dispatch queues, on the other hand, have been battle-hardened over decades. Dispatching to a local queue on a local thread has very low overhead, and will introduce far less latency than calling NSProcessInfo, so this will on the whole run faster.

## Changeset

`bsg_mach_headers_add_image` now uses a dispatch queue rather than `os_unfair_lock` or `OSSpinLock`.

## Testing

Re-ran all unit tests, and did manual tests to ensure end-to-end symbolication.


nickdowell (approved):
Looks good to me

kattrali (approved):

PR #839, opened 2020-10-12
[PLAT-5204] Fix recording of foregroundDuration

nickdowell:
## Goal

This changes the tracking of `foregroundDuration` to include both `UIApplicationStateActive` and `UIApplicationStateInactive` which are both described as "foreground" states in Apple documentation.

Previously only `UIApplicationStateActive` was being counted towards this total duration.

## Design

This simplifies the logic behind duration tracking and makes our APIs more consistent and in line with Apple's documentation.

## Changeset

`BSG_KSCrash_State` now only observes whether the app is in a foreground or background state. Code related to tracking active vs inactive has been removed.

`activeDuration...` struct fields have been renamed to `foregroundDuration...` for clarity

Test cases that only varied the active vs inactive state have been removed.

## Testing

Manually verified that the `duration` and `durationInForeground` values increment as expected in payloads sent to the back-end.

Unit tests and e2e tests passing.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
@@ -327,7 +327,7 @@ IMPLEMENT_EXCLUSIVE_SHARED_INSTANCE(BSG_KSCrash)
     BSG_KSCrash_State state = crashContext()->state;
     bsg_kscrashstate_updateDurationStats(&state);
     NSMutableDictionary *dict = [NSMutableDictionary new];
-    BSGDictSetSafeObject(dict, @(state.activeDurationSinceLaunch), @BSG_KSCrashField_ActiveTimeSinceLaunch);
+    BSGDictSetSafeObject(dict, @(state.foregroundDurationSinceLaunch), @BSG_KSCrashField_ActiveTimeSinceLaunch);
     BSGDictSetSafeObject(dict, @(state.backgroundDurationSinceLaunch), @BSG_KSCrashField_BGTimeSinceLaunch);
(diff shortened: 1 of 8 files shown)


kattrali (commented):
More of a general question but would it be useful to know whether the app was active and for how long in addition to whether it was in the foreground? I could potentially see it as separate metadata in addition to correcting the definition of "in foreground."

kattrali (approved):

kstenerud:
This is probably going to get confusing since we now have the concept of "foreground duration" and "background duration", which leads to the assumption that they're opposite sides of the same state switch when they're not. The temptation will still be there to just add the two durations together in order to get "total app time", which will still be incorrect.

The way it was before, it had duration information for time backgrounded, time active (in fg, not interrupted), but was missing time interrupted and total time. We'll lose information about this kind of ios-specific information with this change, and recovering it in a later version will require changing the meaning of the fields again, and dealing with the headache of legacy versions of bugsnag in the field.

I think we should be very sure we're okay with this before committing to such a change...

nickdowell:
At the [public API level](https://github.com/bugsnag/bugsnag-cocoa/blob/master/Bugsnag/include/Bugsnag/BugsnagAppWithState.h#L19-L33) and in the crash reports we only expose `duration` (total) and `durationInForeground`. Note that there is some ambiguity because we don't define what it means for the app to be "running" - currently we include time spent [suspended](https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle?language=objc) (when the app stops getting CPU time) in the total.

Prior to this change;
* duration = time spent in the active and background UIApplicationStates
* durationInForeground = time spent in the active UIApplicationState

After this change;
* duration = time spent in the active, inactive and background UIApplicationStates
* durationInForeground = time spent in the active and inactive UIApplicationStates

PR #841, opened 2020-10-14
[PLAT-5224] Fix test flakes in enabled_error_types.feature

nickdowell:
## Goal

Occasionally the tests would fail due to a session payload being sent twice.

In these cases the notifier was not receiving the response to its HTTP request before the scenario crashed the app, but the server had received it. When Bugsnag was reinitialised it would re-send the first session since it could not know that it had been already been received by the server.

## Design

By removing the 2 second delay (which is not always enough time to receive the HTTP response) we can be sure the session from the first launch of the app has not been sent, which avoid duplicates.

## Changeset

Removed the delay before crashing the app.

I could not find any comments explaining why the delay had been added in commit d2e04a05ed63bb3ff893d55fd22c5ae2fa536c15

## Testing

I created a separate branch with extra logging and multiple copies of the flakey tests scenario to validate the fix, [results are here](https://buildkite.com/bugsnag/cocoa-bugsnag-notifier/builds/1106)

### DIFF
--- a/features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/EnabledErrorTypesScenario.m
+++ b/features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/EnabledErrorTypesScenario.m
@@ -32,10 +32,8 @@
 
 - (void)run {
     // From null prt scenario
-    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(2 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
-        volatile char *ptr = NULL;
-        (void) *ptr;
-    });
+    volatile char *ptr = NULL;
+    (void) *ptr;
 }
 
 @end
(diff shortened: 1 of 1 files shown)


features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/EnabledErrorTypesScenario.m
```diff
@@ -32,10 +32,8 @@ - (void)startBugsnag {
 
 - (void)run {
     // From null prt scenario
-    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(2 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
-        volatile char *ptr = NULL;
-        (void) *ptr;
-    });
+    volatile char *ptr = NULL;
+    (void) *ptr;
```
kstenerud, line 36: What are these for?

nickdowell: This intentionally crashes the app when the scenario is run from the test script:

```cucumber
  Scenario: All Crash reporting is disabled
    # Sessions: on, unhandled crashes: off
    When I run "DisableAllExceptManualExceptionsAndCrashScenario" and relaunch the app

```

This particular test script then goes on to verify that the crash is not reported by the notifier.

twometresteve (approved):
LGTM.  The approach of forcing both sessions to be sent after relaunch is quite cunning and I would note that they are incidental to the test scenario itself.  The test is trying to prove a negative, with checks on the sessions only really proving that the app really did launch twice.
A better way of writing scenarios like this would be to provide a custom logger that sends out HTTP requests, which MazeRunner could then store and us write assertions on.
We need to do some groundwork refactoring in MazeRunner first to provide that though.

Cawllec (approved):

kstenerud (approved):

PR #842, opened 2020-10-14
[PLAT-4357] Re-Add Out of Memory e2e test scenarios

Cawllec:
## Goal

- Re-establish some OOM scenarios that were removed due to our move to BrowserStack and changes to the OOM triggers.

## Design

While the second commit shows writing the appropriate files to disk, before Bugsnag starts, to trigger the OOM detection this was removed in the final draft to test OOM detection as well as the data present in OOMs and how they relate to configuration options.

Note: This may prevent these scenarios working correctly when running the tests against a local device - this is being investigated.  If this is the case discussion will take place internally to decide whether these scenarios should be amended or skipped whilst running locally.

Note++: The OOMSessionScenario is a placeholder until PLAT-5247 is resolved.  It should be utilised as part of that task to ensure OOM/Session coverage is present.


features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/OOMSessionScenario.swift
```diff
@@ -0,0 +1,23 @@
...
+class OOMSessionScenario: Scenario {
+
+    override func startBugsnag() {
+        config.enabledErrorTypes.ooms = true
+        config.autoTrackSessions = true
+        Bugsnag.start(with: config)
+    }
+
+    override func run() {
+//        Bugsnag.startSession()
```
twometresteve, line 21: Presumably this should be removed?

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/OOMDefaultscenario.swift
```diff
@@ -0,0 +1,31 @@
...
+// Copyright (c) 2018 Bugsnag. All rights reserved.
+//
+
+import Foundation
+import Bugsnag
+
+/**
+ * Sends a handled Error to Bugsnag
+ */
+class OOMDefaultScenario: OOMBaseScenario {
```
twometresteve, line 12: Why can't I see where OOMBaseScenario is defined?  Should the other classes inherit from it?

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/OOMSessionScenario.swift
```diff
@@ -0,0 +1,23 @@
+//
+//  OOMLoadScenario.swift
```
twometresteve, line 2: Looks like a copy/paste error.

features/out_of_memory.feature
```diff
@@ -0,0 +1,65 @@
...
+    And the event "unhandled" is false
+    And the exception "message" equals "OOMLoadScenario"
+    And the event has a "manual" breadcrumb named "OOMLoadScenarioBreadcrumb"
+    And I discard the oldest request
+
+    When I relaunch the app
+    And I configure Bugsnag for "OOMLoadScenario"
+    And I wait to receive a request
+    Then the "Bugsnag-API-Key" header equals "0192837465afbecd0192837465afbecd"
+    And the error is an OOM event
```
twometresteve, line 19: At the moment I'm missing the bit of the jigsaw that makes this happen - please explain IRL :-)

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/OOMSessionScenario.swift
Cawllec: Ah, that bit is correct, the `autoTrackSessions` opt needed to be changed

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/OOMDefaultscenario.swift
Cawllec: Sorry, that file was removed from the project but not the repo

features/out_of_memory.feature
Cawllec: If you mean the step, it's present here: https://github.com/bugsnag/bugsnag-cocoa/pull/842/files#diff-3be9e7e3abbbdb0df337cf5d28bc9f8548167b7f6232f6b9c4b18da06e7c3012R242

twometresteve: Oh, sorry - ambiguous comment.  I meant how does the OOM actually occur?  There's a sentinel file that's written during the test, right?  I can't see the code that does that, but presumably that's because I can't find OOMBaseScenario.

features/fixtures/ios-swift-cocoapods/iOSTestApp/scenarios/OOMDefaultscenario.swift
twometresteve: Can you spell the path out to me, I've grepped my checkout and can't see the implementation.

twometresteve: Ah, I think I see what you mean now - `OOMDefaultscenario.swift`should have been removed and you have done so now.

features/out_of_memory.feature
Cawllec: That's been removed in favour of force restarting the app in the middle of a test run.  For example, [https://github.com/bugsnag/bugsnag-cocoa/pull/842/files#diff-776a4177f78bb5fb9417fbf06945d091b0956ac9a1e0cd6d94bc17601eb074aeR15](this step) will cause appium to kill and restart the app, testing the OOM triggering mechanism as well as the configured response.

While I feel this is a better solution in general as it tests the whole of the OOM process, it does mean these tests may be less compatible with non-browserstack devices, as the way appium kills the app and whether it triggers the OOM mechanism, may differ between platforms.

twometresteve: Thanks, that makes a lot more sense now.  In that case, can I suggest we write a comment something like the following at the top of the feature file, and annotate each `I relaunch the app` step with "See note above"?

> Due to the combination of BrowserStack's behaviour when resetting the app and the way that our OOM detection works, the `I relaunch the app` steps are currently sufficient to trigger the OOM mechanism.  However, these tests may not behave in the same way on local devices, device farms other that BrowserStack, or if we change that OOM detection works.

Cawllec: As a solution to this I'd suggest skipping the OOM test by default, but run them if a flag is present.  This will allow us to verify OOMs on CI without breaking our E2E tests running locally on device.

twometresteve: I'm not a fan of skipping tests by default, in case we forget to add the necessary flag when the tests should be run.  For now there are other issues running tests on local devices (I think any unhandled errors), so this will probably come out in the wash - but in principle yes we could certainly do something to opt-out of running them on local devices.

twometresteve (commented):
Looks good from a test perspective and my understanding of OOMs.  I'd be happy to see this merged if @nickdowell is happy.

nickdowell:
Looks good to me 👍 

Once our OOM detection is improved we'll have to create more realistic OOM scenarios.

nickdowell (approved):

PR #851, opened 2020-10-19
Use os_proc_available_memory if available

kstenerud:
## Goal

Apple provides a more accurate measure of usable memory in iOS 13 compared to what we use currently. We should use it when it's available.

## Design

Modified `bsg_ksmachusableMemory()` to call `os_proc_available_memory()` when it's available.

## Testing

Ran all unit tests on all platforms, and did manual tests to verify that the free memory values make sense.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMach.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMach.c
@@ -35,6 +35,9 @@
 #include <mach-o/arch.h>
 #include <mach/mach_time.h>
 #include <sys/sysctl.h>
+#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
+#import <os/proc.h>
+#endif
 
 // Avoiding static functions due to linker issues.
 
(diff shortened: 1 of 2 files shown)


Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMach.c
```diff
@@ -65,6 +68,15 @@ uint64_t bsg_ksmachfreeMemory(void) {
 }
 
 uint64_t bsg_ksmachusableMemory(void) {
+#if TARGET_OS_IOS
+    // Crashes on TVOS, not available on MacOS.
+    if (__builtin_available(iOS 13.0, *)) {
+        size_t mem = os_proc_available_memory();
```
nickdowell, line 74: I think this belongs in `bsg_ksmachfreeMemory`

As the memory footprint of the app increases, the value returned by `os_proc_available_memory` tends towards 0

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ Changelog
 
 ## Enhancements
 
+* Use new Apple API to get free memory, if available.
```
nickdowell, line 8: Can we just say "Now using os_proc_available_memory" instead of "new Apple API"?

nickdowell (changes requested):

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMach.c
kstenerud: `bsg_ksmachfreeMemory` is a relic from the old KSCrash days before I understood much about how Apple's memory systems work. It should be removed, and places that use it updated to use a different call.

nickdowell: The value returned from `bsg_ksmachfreeMemory` is used for `freeMemory` in the crash reports / dashboard, and the value from `bsg_ksmachusableMemory` is shown as `totalMemory`

kstenerud: Yes you're right. The existing calculation of free memory is so wrong it threw me off :P

usable memory is also incorrect, but that's for another day.

nickdowell (approved):

PR #853, opened 2020-10-21
[PLAT-5298] Remove breadcrumbs from client metadata

nickdowell:
## Goal

These changes reduce the CPU and memory impact of leaving breadcrumbs by stopping the breadcrumbs array from being serialised into an in-memory JSON buffer every time a new one is added.

## Changeset

`BSSerializeDataCrashHandler` which is invoked in the event of an unhandled error now reads in the breadcrumbs data from the file on disk. Previously, the breadcrumbs were included in `bsg_g_bugsnag_data.stateJSON`

This required a slight change to where the breadcrumbs reside within the KSCrashReport JSON structure, so the `BSGParseBreadcrumbs` function was updated to reflect this.

## Testing

A unit test was added to measure the performance of leaving breadcrumbs. This should report any regressions against the baseline as a test failure.

End-to-end tests check the breadcrumbs payload and these have passed.

Manually verified that breadcrumbs are still reported to the dashboard.


kstenerud (approved):

kattrali (approved):
looks good 🎉 

Even though this doesn't have a public interface change, I'd recommend adding a changelog entry for this as its intended to address reported performance concerns and can be referenced in the future as a part of the next release.

PR #856, opened 2020-10-22
Catch JSON exceptions rather than crashing

kstenerud:
## Goal

We've previously made changes to wrap `NSJSONSerialization` to catch unexpected exceptions while encoding/decoding, but there is still an alternative path that hasn't been handled.

## Design

Replicate the technique from `BSGJSONSerialization` in `BSG_KSJSONCodec` (wrap top-level APIs in try/catch blocks)

## Changeset

`BSG_KSJSONCodec` has been modified such that the top-level APIs are guarded with try/catch blocks that log any errors.

## Testing

Created a test app that exhausts memory in a region that uses the JSON serialization and observed the error messages prior to it crashing elsewhere from OOM.


nickdowell:
Is there a way to have automated test coverage of this?

kstenerud:
I'm not sure... So far the only exceptions I've managed are alloc exceptions during OOM situations. Other kinds of exceptions would require a bug in the codec.

nickdowell (approved):

PR #858, opened 2020-10-22
Implement granular breadcrumb storage

nickdowell:
## Goal

This is the first part of implementing granular breadcrumb storage - handling the writing of breadcrumbs and reading them from disk in the event of an OOM.

Inserting breadcrumbs into KSCrashReports for unhandled errors is to follow in a subsequent PR.

## Changeset

* `breadcrumbs` is now an immutable atomic `NSArray` - this eliminates `mutated while being enumerated` crashes. `getBreadcrumbs` has been removed now that `breadcrumbs` is safe to access.
* Breadcrumbs are now stored in individual files, using an always-incrementing naming scheme, and deleting old files once `maxBreadcrumbs` has been reached.
* Breadcrumbs will no longer be added to the in-memory array if they cannot be serialised.
* `cachedBreadcrumbs` has been updated to read from these individual files.
* `removeAllBreadcrumbs` is called by `BugsnagClient` during start-up to remove breadcrumb files from the previous session.

## Testing

The unit tests have been updated, as some were relying on `breadcrumbs` being a mutable array.

Manually verified that crash reports still reach the dashboard - albeit without any breadcrumbs for unhandled errors (expected at this stage.)

E2E test will fail as a result of missing breadcrumbs in unhandled errors.

### DIFF
--- a/Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
+++ b/Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
@@ -23,8 +23,15 @@
 + (instancetype _Nullable)breadcrumbFromDict:(NSDictionary *_Nonnull)dict;
 @end
 
+#pragma mark -
+
 @interface BugsnagBreadcrumbs ()
+
 @property BugsnagConfiguration *config;
+@property NSArray<BugsnagBreadcrumb *> *breadcrumbs;
+@property NSUInteger lastFileNumber;
+@property NSUInteger maxBreadcrumbs;
+
 @end
 
 #pragma mark -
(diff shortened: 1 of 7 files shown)


Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
```diff
@@ -23,28 +23,38 @@ + (instancetype _Nullable)breadcrumbWithBlock:
...
+        return nil;
     }
+    
+    _config = config;
+    _breadcrumbs = [NSArray array];
+    // Capture maxBreadcrumbs to protect against config being changed after initialization
+    _maxBreadcrumbs = config.maxBreadcrumbs;
+    
+    NSError *error = nil;
+    NSString *cachesDir = NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES).firstObject;
```
kstenerud, line 52: There's code for this in https://github.com/bugsnag/bugsnag-cocoa/blob/master/Bugsnag/Helpers/BSGCachesDirectory.m#L38

Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
```diff
@@ -55,35 +65,29 @@ - (void)addBreadcrumb:(NSString *)breadcrumbMessage {
...
+    NSUInteger fileNumber;
+    @synchronized (self) {
+        NSMutableArray<BugsnagBreadcrumb *> *breadcrumbs = [self.breadcrumbs mutableCopy];
+        if (breadcrumbs.count >= self.maxBreadcrumbs) {
+            [breadcrumbs removeObjectAtIndex:0];
+        }
+        [breadcrumbs addObject:crumb];
+        self.breadcrumbs = [NSArray arrayWithArray:breadcrumbs];
+        fileNumber = self.lastFileNumber + 1;
+        self.lastFileNumber = fileNumber;
```
kstenerud, line 88: We could remove the synchronized part here if self.breadcrumbs were atomic and self.lastFileNumber were an atomic int.

kstenerud (changes requested):

Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
nickdowell: I did consider that but opted for this approach instead because
* the `OSAtomic` routines are deprecated in favour of C++ `std::atomic` and wanted to avoid switching this to a `.mm`
* the `breadcrumbs` setter is now atomic but that doesn't prevent a race condition where a breadcrumb could be lost if two threads attempt to add one simultaneously - it's the "transaction" that need to be atomic, and therefore requires the `@synchronized` block

kstenerud (approved):

Bugsnag/Client/BugsnagClient.m
```diff
@@ -164,7 +164,9 @@ void BSSerializeDataCrashHandler(const BSG_KSCrashReportWriter *writer, int type
             writer->addJSONElement(writer, "state", bsg_g_bugsnag_data.stateJSON);
         }
         if (bsg_g_bugsnag_data.breadcrumbsPath) {
-            writer->addJSONFileElement(writer, "breadcrumbs", bsg_g_bugsnag_data.breadcrumbsPath);
+            // FIXME: This needs to be updated to cater for new breadcrumb storage scheme
+            // FIXME: addJSONFileElement generates broken JSON if the file cannot be opened
```
nickdowell, line 168: Correction: addJSONFileElement generates broken JSON if passed the path to a directory

kattrali (approved):

PR #859, opened 2020-10-22
Adding custom tabs to OOM reports.

tubatintin:
## Goal

Using the metadata client set for showing as new tabs in OOM reports.

## Testing

It was tested in a real project and client set metadata are visible as custom tabs in OOM reports like in other reports.

### DIFF
--- a/Bugsnag/Client/BugsnagClient.m
+++ b/Bugsnag/Client/BugsnagClient.m
@@ -1562,6 +1562,10 @@ NSString *const BSGBreadcrumbLoadedMessage = @"Bugsnag loaded";
           toSection:(NSString *_Nonnull)sectionName
 {
     [self.metadata addMetadata:metadata withKey:key toSection:sectionName];
+    [self.configuration addOnSendErrorBlock:^BOOL (BugsnagEvent *event) {
+        [event addMetadata:metadata withKey:key toSection:sectionName];
+        return YES;
+    }];
 }
 
 - (id _Nullable)getMetadataFromSection:(NSString *_Nonnull)sectionName


johnkiely1:
Hi @tubatintin,

Thanks for the PR, however I'm not sure it will work as you expect.

A little background on OOMs (apologies if this is something that you already know!): Out of Memory errors on iOS cannot be detected directly by Bugsnag, so we use a heuristic. We monitor for the app closing unexpectedly and infer that it’s an out of memory error on next relaunch by ruling out other known app termination causes. We then send the report to the dashboard. Any addOnSendErrorBlocks are run at delivery time so is no longer in the same session as when the OOM occurred. 

By using the `addOnSendErrorBlock` like this you would be effectively attaching metadata associated with the new session (from relaunch) with that of the report captured for the OOM report (and any other report) which would be about to be sent. While static information could be added it would not be specific to the original event session.

While there may be circumstances where this would be useful in general this would be confusing and unexpected behaviour if we were to adopt it. In this scenario we could not accept the PR, but thanks for the effort, we do appreciate it.

If this is the behaviour you require for your circumstances you can use a [global onSendError block](https://docs.bugsnag.com/platforms/ios/customizing-error-reports/#updating-events-using-callbacks) and apply the metadata there. If you only wanted this to apply only to OOMs you could [determine the errorclass](https://docs.bugsnag.com/platforms/ios/customizing-error-reports/#errors) contained within the event first. But again this is adding data after the fact and will not be from the OOM session. 

As an aside we are currently working on something to allow attaching metadata to OOMs.

Thanks.

PR #860, opened 2020-10-23
Insert breadcrumbs into KSCrashReport

nickdowell:
## Goal

The previous PR split the breadcrumbs into separate files, this PR implements the logic to include the breadcrumbs into KSCrashReports.

## Design

I found that the `BSG_KSCrashReportWriter` element APIs do work for array containers - they simply ignore the passed in `name`. Some unit tests have been added to verify this behaviour.

## Testing

Manually verified that crashes from the example app contain breadcrumbs.

Covered by E2E tests, which now pass.


Bugsnag/Client/BugsnagClient.m
```diff
@@ -163,10 +162,17 @@ void BSSerializeDataCrashHandler(const BSG_KSCrashReportWriter *writer, int type
...
-            // FIXME: This needs to be updated to cater for new breadcrumb storage scheme
-            // FIXME: addJSONFileElement generates broken JSON if the file cannot be opened
-            // writer->addJSONFileElement(writer, "breadcrumbs", bsg_g_bugsnag_data.breadcrumbsPath);
+        if (bsg_g_bugsnag_data.breadcrumbs) {
+            writer->beginArray(writer, "breadcrumbs");
+            BugsnagBreadcrumbsContext *breadcrumbs = bsg_g_bugsnag_data.breadcrumbs;
+            for (int i = breadcrumbs->firstFileNumber; i < breadcrumbs->nextFileNumber; i++) {
+                static char path[PATH_MAX];
+                if (snprintf(path, sizeof(path), "%s/%d.json", breadcrumbs->directoryPath, i) > sizeof(path)) {
+                    continue;
```
kstenerud, line 171: We should spit out an error if this happens

Bugsnag/Client/BugsnagClient.m
```diff
@@ -163,10 +162,17 @@ void BSSerializeDataCrashHandler(const BSG_KSCrashReportWriter *writer, int type
...
         }
-        if (bsg_g_bugsnag_data.breadcrumbsPath) {
-            // FIXME: This needs to be updated to cater for new breadcrumb storage scheme
-            // FIXME: addJSONFileElement generates broken JSON if the file cannot be opened
-            // writer->addJSONFileElement(writer, "breadcrumbs", bsg_g_bugsnag_data.breadcrumbsPath);
+        if (bsg_g_bugsnag_data.breadcrumbs) {
+            writer->beginArray(writer, "breadcrumbs");
+            BugsnagBreadcrumbsContext *breadcrumbs = bsg_g_bugsnag_data.breadcrumbs;
+            for (int i = breadcrumbs->firstFileNumber; i < breadcrumbs->nextFileNumber; i++) {
+                static char path[PATH_MAX];
```
kstenerud, line 169: `static` would make this not thread-safe

Bugsnag/Client/BugsnagClient.m
```diff
@@ -416,9 +422,7 @@ - (id)initWithConfiguration:(BugsnagConfiguration *)initConfiguration {
                                                          }];
 
         self.breadcrumbs = [[BugsnagBreadcrumbs alloc] initWithConfiguration:self.configuration];
-        if (self.breadcrumbs.cachePath != nil) {
-            bsg_g_bugsnag_data.breadcrumbsPath = strdup(self.breadcrumbs.cachePath.fileSystemRepresentation);
-        }
+        bsg_g_bugsnag_data.breadcrumbs = self.breadcrumbs.context;
```
kstenerud, line 425: This feels really weird, setting a global from an instance variable... I worry about future refactorings, when deep knowledge of how this works and what's connected to what dims. Is there another way to do this?

kstenerud (changes requested):

Bugsnag/Client/BugsnagClient.m
nickdowell: Agreed, this is not nice, there's a bit of a mismatch here because we have a global `bsg_g_bugsnag_data` but `BugsnagClient` is not strictly a singleton. Note that [a few lines above](https://github.com/bugsnag/bugsnag-cocoa/blob/a6596e573b5979b83c84cd4ab29a5bf5b2da9648/Bugsnag/Client/BugsnagClient.m#L400) we have a similar bit of code:

```objc
bsg_g_bugsnag_data.onCrash = (void (*)(const BSG_KSCrashReportWriter *))self.configuration.onCrashHandler;
```

An alternative approach that could be used is to call from `BSSerializeDataCrashHandler` into a C function implemented in `BugsnagBreadcrumbs.m` - that would at least encapsulate the strangeness.

Bugsnag/Client/BugsnagClient.m
nickdowell: Will allocate it on the stack 👍

kstenerud (approved):

Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
```diff
@@ -197,3 +217,19 @@ - (void)deleteBreadcrumbFiles {
 }
 
 @end
+
+#pragma mark -
+
+void BugsnagBreadcrumbsWriteCrashReport(BSG_KSCrashReportWriter *writer) {
+    char path[PATH_MAX];
+    writer->beginArray(writer, "breadcrumbs");
+    for (int i = g_context.firstFileNumber; i < g_context.nextFileNumber; i++) {
```
kattrali, line 226: `firstFileNumber` and `nextFileNumber` are unsigned ints while `i` is an int.

Bugsnag/Client/BugsnagClient.m
```diff
@@ -163,11 +161,7 @@ void BSSerializeDataCrashHandler(const BSG_KSCrashReportWriter *writer, int type
         if (bsg_g_bugsnag_data.stateJSON) {
             writer->addJSONElement(writer, "state", bsg_g_bugsnag_data.stateJSON);
         }
-        if (bsg_g_bugsnag_data.breadcrumbsPath) {
-            // FIXME: This needs to be updated to cater for new breadcrumb storage scheme
-            // FIXME: addJSONFileElement generates broken JSON if the file cannot be opened
-            // writer->addJSONFileElement(writer, "breadcrumbs", bsg_g_bugsnag_data.breadcrumbsPath);
-        }
+        BugsnagBreadcrumbsWriteCrashReport(writer);
```
kattrali, line 164: This is more of a nit since I don't see the case where it would happen, but `BugsnagBreadcrumbsWriteCrashReport` takes a nonnull writer parameter, while `BSSerializeDataCrashHandler`'s writer is not marked as nonnull. I'd only see this becoming an issue if/when analyzer results are added to the test builds because it will probably set it off.

Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
```diff
@@ -197,3 +217,19 @@ - (void)deleteBreadcrumbFiles {
...
 
 @end
+
+#pragma mark -
+
+void BugsnagBreadcrumbsWriteCrashReport(BSG_KSCrashReportWriter *writer) {
+    char path[PATH_MAX];
+    writer->beginArray(writer, "breadcrumbs");
+    for (unsigned int i = g_context.firstFileNumber; i < g_context.nextFileNumber; i++) {
+        int result = snprintf(path, sizeof(path), "%s/%d.json", g_context.directoryPath, i);
```
kattrali, line 227:
```suggestion
        int result = snprintf(path, sizeof(path), "%s/%lu.json", g_context.directoryPath, i);
```
?

kattrali (approved):

Bugsnag/Client/BugsnagClient.m
nickdowell: `BSSerializeDataCrashHandler` does not defined any nullability and therefore defaults to `_Null_unspecified`.

This means the static analyser doesn't perform the checks - I have verified that it doesn't report any issues here

Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.m
nickdowell: Fixed in b2a806d792234e64dd73dd8d1b364eb6bee10095

Bugsnag/Client/BugsnagClient.m
kattrali: Ah, good! I thought it might resolve to nullable 🙏

PR #861, opened 2020-10-26
Remove in-memory breadcrumbs

nickdowell:
## Goal

This reduces memory consumption by removing the in-memory copy of breadcrumbs.

## Design

Breadcrumbs are now loaded from disk whenever access is required.

As a side-effect this removes the possibility of crashes due to multithreaded access to breadcrumbs.

## Testing

This behaviour is covered by unit tests (which required some modifications) and E2E test.

Verified that crash reports in the dashboard still contain the expected breadcrumbs.

Verified that breadcrumb modifications in an `OnSendErrorBlock` still alter what is sent to the dashboard.

### DIFF
--- a/Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.h
+++ b/Bugsnag/Breadcrumbs/BugsnagBreadcrumbs.h
@@ -22,6 +22,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (instancetype)initWithConfiguration:(BugsnagConfiguration *)config;
 
+/**
+ * The current breadcrumbs, loaded from disk.
+ */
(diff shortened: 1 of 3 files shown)


kstenerud (approved):

kattrali (approved):

sethfri:
👋 Customer question if you don't mind: How does this impact performance for reading/writing breadcrumbs? Why was the decision made to move the in-memory version rather than synchronizing access to the resource?

nickdowell:
Hi @sethfri this change reduces the memory overhead of writing breadcrumbs, and other related changes have significantly reduced the CPU overhead of doing so. It will now be slower to read the breadcrumbs, but since writes are much more common than reads it makes sense to optimize it this way round.

PR #862, opened 2020-10-26
5263 persist default version info

kstenerud:
## Goal

When version data is not manually specified, it doesn't get stored to the config metadata JSON buffer, and thus doesn't get stored in any crash events. When loading the event, BugsnagApp tries to fetch the data from the event, then the config, then system data. Since the event data is not present, it ends up fetching current data instead of past data, and gives mismatched values for `app version` and `version` if they happen to change between the last crash and the next launch.

## Design

Force version and app version to be persisted to the config metadata so that it's always available from config no matter how it was initialised.

`ReleaseStage` and `AppType` have also been modified to store their defaults this way.

## Testing

Tested crashing the example app and changing version info before relaunch under the following conditions:

- Don't manually config. Change info.plist version info between launches (where the bug occurred).
- Manually config. Change manually set version info between launches.
- Manually config. Change info.plist between launches (should make no difference to reported version info).
- Manually config. Change info.plist and manually set version info between launches (should show manually set info).

### DIFF
--- a/Bugsnag/Configuration/BugsnagConfiguration.m
+++ b/Bugsnag/Configuration/BugsnagConfiguration.m
@@ -172,7 +172,6 @@ NSString * const kBugsnagUserUserId = @"BugsnagUserUserId";
 
     _metadata = [[BugsnagMetadata alloc] init];
     _config = [[BugsnagMetadata alloc] init];
-    _bundleVersion = NSBundle.mainBundle.infoDictionary[@"CFBundleVersion"];
     _endpoints = [BugsnagEndpointConfiguration new];
     _sessionURL = [NSURL URLWithString:@"https://sessions.bugsnag.com"];
     _autoDetectErrors = YES;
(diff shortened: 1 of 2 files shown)


Bugsnag/Configuration/BugsnagConfiguration.m
```diff
@@ -216,6 +217,11 @@ - (instancetype _Nonnull)initWithApiKey:(NSString *_Nonnull)apiKey
     #elif BSG_PLATFORM_OSX
         _appType = @"macOS";
     #endif
+
+    // Quick fix to get version info working, but the rest of the default values are still not persisted.
+    [self setAppVersion:_appVersion];
```
tomlongridge, line 222: The only other fields I can see in the same category are `appType` (which won't change between launches) and `releaseStage` (which is possibly but unlikely to change). The other fields all control functionality that we'd want to use the value from the current (not the persisted) config.

Suggest you remove the direct setting of the fields from lines 175/176 and simply call the setters for app version, bundle version, app type and release stage here.

I don't see why this is a quick fix - this seems legitimate.

Bugsnag/Configuration/BugsnagConfiguration.m
```diff
@@ -216,6 +215,10 @@ - (instancetype _Nonnull)initWithApiKey:(NSString *_Nonnull)apiKey
     #elif BSG_PLATFORM_OSX
         _appType = @"macOS";
     #endif
+
+    [self setAppVersion:NSBundle.mainBundle.infoDictionary[@"CFBundleShortVersionString"]];
```
nickdowell, line 219: Do we definitely need to do this?

IIRC the bug was that we were reporting the wrong `bundleVersion` - isn't the removal of `_bundleVersion = NSBundle.mainBundle.infoDictionary[@"CFBundleVersion"];` enough to fix that?

kstenerud: There were two problems that this change addresses:
- `CFBundleShortVersionString` wasn't being stored at all, which caused the issue where changing app version between crashing and reporting caused mismatched `CFBundleShortVersionString` and `CFBundleVersion`, reporting to a nonexistent version in the dashboard.
- Neither `CFBundleVersion` nor `CFBundleShortVersionString` were being put into `BugsnagConfiguration` internal cache because they were being stored to the member variable directly rather than using the setter. This caused them to be omitted from all events, which caused the event reader to fallback to current values for these, causing the old error to be reported to the new version if the app version changed between crashing and reporting.

nickdowell (approved):

Bugsnag/Configuration/BugsnagConfiguration.m
kattrali: Looks like Tom made an unresolved point here - the same behavior as was problematic for `appVersion` and `bundleVersion` is also being used for release stage, which can be considered versioning info. It looks like it also needs to be propagated into config.

kattrali (approved):

PR #879, opened 2020-11-10
[PLAT-5382] Prepare fixtures for running Mac E2E tests

nickdowell:
## Goal

There were some small differences between the test fixtures that were problematic for running the E2E tests on macOS.

## Changeset

* auto_detect_errors.feature contained a bug - it was entering an invalid scenario name resulting the fixture app crashing
* accessibility identifiers have been unified in the iOS and macOS fixtures (following the React Native fixture's naming convention)
* the Mac fixture's Info.plist now matches iOS as some of the test scenarios rely on specific configuration in there
* some Mac changes were required to ensure that the app crashes for uncaught exceptions (NSApplication and other mechanisms have a [tendency to swallow exceptions](https://www.chimehq.com/blog/sad-state-of-exceptions) thrown on the main thread)

## Testing

The iOS E2E tests should continue to run as before.

This branch does not contain all the changes required to run the E2E tests on a Mac - there will be some changes to maze runner and most of the .feature files also require some modifications to cater for multiple platforms.

### DIFF
--- a/features/fixtures/macos/macOSTestApp/main.m
+++ b/features/fixtures/macos/macOSTestApp/main.m
@@ -9,5 +9,15 @@
 #import <Cocoa/Cocoa.h>
 
 int main(int argc, const char * argv[]) {
+    [[NSUserDefaults standardUserDefaults] registerDefaults:@{
+        // Disable state restoration to prevent the following dialog being shown after crashes
+        // "The last time you opened macOSTestApp, it unexpectedly quit while reopening windows.
+        //  Do you want to try to reopen its windows again?"
+        // https://developer.apple.com/library/archive/releasenotes/AppKit/RN-AppKitOlderNotes/index.html#10_7StateRestoration
+        @"ApplePersistenceIgnoreState": @YES,
+        // Stop NSApplication swallowing NSExceptions thrown on the main thread.
+        @"NSApplicationCrashOnExceptions": @YES,
+    }];
+    
     return NSApplicationMain(argc, argv);
 }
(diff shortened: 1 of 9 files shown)


features/fixtures/macos/macOSTestApp/main.m
```diff
@@ -9,5 +9,15 @@
...
 
 int main(int argc, const char * argv[]) {
+    [[NSUserDefaults standardUserDefaults] registerDefaults:@{
+        // Disable state restoration to prevent the following dialog being shown after crashes
+        // "The last time you opened macOSTestApp, it unexpectedly quit while reopening windows.
+        //  Do you want to try to reopen its windows again?"
+        // https://developer.apple.com/library/archive/releasenotes/AppKit/RN-AppKitOlderNotes/index.html#10_7StateRestoration
+        @"ApplePersistenceIgnoreState": @YES,
+        // Stop NSApplication swallowing NSExceptions thrown on the main thread.
+        @"NSApplicationCrashOnExceptions": @YES,
```
kattrali, line 19: This should align with the [docs](https://docs.bugsnag.com/platforms/macos/#reporting-unhandled-exceptions), with either the tests and the docs advocating for this defaults change or by subclassing NSApplication

nickdowell: Subclassing NSApplication like the docs suggest will not result in the app crashing as expected by our test scenarios, and so the test will not pass.

kattrali: Right, no argument there. I'm saying that if that's what's expected in the tests, then the docs should match. Recommending that people turn it on when configuring the library (or more generally to avoid corrupting app state) also seems like a reasonable position.

twometresteve (approved):
LGTM, pending CI.

kattrali (approved):

kstenerud (approved):

features/fixtures/macos/macOSTestApp/main.m
nickdowell: https://github.com/bugsnag/docs.bugsnag.com/pull/1256

PR #902, opened 2020-11-23
[PLAT-5412] Add BugsnagStackframe.type

nickdowell:
## Goal

Adding `type` to `BugsnagStackframe` is a prerequisite for supporting mixed stack traces containing multiple languages.

## Design

The `type` parameter is intended to be `nil` in most cases, and only set if the stack trace contains more than one language - e.g. a `reactnativejs` exception that contains some `cocoa` frames.

Because the existing `BSGErrorType` is an int-based enum and cannot naturally represent a nil value, a new Objective-C style string enum `BugsnagStackframeType` has been created specifically for this property.

In a future notifier update when we are making breaking API changes, we should consider switching all our enums that are serialized as strings to NS_TYPED_ENUMs. 

## Changeset

* Added `type` parameter
* Added `BugsnagStackframe+Private.h` to remove duplicate declarations
* Adopted `NS_ASSUME_NONNULL` for stack frame header

## Testing

* Implemented unit tests to validate expected behaviour of `type` parameter
* Manually verified using a customised version of BugsnagReactNative

### DIFF
--- a/Bugsnag/include/Bugsnag/BugsnagStackframe.h
+++ b/Bugsnag/include/Bugsnag/BugsnagStackframe.h
@@ -8,6 +8,12 @@
 
 #import <Foundation/Foundation.h>
 
+NS_ASSUME_NONNULL_BEGIN
+
+typedef NSString * BugsnagStackframeType NS_TYPED_ENUM;
+
+FOUNDATION_EXPORT BugsnagStackframeType const BugsnagStackframeTypeCocoa;
+
 /**
  * Represents a single stackframe from a stacktrace.
  */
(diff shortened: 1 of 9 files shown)


Bugsnag/include/Bugsnag/BugsnagStackframe.h
```diff
@@ -58,11 +64,18 @@
  */
 @property BOOL isLr;
 
+/**
+ * The type of the stack frame, if it differs from that of the containing error or event.
+ */
+@property(nullable) BugsnagStackframeType type;
```
tomlongridge, line 70: I think we'd want to use `BSGErrorType` rather than a new type as the possible enum values will always be the same as the error type.

nickdowell: Unfortunately `BSGErrorType` was implemented as an integer-based enum and so cannot represent a "none" value which is a requirement for this new functionality.

I could change `BSGErrorType` to be a string-based enum but that seemed more disruptive.

Bugsnag/include/Bugsnag/BugsnagStackframe.h
```diff
@@ -8,6 +8,12 @@
 
 #import <Foundation/Foundation.h>
 
+NS_ASSUME_NONNULL_BEGIN
+
+typedef NSString * BugsnagStackframeType NS_TYPED_ENUM;
+
+FOUNDATION_EXPORT BugsnagStackframeType const BugsnagStackframeTypeCocoa;
```
kstenerud, line 15: Would we also need enum values for react-native and other frameworks on top of ios?

nickdowell: No - this exists only so that BugsnagReactNative can set the type of native stack frames. The reactnativejs type is applied to the containing error, which specifies the type for all frames that don't explicitly state a type.

kstenerud (approved):

kattrali (approved):

PR #905, opened 2020-11-24
Add private headers for error, event & thread

nickdowell:
## Goal

The private interfaces for `BugsnagError` `BugsnagEvent` and `BugsnagThread` need to be amended in order to implement the new `Bugsnag-Stacktrace-Types` HTTP header.

The private interfaces were redeclared throughout the codebase by the classes that needed to access them, leaving a maintenance problem. The goal here is to provide a single definition of the private interface in dedicated header files.

## Changeset

Removed all the private `@interface` declarations from .m files and copied them to individual header files that are not accessible publicly since they are not in the "include" folder.

In some files I sorted the list of imports alphabetically, which makes it easer to scan and know where to put new statements.

## Testing

Ran the unit tests and static analyzer locally.


kstenerud (approved):
Very nice! The scattered private declarations were a bit of a nightmare ;-)

kattrali (approved):

PR #906, opened 2020-11-24
[PLAT-5421] Implement Bugsnag-Stacktrace-Types HTTP header

nickdowell:
## Goal

The new `Bugsnag-Stacktrace-Types` HTTP header informs the back-end about the contents of the payload without having to parse it.

## Changeset

`BugnagErrorReportSink` sets the new HTTP header on outgoing requests.

Added a `stacktraceTypes` computed property on BugsnagEvent to make the logic testable.

Note that since (cocoa) threads are captured by default, all React Native errors will contain `reactnativejs` and `cocoa` types.

## Testing

Added unit tests to verify the expected types are returned for different kinds of events.

Manually verified the changes when used with BugsnagReactNative.

### DIFF
--- a/Bugsnag/Payload/BugsnagEvent.m
+++ b/Bugsnag/Payload/BugsnagEvent.m
@@ -807,4 +807,37 @@ NSDictionary *BSGParseCustomException(NSDictionary *report,
     self.handledState.unhandled = val;
 }
 
+#pragma mark -
+
+- (NSArray<NSString *> *)stacktraceTypes {
+    NSMutableSet *stacktraceTypes = [NSMutableSet set];
+    
+    // The error in self.errors is not always the error that will be sent; this is the case when used in React Native.
+    // Using [self toJson] to ensure this uses the same logic of reading from self.customException instead.
+    NSDictionary *json = [self toJson];
+    NSArray *exceptions = json[BSGKeyExceptions];
+    for (NSDictionary *exception in exceptions) {
+        BugsnagError *error = [BugsnagError errorFromJson:exception];
+        
+        [stacktraceTypes addObject:BSGSerializeErrorType(error.type)];
+        
+        for (BugsnagStackframe *stackframe in error.stacktrace) {
+            if (stackframe.type) {
+                [stacktraceTypes addObject:stackframe.type];
+            }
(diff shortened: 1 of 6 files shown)


Bugsnag/Payload/BugsnagEvent.m
```diff
@@ -807,4 +807,43 @@ - (void)updateUnhandled:(BOOL)val {
...
+            if (stackframe.type) {
+                [stacktraceTypes addObject:stackframe.type];
+            }
+        }
+        
+        for (BugsnagStackframe *stackframe in error.stacktrace) {
+            if (stackframe.type) {
+                [stacktraceTypes addObject:stackframe.type];
+            }
+        }
```
kattrali, line 834: Accidental duplication?

nickdowell: Removed the dupe

Tests/BugsnagEventTests.m
```diff
@@ -539,6 +540,47 @@ - (void)testApiKey {
...
+    event.threads = @[thread1];
+    XCTAssertEqualObjects(sorted(event.stacktraceTypes), (@[@"c", @"cocoa", @"java"]));
+
+    BugsnagThread *thread2 = [[BugsnagThread alloc] init];
+    thread2.stacktrace = @[
+        [BugsnagStackframe frameFromJson:@{@"type": @"csharp"}],
+        [BugsnagStackframe frameFromJson:@{@"type": @"android"}],
+    ];
+    event.threads = @[thread1, thread2];
+    XCTAssertEqualObjects(sorted(event.stacktraceTypes), (@[@"android", @"c", @"cocoa", @"csharp", @"java"]));
```
kstenerud, line 581: Why is it expecting `android`?

nickdowell: Having already used `cocoa` and `reactnativejs` in error.stacktrace, and `c` and `java` in thread 1, I needed some additional values to use in thread 2 to ensure they all get picked up correctly.

Any string could have been used in this test but I thought sticking to the prescribed values would be safest.

kattrali (approved):

PR #912, opened 2020-11-26
Allow overriding the unhandled flag in non-fatal errors.

kstenerud:
## Goal

Make the `unhandled` property of `BugsnagEvent` writable, and trace if it has been changed.

## Design

See ROAD-732

## Changeset

The existing `unhandled` property of `BugsnagEvent` will be made readwrite and `-[BugsnagEvent setUnhandled:]` will be implemented to propagate the value to `BugsnagHandledState.unhandled`

* A new `unhandledOverridden` property will be added to `BugsnagHandledState`
* `-[BugsnagEvent toJson]` will be updated to include `unhandledOverridden`
* `+[BugsnagHandledState handledStateFromJson:]` will be updated to restore the saved value of `unhandledOverridden`
* `-[BugsnagClient notifyInternal:block:]` will be updated to track whether the `unhandled` value has changed by the `OnErrorBlock` and set `unhandledOverridden` if so.

## Testing

Added unit tests:

* Run code that triggers `unhandled` setters.
* Verify that `[BugsnagClient notifyInternal:block:]` properly updates `unhandledOverridden` when `unhandled` changes
* Verify that serde functions properly deal with `unhandledOverridden`

### DIFF
--- a/Bugsnag/Payload/BugsnagHandledState.m
+++ b/Bugsnag/Payload/BugsnagHandledState.m
@@ -29,6 +29,7 @@ NSString *BSGFormatSeverity(BSGSeverity severity) {
 }
 
 static NSString *const kUnhandled = @"unhandled";
+static NSString *const kUnhandledOverridden = @"unhandledOverridden";
 static NSString *const kSeverityReasonType = @"severityReasonType";
 static NSString *const kOriginalSeverity = @"originalSeverity";
 static NSString *const kCurrentSeverity = @"currentSeverity";
@@ -49,6 +50,7 @@ static NSString *const kUserCallbackSetSeverity = @"userCallbackSetSeverity";
 
 + (instancetype)handledStateFromJson:(NSDictionary *)json {
     BOOL unhandled = [json[@"unhandled"] boolValue];
+    BOOL unhandledOverridden = [json[@"unhandledOverridden"] boolValue];
     NSDictionary *data = json[@"severityReason"];
     BSGSeverity severity = BSGParseSeverity(json[@"severity"]);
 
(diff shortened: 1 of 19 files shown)


Tests/BugsnagErrorReportSinkTests.m
```diff
@@ -356,6 +357,42 @@ - (void)testHandledSerialization {
...
 }
 
+- (void)testHandledOverriddenSerialization {
+    BugsnagHandledState *state =
+    [BugsnagHandledState handledStateWithSeverityReason:HandledException];
+    state.unhandled = YES;
+    state.unhandledOverridden = YES;
+    NSDictionary *payload = [self reportFromHandledState:state];
+    
+    XCTAssertEqualObjects(@"warning", payload[@"severity"]);
```
nickdowell, line 367: The arguments are the wrong way round; the expected value should be the second argument to `XCTAssertEqualObjects`

The same comment applies to the other uses of `XCTAssertEqualObjects` in this test method.

nickdowell (changes requested):
The changes look good 👍 

I think there needs to be some coverage added for this in the E2E tests, though

nickdowell (approved):

Bugsnag/Payload/BugsnagHandledState.m
```diff
@@ -199,6 +206,7 @@ + (SeverityReasonType)severityReasonFromString:(NSString *)string {
 - (NSDictionary *)toJson {
     NSMutableDictionary *dict = [NSMutableDictionary new];
     dict[kUnhandled] = @(self.unhandled);
+    dict[kUnhandledOverridden] = @(self.unhandledOverridden);
```
kattrali, line 209: This is specified in the product design to default to `false` if omitted, so a check here to avoid sending a few extra bytes seems reasonable.

PR #916, opened 2020-11-27
Update travis to include latest Xcode, and cull unsupported ios,tvos versions

kstenerud:
## Goal

Travis is currently building test environments using xcode versions that can no longer submit apps to the app store. Since effectively nobody can use these xcode versions anymore, we shouldn't be wasting resources running tests with them.

## Changeset

Modified the test runs to:

| OS | Xcode |
| -- | -- |
| macos 10.15.7 | 12.2 |
| macos 10.15.5 | 11.6 |
| macos 10.14 | 10.2 |
| macos 10.13 | 9.4 |
| ios 11-14 | 12.2 |
| ios 10 | 11 |
| ios 9 | 10.2 |
| tvos 11-14 | 12.2 |
| tvos 10 | 11 |
| tvos 9 | 10.2 |

## Testing

Ran all of the tests on Travis (see testing report below)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,16 @@ jobs:
   # Unit Tests - Mac
   # ----------------------------------------------------------------------------
 
+  - osx_image: xcode12.2 # macos 10.15.7
+    name: macOS 10.15.7 unit tests
+    stage: macOS unit tests
+    env: PLATFORM=macOS
+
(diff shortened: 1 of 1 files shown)


nickdowell:
I think we need to continue running our tests on iOS / tvOS 9.x as we still officially support those...

nickdowell (approved):

PR #918, opened 2020-12-01
Update NSException-disabled e2e test to account for SIGABRT

kstenerud:
## Goal

When NSException handling is disabled, the default handler calls `abort()`, which generates a `SIGABRT` signal. The current test doesn't account for this.

## Design

Update the DisableNSExceptionScenario test to look for a `SIGABRT` signal. I've also removed the `dispatch_after` call because it's unreliable as to whether the test rig will capture it properly or not (and also it's not needed in this case).

## Testing

Re-ran e2e test multiple times to make sure the test no longer flakes.


nickdowell (approved):

kattrali (approved):

twometresteve (commented):
One comment, but generally happy if others are.

PR #919, opened 2020-12-01
Provide the ability for use without UIKit

nickdowell:
## Goal

Some types of app extensions have such severe memory limits (e.g. 30MB for a file provider extension) that linking UIKit is undesirable.

These changes make it possible to use Bugsnag without linking UIKit.

## Design

The approach used here to avoid the use of any UIKit symbols - i.e. Objective-C classes, function names or symbols like NSNotification names.

We can still `#import <UIKit/UIKit.h>` because Clang's autolinking works on the basis of the symbols referenced by code.

_Note: this is contrary to the Quick Help Summary show in Xcode:_

> Automatically link SDK frameworks that are referenced using `#import` or `#include`. This feature requires also enabling support for modules. This build setting only applies to C-family languages.

_Even with `CLANG_MODULES_AUTOLINK = YES` (the default) `otool -L` shows that removing the use of all UIKit symbols from our code also removes the linker dependency._

We can also still send messages to UIKit objects in the normal way as this does not introduce any link-time dependencies.

## Changeset

* `CLANG_MODULES_AUTOLINK` has been set to `NO` in the top-level Bugsnag project - this will cause a build failure if we reintroduce a link-time dependency on UIKit e.g. `Undefined symbol: _OBJC_CLASS_$_UIDevice`
* Messages to UIKit classes now use `NSClassFromString()` to avoid linking against the class symbol
* Notification names are now hardcoded

## Testing

* Added a unit test case to verify that the hard-coded notification names match those used by the OS.
* Manually verified that Bugsnag works when used in a file provider extension without UIKit.

### DIFF
--- a/Bugsnag/BugsnagSystemState.m
+++ b/Bugsnag/BugsnagSystemState.m
@@ -10,7 +10,7 @@
 #if TARGET_OS_OSX
 #import <AppKit/AppKit.h>
 #else
-#import <UIKit/UIKit.h>
+#import "BSGUIKit.h"
 #endif
 
 #import "BugsnagSystemState.h"
--- a/Bugsnag/Client/BugsnagClient.m
+++ b/Bugsnag/Client/BugsnagClient.m
@@ -68,7 +68,7 @@
 #endif
 
 #if BSG_PLATFORM_IOS
-#import <UIKit/UIKit.h>
+#import "BSGUIKit.h"
 #elif BSG_PLATFORM_OSX
 #import <AppKit/AppKit.h>
 #endif
@@ -359,7 +359,7 @@ NSString *_lastOrientation = nil;
                                                                   client:self];
 
 #if BSG_PLATFORM_IOS
-        _lastOrientation = BSGOrientationNameFromEnum([UIDevice currentDevice].orientation);
+        _lastOrientation = BSGOrientationNameFromEnum([UIDEVICE currentDevice].orientation);
 #endif
         _user = self.configuration.user;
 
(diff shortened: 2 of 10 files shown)


Bugsnag/BugsnagSystemState.m
```diff
@@ -189,27 +190,27 @@ - (instancetype)initWithConfiguration:(BugsnagConfiguration *)config {
             [weakSelf setValue:@NO forAppKey:SYSTEMSTATE_APP_IS_IN_FOREGROUND];
         }];
 #else
-        [center addObserverForName:UIApplicationWillTerminateNotification object:nil queue:nil usingBlock:^(NSNotification * _Nonnull note) {
+        [center addObserverForName:@"UIApplicationWillTerminateNotification" object:nil queue:nil usingBlock:^(NSNotification * _Nonnull note) {
```
kstenerud, line 193: This feels dangerous, because any typo won't be detected. Could we instead define our own constants in one place, and then in the tests use a macro to check against the UIKit ones? e.g.

```c
#define ASSERT_NOTIFICATION_NAME(name) XCTAssertEqualObjects(BSGUI ## name, UI ## name)
ASSERT_NOTIFICATION_NAME(ApplicationDidBecomeActiveNotification);
ASSERT_NOTIFICATION_NAME(ApplicationDidEnterBackgroundNotification);
```

Bugsnag/Client/BugsnagClient.m
```diff
@@ -709,8 +709,8 @@ - (void)unsubscribeFromNotifications:(id)sender {
     [BSGConnectivity stopMonitoring];
 
 #if BSG_PLATFORM_IOS
-    [UIDevice currentDevice].batteryMonitoringEnabled = NO;
-    [[UIDevice currentDevice] endGeneratingDeviceOrientationNotifications];
+    [NSClassFromString(@"UIDevice") currentDevice].batteryMonitoringEnabled = NO;
+    [[NSClassFromString(@"UIDevice") currentDevice] endGeneratingDeviceOrientationNotifications];
```
kstenerud, line 713: Since this is used in a bunch of places, can we have it as a utility function?

kstenerud (changes requested):

kstenerud (approved):

kattrali (approved):

PR #921, opened 2020-12-02
Add unhandled override support for unhandled errors

kstenerud:
## Goal

With the last override PR, we now support overriding the `unhandled` status of handled exceptions (allowing the user to mark handled errors as "unhanded". This PR adds functionality to mark "unhandled" errors as "handled", setting the same "unhandledOverridden" field whenever this happens.

## Design

See ROAD-732

## Changeset

BugsnagEvent now detects whether an error has changed from unhandled to handled or vice versa, and sets `unhandledOverridden` to mark when it happens, adjusting `handledCount` and `unhandledCount` to compensate.

`unhandledOverridden` only gets encoded to JSON when it's true.

## Testing

Added "override" versions of the unhandled exception tests which change `unhandled` to false in a crash callback function.

### DIFF
--- a/Bugsnag/Payload/BugsnagEvent.m
+++ b/Bugsnag/Payload/BugsnagEvent.m
@@ -339,6 +339,25 @@ NSDictionary *BSGParseCustomException(NSDictionary *report,
     NSDictionary *error = [event valueForKeyPath:@"crash.error"];
     NSString *errorType = error[BSGKeyType];
 
+    // Always assume that a report coming from KSCrash is by default an unhandled error.
+    BOOL isUnhandled = YES;
+    BOOL isUnhandledOverridden = NO;
+    BOOL hasBecomeHandled = [event valueForKeyPath:@"user.unhandled"] != nil &&
+            [[event valueForKeyPath:@"user.unhandled"] boolValue] == false;
+    if (hasBecomeHandled) {
+        const int handledCountAdjust = 1;
+        isUnhandled = NO;
+        isUnhandledOverridden = YES;
+        NSMutableDictionary *user = [event[BSGKeyUser] mutableCopy];
+        user[@"unhandled"] = @(isUnhandled);
+        user[@"unhandledOverridden"] = @(isUnhandledOverridden);
+        user[@"unhandledCount"] = @([user[@"unhandledCount"] intValue] - handledCountAdjust);
+        user[@"handledCount"] = @([user[@"handledCount"] intValue] + handledCountAdjust);
+        NSMutableDictionary *eventCopy = [event mutableCopy];
+        eventCopy[BSGKeyUser] = user;
+        event = eventCopy;
+    }
+
     id userMetadata = [event valueForKeyPath:@"user.metaData"];
     BugsnagMetadata *metadata;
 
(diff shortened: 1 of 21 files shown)


nickdowell:
I'm a little concerned about the effect that adding so many new E2E scenarios will have on the run time. I think there's a 1 hour time limit which exceeding will result in a failure.

Is it important to check the unhandled -> handled override for so many different types of crashes?

kstenerud:
Hmm I suppose since the same path is taken by all unhandled exceptions, we'd only really need it in one place.

Bugsnag/Payload/BugsnagEvent.m
```diff
@@ -339,6 +339,30 @@ - (instancetype)initWithKSCrashData:(NSDictionary *)event {
     NSDictionary *error = [event valueForKeyPath:@"crash.error"];
     NSString *errorType = error[BSGKeyType];
 
+    // Data coming from KSCrash will have either user.handledCount = 0 and user.unhandledCount > 0, or the opposite
+    BOOL wasUnhandled = [event valueForKeyPath:@"user.unhandledCount"] == nil ||
```
nickdowell, line 343: I don't understand this logic... aren't all KSCrashes unhandled unless the developer has specifically called the following in their `onCrash` handler?

```
writer->addBooleanElement(writer, "unhandled", false);
```

nickdowell: Note: handled errors (i.e. `Bugsnag.notify()`) go through `-initWithUserData:`

kstenerud: I wasn't really sure if that's always the case. It felt safer to check that it was indeed originally marked unhandled.

nickdowell: This initializer is only used for unhandled errors;

```
/**
 * Creates a BugsnagEvent from unhandled error JSON. Unhandled errors use
 * the JSON schema supplied by the KSCrash report rather than the Bugsnag
 * Error API schema, which is more complex to parse.
 *
 * @param event a KSCrash report
 *
 * @return a BugsnagEvent containing the parsed information
 */
- (instancetype)initWithKSCrashData:(NSDictionary *)event {
```

Bugsnag/Payload/BugsnagEvent.m
```diff
@@ -339,6 +339,30 @@ - (instancetype)initWithKSCrashData:(NSDictionary *)event {
     NSDictionary *error = [event valueForKeyPath:@"crash.error"];
     NSString *errorType = error[BSGKeyType];
 
+    // Data coming from KSCrash will have either user.handledCount = 0 and user.unhandledCount > 0, or the opposite
```
nickdowell, line 342: This comment is misleading - the values of `handledCount` and `handledCount` come from the session tracker and we cannot assume that one of them will be 0.

For example, prior to the crash there may have been multiple non-fatal errors reported, and they could have been counted as handled or unhandled depending on whether the developer overrode the unhandled parameter.

nickdowell (approved):

Bugsnag/Payload/BugsnagEvent.m
```diff
@@ -339,6 +339,25 @@ - (instancetype)initWithKSCrashData:(NSDictionary *)event {
     NSDictionary *error = [event valueForKeyPath:@"crash.error"];
     NSString *errorType = error[BSGKeyType];
 
+    // Always assume that a report coming from KSCrash is by default an unhandled error.
+    BOOL isUnhandled = true;
+    BOOL isUnhandledOverridden = false;
```
kattrali, line 344:
```suggestion
    BOOL isUnhandled = YES;
    BOOL isUnhandledOverridden = NO;
```

Bugsnag/Payload/BugsnagEvent.m
```diff
@@ -339,6 +339,25 @@ - (instancetype)initWithKSCrashData:(NSDictionary *)event {
...
+    BOOL hasBecomeHandled = [event valueForKeyPath:@"user.unhandled"] != nil &&
+            [[event valueForKeyPath:@"user.unhandled"] boolValue] == false;
+    if (hasBecomeHandled) {
+        const int handledCountAdjust = 1;
+        isUnhandled = false;
+        isUnhandledOverridden = true;
```
kattrali, line 350:
```suggestion
        isUnhandled = NO;
        isUnhandledOverridden = YES;
```

kattrali (changes requested):

kattrali (approved):

PR #923, opened 2020-12-03
[PLAT-5477] Update appium version for iOS 12+ end to end tests

Cawllec:
## Goal
This should clear up a persistent error present in the appium logs.

Unfortunately iOS 10 only supports up to Appium v1.15, and through looking at the test runs iOS 11 seems to struggle beyond iOS 1.16.


twometresteve (approved):
Excellent - pleasantly surprised the tests just worked after making this change.  Not surprised to hear that about iOS 10/11 either.

nickdowell (approved):

PR #928, opened 2020-12-07
Remove BSGDictSetSafeObject and BSGArrayAddSafeObject

nickdowell:
## Goal

`BSGDictSetSafeObject` does not make our code safer - in fact it makes it more error prone because it means we can end up with `[NSNull null]` values in dictionaries, which then need to be guarded against when reading back from the dictionary (this is something our code does not currently do.)

## Changeset

`BSGDictSetSafeObject` and `BSGArrayAddSafeObject` have been removed.

## Testing

Tested by running unit tests.


Bugsnag/Payload/BugsnagApp.m
```diff
@@ -32,7 +32,7 @@ + (BugsnagApp *)deserializeFromJson:(NSDictionary *)json {
         app.id = json[@"id"];
         app.releaseStage = json[@"releaseStage"];
         app.type = json[@"type"];
-        app.version = json[@"version"];
+        app.version = [json[@"version"] isKindOfClass:[NSString class]] ? json[@"version"] : nil;
```
kstenerud, line 35: Why is this needed?

nickdowell: It isn't - was left over from an earlier revision... corrected now 👍

kstenerud (approved):

PR #935, opened 2020-12-11
Fix some flaky tests

kstenerud:
## Goal

Tests that relied upon session tracking to detect the non-sending of events can sometimes flake.

## Design

If we manually send a handled error, we can just check for only that error being received. This is already used elsewhere in the tests.

## Changeset

Modified affected tests to disable session tracking, and only expect one request.

## Testing

Re-ran e2e tests.

### DIFF
--- a/features/enabled_error_types.feature
+++ b/features/enabled_error_types.feature
@@ -4,28 +4,19 @@ Feature: Enabled error types
     Given I clear all persistent data
 
   Scenario: All Crash reporting is disabled
-    # Sessions: on, unhandled crashes: off
+    # enabledErrorTypes = None, Generate a manual notification, crash
     When I run "DisableAllExceptManualExceptionsAndCrashScenario" and relaunch the app
     And I configure Bugsnag for "DisableAllExceptManualExceptionsAndCrashScenario"
-    # Give ignored report time to be processed
-    And I wait for 2 seconds
-    And I wait to receive 2 requests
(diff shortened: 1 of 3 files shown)


features/enabled_error_types.feature
```diff
@@ -4,28 +4,19 @@ Feature: Enabled error types
...
     When I run "DisableAllExceptManualExceptionsAndCrashScenario" and relaunch the app
     And I configure Bugsnag for "DisableAllExceptManualExceptionsAndCrashScenario"
-    # Give ignored report time to be processed
-    And I wait for 2 seconds
-    And I wait to receive 2 requests
-    Then the request is valid for the session reporting API version "1.0" for the "iOS Bugsnag Notifier" notifier
-    And I discard the oldest request
-    Then the request is valid for the session reporting API version "1.0" for the "iOS Bugsnag Notifier" notifier
-
-  Scenario: All Crash reporting is disabled but manual notification works
```
nickdowell, line 17: Why has this scenario been removed?

kstenerud: It was redundant. When all types are disabled, manual reporting still works, so having two tests about this isn't necessary.

nickdowell: Ok I see now - one scenario was calling `notify` while the other was relying on the session payload, so the adjusted scenarios would be identical 👍

nickdowell (approved):

PR #936, opened 2020-12-11
Add new config option: maxPersistedEvents

kstenerud:
## Goal

To harmonize configuration options, we need a maxPersistedEvents config option

## Design

Add maxPersistedEvents to the configuration object

## Testing

Updated unit tests to also verify maxPersistedEvents.

We can't test interrupted network connectivity, so I did some manual testing where I set the max to 2, generated 5 errors with airplane mode on, and then verified that only the last 2 events were kept and sent.


Bugsnag/Configuration/BugsnagConfiguration.m
```diff
@@ -415,6 +417,26 @@ -(void)deletePersistedUserData {
...
+
+- (void)setMaxPersistedEvents:(NSUInteger)maxPersistedEvents {
+    @synchronized (self) {
+        if (maxPersistedEvents <= 100) {
+            _maxPersistedEvents = maxPersistedEvents;
+        } else {
+            bsg_log_err(@"Invalid configuration value detected. Option maxPersistedEvents "
+                        "should be an integer between 0-100. Supplied value is %lu",
+                        (unsigned long) maxPersistedEvents);
+        }
```
kattrali, line 436: 1. Can `notify` still function (e.g. actually deliver a report) if `maxPersistentEvents` is set to 0? if not, I'd recommend that for cocoa this value should be between 1-100 rather than 0-100.

2. Perhaps attempting to set an invalid value should follow the intent instead of being ignored? So if I set the value to 500, there will be a warning and it will be set to 100.

Bugsnag/Configuration/BSGConfigurationBuilder.m
```diff
@@ -29,6 +29,7 @@ + (BugsnagConfiguration *)configurationFromOptions:(NSDictionary *)options {
         BSGKeyBundleVersion,
         BSGKeyEnabledReleaseStages,
         BSGKeyEndpoints,
+        BSGKeyMaxPersistedEvents,
```
nickdowell, line 32: Pedantic: can we keep this list sorted alphabetically? 🙃

Bugsnag/Helpers/BugsnagKeys.h
```diff
@@ -62,6 +62,7 @@ extern NSString *const BSGKeyMachoFile;
 extern NSString *const BSGKeyMachoLoadAddr;
 extern NSString *const BSGKeyMachoUUID;
 extern NSString *const BSGKeyMachoVMAddress;
+extern NSString *const BSGKeyMaxPersistedEvents;
```
nickdowell, line 65: Pedantic: can we keep this list sorted alphabetically? 🙃

nickdowell (changes requested):

nickdowell (approved):

Bugsnag/Configuration/BugsnagConfiguration.m
kattrali: Unresolving for the latter half:

Perhaps attempting to set an invalid value should follow the intent instead of being ignored? So if I set the value to 500, there will be a warning and it will be set to 100.

kstenerud: We'd have to change the rest of the checks to match this pattern then. This code is basically copied from another check in BugsnagConfiguration, which refuses to change the value if it's out of range.

kattrali: Making the change elsewhere may not be a bad thing, and can be done incrementally. For now I'm asking 1) whether it was considered and 2) if it would be the ideal behavior, since this is new functionality and we have the opportunity to get it right without consequence.

kstenerud: I'm not sure. We should probably be harmonizing the behaviour across systems. This code is copied from the breadcrumbs system. Specifically:

- Setting to an out-of-range value prints an error and refuses to set the value.
- Setting to 0 effectively disables breadcrumbs.

Is this the behaviour we want?

PR #937, opened 2020-12-11
Fix unhandledOverride

kstenerud:
## Goal

The last unhandled override PR incorrectly placed `unhandledOverride` at the top level, when it should have been inside of `severityReason`.

## Changeset

Update JSON serializer and deserializer code to place it at the correct level, and update tests to match.

## Testing

Re-ran unit tests and e2e tests.


nickdowell (approved):
LGTM if the E2E tests pass

PR #938, opened 2020-12-14
[PLAT-4646] Implement discardClasses configuration option

nickdowell:
## Goal

Implements the `discardClasses` configuration option which contains strings or regular expressions that, when the `BugsnagError.errorClass` matches, will cause the error to be discarded.

## Design

For unhandled errors the evaluation is performed at next launch because it's not possible to evaluate regular expressions in an async-safe manner at crash time.

## Changeset

* Added `discardClasses` property to `BugsnagConfiguration`
* Added `-shouldDiscardErrorClass:` method to `BugsnagConfiguration`
* For handled errors: modified `BugsnagClient` to discard if `-shouldDiscardErrorClass:` returns `YES`
* For unhandled errors: modified `BugsnagErrorReportSink` to discard if `-shouldDiscardErrorClass:` returns `YES`

## Testing

* Added unit tests to test `-shouldDiscardErrorClass:` with a variety of `discardClasses` values against a set of real-world error classes.
* Added E2E test scenarios to verify that the expected errors are discarded, and that others are not.

### DIFF
--- a/Bugsnag/include/Bugsnag/BugsnagConfiguration.h
+++ b/Bugsnag/include/Bugsnag/BugsnagConfiguration.h
@@ -150,8 +150,10 @@ typedef BOOL (^BugsnagOnSessionBlock)(BugsnagSession *_Nonnull session);
 @property(readwrite, retain, nullable) NSSet<id> *redactedKeys;
 
 /**
- * A set of strings and NSRegularExpression objects that determines which errors should
+ * A set of strings and / or NSRegularExpression objects that determine which errors should
  * be discarded based on their `errorClass`.
+ *
+ * Comparisons are case sensitive.
  *
  * OnError / OnSendError blocks will not be called for discarded errors.
  *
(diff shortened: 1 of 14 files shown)


Bugsnag/include/Bugsnag/BugsnagConfiguration.h
```diff
@@ -149,6 +149,18 @@ typedef BOOL (^BugsnagOnSessionBlock)(BugsnagSession *_Nonnull session);
...
+ * A set of strings and NSRegularExpression objects that determines which errors should
+ * be discarded based on their `errorClass`.
+ *
+ * OnError / OnSendError blocks will not be called for discarded errors.
+ *
+ * Some examples of errorClass are: Objective-C exception names like "NSRangeException",
+ * signal names like "SIGABRT", mach exception names like "EXC_BREAKPOINT", and Swift
+ * error names like "Fatal error".
+ */
+@property(readwrite, copy, nullable) NSSet<id> *discardClasses;
```
kstenerud, line 162: Should probably mention that comparisons are case sensitive, since people will invariably get this wrong.

kstenerud (approved):

kattrali (approved):

PR #944, opened 2020-12-15
Wait for network connectivity before running E2E scenario

nickdowell:
## Goal

A small proportion of E2E test runs fail to receive a request payload when running the first scenario.

This change aims to make it easier to diagnose these failures by checking for connectivity to Maze Runner before running each test scenario, and including relevant messages in the device logs.

## Design

The pre-flight step works by performing an HTTP GET request on the notify endpoint, and retrying indefinitely until a success status code is received.

The GET request is not logged by Maze Runner and so does not get added to its lists of requests that would alter the test scenarios.

An earlier implementation using low-level TCP sockets failed because connectivity is provided through an HTTP proxy.

## Changeset

Added a pre-flight step that runs before each scenario.

In normal conditions this does not measurably impact the running time of the tests.

## Testing

Running E2E tests

### DIFF
--- a/features/fixtures/shared/scenarios/Scenario.m
+++ b/features/fixtures/shared/scenarios/Scenario.m
@@ -45,10 +45,36 @@ void markErrorHandledCallback(const BSG_KSCrashReportWriter *writer) {
     return self;
 }
 
+- (void)waitForNetworkConnectivity {
+    NSDictionary *proxySettings = (__bridge_transfer NSDictionary *)CFNetworkCopySystemProxySettings();
+    NSLog(@"*** Proxy settings = %@", proxySettings);
+    
+    // This check uses HTTP rather than sockets because connectivity is commonly provided via an HTTP proxy.
+    
+    NSURL *url = [NSURL URLWithString:self.config.endpoints.notify];
+    NSURLRequest *request = [NSURLRequest requestWithURL:url cachePolicy:NSURLRequestReloadIgnoringLocalCacheData timeoutInterval:3];
+    NSLog(@"*** Checking for connectivity to %@", url);
+    while (1) {
+        NSURLResponse *response = nil;
+        NSError *error = nil;
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+        [NSURLConnection sendSynchronousRequest:request returningResponse:&response error:&error];
+#pragma clang diagnostic pop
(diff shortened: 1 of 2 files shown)


features/fixtures/shared/scenarios/Scenario.m
```diff
@@ -45,10 +45,36 @@ - (instancetype)initWithConfig:(BugsnagConfiguration *)config {
     return self;
 }
 
+- (void)waitForNetworkConnectivity {
+    NSDictionary *proxySettings = (__bridge_transfer NSDictionary *)CFNetworkCopySystemProxySettings();
+    NSLog(@"*** Proxy settings = %@", proxySettings);
+    
+    // BrowserStack appears to use a proxy to establish connectivity with Maze Runner...
```
twometresteve, line 52: I'm not sure we need to mention BrowserStack at all here.  You're using just the same protocol and URL as the notifier does, so it seems like a separate issue to me (in fact I don't think it's an issue at all, just a feature of the test environment).  If/when we move device farm the comment will be redundant anyway.

twometresteve (approved):
I think this is a reasonable change and might save some failures (or at least provide better diagnostics when a step fails).

features/fixtures/shared/scenarios/Scenario.m
```diff
@@ -45,10 +45,36 @@ - (instancetype)initWithConfig:(BugsnagConfiguration *)config {
...
+        NSURLResponse *response = nil;
+        NSError *error = nil;
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+        [NSURLConnection sendSynchronousRequest:request returningResponse:&response error:&error];
+#pragma clang diagnostic pop
+        NSLog(@"*** %@", response ?: error);
+        if ([response isKindOfClass:[NSHTTPURLResponse class]] && ((NSHTTPURLResponse *)response).statusCode / 100 == 2) {
+            break;
+        }
```
twometresteve, line 68: Could we consider putting an `else` on this to log something like "MazeRunner probe failed, retrying in 1 second"?  We could then gain easier insight into how frequently it actually helps us out.

nickdowell: Done 👍

features/fixtures/shared/scenarios/Scenario.m
nickdowell: Updated the comment

twometresteve (approved):
LGTM

PR #946, opened 2020-12-17
[PLAT-5493] Move 10.15 requiring unit tests to buildkite

Cawllec:
## Goal

Moves unit tests that require MacOS 10.15 to run to buildkite away from travis.  Other unit tests will be moved across in separate PRs.

## Design

Unit tests have been split into separate test runs with a 3 concurrency limit to ensure they don't use all of our build machines.

## Questions going forward
- We had tests for MacOS 10.15.5 & 10.15.7 on travis.  Are these both required still for good coverage, or can we just test against one 10.15 version?
- Are there any unit test sets we can see here which are less useful to run and can be removed?  For example older tvOS versions etc.
- Would running tests on the same box sequentially (as in a single build step) be practical in terms of parsing test results?  This would allow us to take up less machines total, while saving some runtime on the step setup and bootstrap process.


Cawllec:
Note - Manually stopped e2e tests as I don't want to take up required BS slots.  We. can look at bringing those back up when we've got more availability.

nickdowell:
> We had tests for MacOS 10.15.5 & 10.15.7 on travis. Are these both required still for good coverage, or can we just test against one 10.15 version?

A single version of 10.15 is sufficient 👍 

> Are there any unit test sets we can see here which are less useful to run and can be removed? For example older tvOS versions etc.

At a minimum I think we'd always want to keep running tests on the oldest OS versions as well as the newest. The in-between versions could potentially be dropped, although I think we'd still want to run on those prior to a release.

> Would running tests on the same box sequentially (as in a single build step) be practical in terms of parsing test results? This would allow us to take up less machines total, while saving some runtime on the step setup and bootstrap process.

It's easier to read the test results when we stick to a single test run per job, but if we could get the BuildKite logs to collapse the green test runs then we'd have something almost as easy to grok...

Cawllec:
Closed in favour of #949

PR #948, opened 2020-12-18
[PLAT-2016] Fix Swift fatal error message reporting

nickdowell:
## Goal

Improve the reliability of capturing Swift fatal error messages passed to the following functions:
* `assertionFailure()`
* `fatalError()`
* `preconditionFailure()`

It also covers implicit messages emitted by the Swift standard library, e.g.

`Unexpectedly found nil while unwrapping an Optional value`

As a bonus, the same mechanism also captures error messages from other libraries such as libdispatch, e.g.

`BUG IN CLIENT OF LIBDISPATCH: dispatch_sync called on queue already owned by current thread`

## Design

The implementation examines the binary image of the crashing thread's top stack frame for a message in its `__crash_info` section, and parses this to extract the error type and message.

The full, raw value of the `__crash_info` message is also made available in the event metadata, which may be useful when diagnosing certain types of crash.

Screenshot 2020-12-17 at 16 28 39

## Changeset

* Updated KSCrash to capture the `__crash_info` message and store it in the KSCrashReport.
* Updated `BugsnagError` and `BugsnagEvent` to read the message stored in the KSCrashReport, parse it, and update the error & event payload.
* Removed code for old mechanism (`RegisterErrorData`)
* Disabled KSCrash's memory introspection option.

## Testing

* Manually verified with example app using iPhone Simulator and real devices, including 32-bit CPUs.
* Added new unit test case to validate parsing of crash info message.
* Updated E2E scenarios to validate that Swift error messages are reported correctly.
* Added new E2E scenario for libdispatch crash.


Bugsnag/Payload/BugsnagError.m
```diff
@@ -132,6 +128,33 @@ + (BugsnagError *)errorFromJson:(NSDictionary *)json {
     return error;
 }
 
+- (void)updateWithCrashInfoMessage:(NSString *)crashInfoMessage {
+    @try {
+        // Messages that match this pattern should override the errorClass (and errorMessage if there is enough information.)
+        NSString *pattern = @"^(Assertion failed|Fatal error|Precondition failed): ((.+): )?file .+, line \\d+\n$";
+        NSRegularExpression *regex = [[NSRegularExpression alloc] initWithPattern:pattern options:0 error:nil];
```
kattrali, line 135: This should be [case insensitive](https://github.com/bugsnag/bugsnag-cocoa/commit/add71b3d91b672d5cb8db540f189f1529bdcc817) to be compatible with every version of swift.

```suggestion
        NSRegularExpression *regex = [[NSRegularExpression alloc] initWithPattern:pattern options:NSRegularExpressionCaseInsensitive error:nil];
```

https://github.com/apple/swift/commit/d03a575279cf5c523779ef68f8d7903f09ba901e

kattrali (approved):

sethfri:
This is so awesome! Thanks for your work on this @nickdowell. Since this covers force unwrapping optionals, do you expect this to also capture the `Error` thrown from a `try!`, described in #950?

nickdowell:
@sethfri I am not about the errors thrown from a `try!` - will be investigating 🤓

nickdowell:
@sethfri quick update: a description of the `Error` is included in the message when `try!` causes a crash 😄  Example:

Screenshot 2020-12-21 at 11 59 53

Bugsnag/Payload/BugsnagError.m
```diff
@@ -132,6 +128,33 @@ + (BugsnagError *)errorFromJson:(NSDictionary *)json {
...
 }
 
+- (void)updateWithCrashInfoMessage:(NSString *)crashInfoMessage {
+    @try {
+        // Messages that match this pattern should override the errorClass (and errorMessage if there is enough information.)
+        NSString *pattern = @"^(Assertion failed|Fatal error|Precondition failed): ((.+): )?file .+, line \\d+\n$";
+        NSRegularExpression *regex = [[NSRegularExpression alloc] initWithPattern:pattern options:NSRegularExpressionCaseInsensitive error:nil];
+        NSArray<NSTextCheckingResult *> *matches = [regex matchesInString:crashInfoMessage options:0 range:NSMakeRange(0, crashInfoMessage.length)];
+        if (matches.count != 1 || matches[0].numberOfRanges != 4) {
+            if (!self.errorMessage.length) {
```
kstenerud, line 138: Maybe put this defensive self.errorMessage setter code at the very top so that an unexpected exception in the regex part above can't leave it potentially empty?

nickdowell: Addressed in https://github.com/bugsnag/bugsnag-cocoa/commit/8e832e8fb72006ae52213a7f1eb67c1b71c214a5

I could not move this code to before the regex part at it would have led to setting the `errorMessage` in cases where it should not be, so there's now additional code in the `@catch` block.

kstenerud (approved):

PR #949, opened 2020-12-18
[PLAT-5434, PLAT-5493] Move unit tests from Travis to buildkite

Cawllec:
## Goal

Moves all existing unit tests from Travis jobs to Buildkite.

Supplants #946, although I'll bring over the discussion from there. 

The Carthage build remains on Travis for now as some work is required to make it compatible with Xcode 12.


Cawllec:
Discussion brought across:

> We had tests for MacOS 10.15.5 & 10.15.7 on travis. Are these both required still for good coverage, or can we just test against one 10.15 version?

>> A single version of 10.15 is sufficient 👍

> Are there any unit test sets we can see here which are less useful to run and can be removed? For example older tvOS versions etc.

>> At a minimum I think we'd always want to keep running tests on the oldest OS versions as well as the newest. The in-between versions could potentially be dropped, although I think we'd still want to run on those prior to a release.

> Would running tests on the same box sequentially (as in a single build step) be practical in terms of parsing test results? This would allow us to take up less machines total, while saving some runtime on the step setup and bootstrap process.

>> It's easier to read the test results when we stick to a single test run per job, but if we could get the BuildKite logs to collapse the green test runs then we'd have something almost as easy to grok...

Cawllec:
@nickdowell Can I leave it to you to resolve any merge conflicts with `next` during review?

nickdowell (approved):

twometresteve (approved):
LGTM

PR #952, opened 2020-12-21
[PLAT-5649] Fix flake in BareboneTestHandledScenario

nickdowell:
## Goal

This scenario was flaking due to a timing issue.

## Changeset

Replaced fixed-length "sleep" with a block-based approach that removes timing sensitivity.

## Testing

Ran the scenario manually.


twometresteve (approved):
LGTM pending CI.

PR #957, opened 2021-01-04
[PLAT-3989] Report framework size in pull requests

nickdowell:
## Goal

To allow gauging the impact of any given PR on the binary size.

## Design

Following the methodology used in [bugsnag-js](https://github.com/bugsnag/bugsnag-js/blob/next/.github/workflows/pr-size.yml); builds the framework with the target and source branches and reports the binary sizes as a comment in the PR (scroll down to see.)

Only the arm64 architecture for iOS is built, in order to give consistent results that will not change if/when armv7 is dropped.

Implemented as a GitHub Action as this simplifies authentication for reporting results to the PR using Danger.

## Changeset

Added a GitHub Action, an associated Dangerfile, and added Danger to the Ruby dependencies.

## Testing

Tested via this PR - [comment](https://github.com/bugsnag/bugsnag-cocoa/pull/957#issuecomment-753939788) shows 0 change in binary size as expected.

Tested locally by altering contents of .size files and verifying output; Danger can be run locally via
```
bundle exec danger dry_run --dangerfile=.github/workflows/framework-size.danger
```


kattrali (approved):
neat!

PR #959, opened 2021-01-05
[PLAT-5627] Refactor CI pipeline into multistage builds

Cawllec:
## Goal

Refactors CI pipeline into:
- Barebones - The default that runs when no other conditions are met.  Minimum and Maximum supported OS versions are tested against in both Unit and E2E tests. Static analysis is run. Minimal E2E tests are run.
- Quick - Only run when "[Quick CI]" is present in the commit message.  Identical to the above, but covers all supported OS versions.
- Integration - Triggered when a merge is targeting `next` or `master` branches, or "[integration ci]" is present in the commit message.  Runs the same steps as the Quick run by default, with manual trigger to run the remainder of tests that would be present in the full test run.
- Full - Runs if the branch is `next` or `master`, or "[full ci]" is present in the commit message.  Runs all tests against all OS versions, including the full scope of E2E tests.


.buildkite/pipeline.yml
```diff
@@ -358,6 +144,10 @@ steps:
       docker-compose#v3.3.0:
         run: cocoa-maze-runner
         command:
+          - "features/app_and_device_attributes.feature"
```
nickdowell, line 147: Assuming this is a "barebones" test run, it should use only `features/barebone_tests.feature`

nickdowell (changes requested):

.buildkite/pipeline.yml
Cawllec: Changed that over in the shortest test runs.

twometresteve (commented):
LGTM if @nickdowell is happy.  We can alway tweak if necessary.

nickdowell (approved):

PR #960, opened 2021-01-05
V1 of new filesystem layout

kstenerud:
## Goal

https://bugsnag.atlassian.net/wiki/spaces/~981233658/pages/1842872321/Cocoa+notifier+file+storage+v1+proposal

## Design

`BugsnagFileLocations` is now the single source of truth about where files live. All other classes must use this when loading or saving files.

It's versioned, although the actual implementation is hardcoded to V1 since we have no incompatible changes yet. When an incompatible change occurs, we can add the interface/implementation details. The high level interface should not change.

## Changeset

Anything that previously grabbed top-level directories or took a path as an initialization field now uses BugsnagFileLocations. This required a number of files to change.

## Testing

New unit test with sample data to test migration and usage of the new file locations.


Bugsnag/Helpers/BSGStorageMigratorV0V1.h
```diff
@@ -0,0 +1,19 @@
...
+//  Copyright © 2021 Bugsnag Inc. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface BSGStorageMigratorV0V1 : NSObject
+
++ (bool) migrate;
```
nickdowell, line 15: In Objective-C we use `BOOL` not `bool`

Bugsnag/Client/BugsnagClient.m
```diff
@@ -66,6 +65,7 @@
 #import "BugsnagSystemState.h"
 #import "BugsnagThread+Private.h"
 #import "BugsnagUser+Private.h"
+#import "BSGFileLocations.h"
```
nickdowell, line 68: It would be nice to keep this list of imports sorted alphabetically

Bugsnag/Storage/BSGFileLocations.m
```diff
@@ -0,0 +1,96 @@
...
+    if (ensureDirExists(subdirPath)) {
+        return subdirPath;
+    }
+    // Make the best of it, just return the root dir.
+    return rootPath;
+}
+
+@interface BSGFileLocations ()
+
+@property(readonly) NSString *root;
```
nickdowell, line 64: Seems like this property is unnecessary

Bugsnag/Storage/BSGFileLocations.m
```diff
@@ -0,0 +1,96 @@
...
+        // If we can't even create the root dir, all is lost, and no file ops can be allowed.
+        if(!ensureDirExists(rootPath)) {
+            rootPath = defaultRootPath;
+        }
+    });
+
+    return rootPath;
+}
+
+static NSString *getSubdir(NSString *rootPath, NSString *relativePath) {
```
nickdowell, line 53: `getSubdir` does not really describe what this function does - it creates the directory...

nickdowell:
I suggest that the `Storage` group / folder would be a better home for `BSGStorageMigratorV0V1` than `Helpers`

nickdowell:
After running the migration via the example app, I noticed there are a few empty directories left behind afterwards.

Could we make the migrator delete these?

```
Library/Caches/Sessions
Library/Caches/KSCrashReports
Library/Caches/bugsnag
Library/Caches/bugsnag/state
```

Tests/BSGStorageMigratorTests.m
```diff
@@ -0,0 +1,196 @@
...
+    [BSGStorageMigratorV0V1 migrate];
+    [BSGStorageMigratorV0V1 migrate];
+    [BSGStorageMigratorV0V1 migrate];
+    [BSGStorageMigratorV0V1 migrate];
+    [BSGStorageMigratorV0V1 migrate];
+    [self postMigrateCheck];
+
+    // Make sure the client still loads and files are still as we expect.
+    BugsnagConfiguration *configuration = [[BugsnagConfiguration alloc] initWithApiKey:DUMMY_APIKEY_32CHAR_1];
+    (void)[[BugsnagClient alloc] initWithConfiguration:configuration];
```
nickdowell, line 149: I don't think it's a good idea to instantiate the BugsnagClient in the BSGStorageMigratorTests

BugsnagClient is quite a heavy object to instantiate, registers itself as an observer for various notifications, and writes to disk based on those notifications. Combined with the fact that BugsnagClient objects never deallocate (possible retain cycle) means that instantiating them in tests can cause weird issues.

nickdowell:
`BSGStorageMigratorTests.m` needs to be added to the Bugsnag-macOSTests and Bugsnag-tvOSTests targets to allow unit testing on those platforms.

Bugsnag/Client/BugsnagClient.m
```diff
@@ -246,22 +246,21 @@ - (instancetype)initWithConfiguration:(BugsnagConfiguration *)configuration {
         self.notifier = [BugsnagNotifier new];
         self.systemState = [[BugsnagSystemState alloc] initWithConfiguration:self.configuration];
 
-        NSString *cachesDir = [BSGCachesDirectory cachesDirectory];
-        NSString *bugsnagDir = [cachesDir stringByAppendingPathComponent:@"bugsnag"];
-        [[NSFileManager defaultManager] createDirectoryAtPath:bugsnagDir withIntermediateDirectories:YES attributes:nil error:nil];
+        BSGFileLocations *fileLocations = [BSGFileLocations current];
+//        NSString *generalDir = [BugsnagFileLocations Current].general;
```
nickdowell, line 250: I think you probably didn't mean to commit this line 😅

nickdowell (changes requested):

Bugsnag/Storage/BSGStorageMigratorV0V1.m
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    for(NSString *path in @[
+        [cachesDir stringByAppendingPathComponent:@"bsg_kvstore"],
+        [cachesDir stringByAppendingPathComponent:@"bugsnag"],
+        [cachesDir stringByAppendingPathComponent:@"KSCrashReports"],
+        [cachesDir stringByAppendingPathComponent:@"Sessions"],
+                          ]) {
+        if(![fm removeItemAtPath:path error:&err]) {
+            bsg_log_err(@"Could not remove %@: %@", path, err);
```
nickdowell, line 78: When I trigger the migrator using one of the example apps, I see this error for `bsg_kvstore`.

```
2021-01-11 08:53:48.256886+0000 Bugsnag Test App[70652:2466634] [Bugsnag] [ERROR] Could not remove /Users/nick/Library/Developer/CoreSimulator/Devices/5AFE2FCA-EB57-45D2-A705-42F81D4031F3/data/Containers/Data/Application/4AE2D40C-55E3-434B-9E78-9DEB080BF5FE/Library/Caches/bsg_kvstore: Error Domain=NSCocoaErrorDomain Code=4 "“bsg_kvstore” couldn’t be removed." UserInfo={NSUserStringVariant=(
    Remove
), NSFilePath=/Users/nick/Library/Developer/CoreSimulator/Devices/5AFE2FCA-EB57-45D2-A705-42F81D4031F3/data/Containers/Data/Application/4AE2D40C-55E3-434B-9E78-9DEB080BF5FE/Library/Caches/bsg_kvstore, NSUnderlyingError=0x600000b52280 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
```

I suggest checking the error code before logging:
```suggestion
            if (!([err.domain isEqual:NSCocoaErrorDomain] && err.code == NSFileNoSuchFileError)) {
                bsg_log_err(@"Could not remove %@: %@", path, err);
            }
```

nickdowell (changes requested):

nickdowell (approved):

PR #962, opened 2021-01-07
[PLAT-5669] Move Carthage to BK, ensuring it uses Xcode 11

Cawllec:
## Goal

Moves away from Travis by moving remaining tests to buildkite, ensuring they'll work correctly by forcing the usage of a compatible version of Xcode.

Targeting the barebones-pipeline branch as a base, as both will potentially require some updating of github status settings before merging.

- Moves Carthage static build to buildkite
- Uses DEVELOPER_DIR environment variable to enforce Xcode 11 usage
- Removes Travis CI file


.buildkite/pipeline.yml
```diff
@@ -24,6 +24,19 @@ steps:
...
 
+  - label: Static Carthage build
+    timeout_in_minutes: 10
+    agents:
+      queue: opensource-mac-cocoa
+    concurrency: 3
+    concurrency_group: cocoa-unit-tests
+    env:
+      LANG: "en_GB.UTF-8"
+      DEVELOPER_DIR: "/Applications/Xcode.app/Contents/Developer"
```
nickdowell, line 35: It would be clearer if the Xcode path included the version, i.e. `/Applications/Xcode_11.7.app`

FYI this is the naming convention [used by GitHub Actions.](https://github.com/actions/virtual-environments/blob/macOS-10.15/20201212.1/images/macos/macos-10.15-Readme.md#xcode)

twometresteve:
image

This I like. :-)

Nick's suggestion above seems reasonable - presumably it just a matter of jumping on the box(es) quickly and renaming the folder?  Anyway, if he's happy then I'm happy.

Cawllec:
Do we know that changing the directory name won't break other things relying on the Xcode tools?

nickdowell:
> Do we know that changing the directory name won't break other things relying on the Xcode tools?

We don't have any hard-coded paths to Xcode so as long as `xcode-select -p` returns a valid path nothing should break.

PR #965, opened 2021-01-12
Allow 75 minutes for E2E test run

nickdowell:
## Goal

The full E2E test runs sometimes fail due to not completing within the 60 minutes allowed.

E.g. iOS 12 & 13 in https://buildkite.com/bugsnag/bugsnag-cocoa/builds/1647

## Changeset

* Increases the timeout to 75 minutes for full E2E
* Makes all unit test jobs use a 10 minute timeout

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -2,7 +2,7 @@ steps:
   - label: ':ios: iOS 13 end-to-end tests'
     depends_on:
       - cocoa_fixture
-    timeout_in_minutes: 60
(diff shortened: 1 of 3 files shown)


twometresteve (approved):
Change LGTM, but do you want to put auto retries for lost agents of the steps that have been recently added at the same time?

nickdowell:
> Change LGTM, but do you want to put auto retries for lost agents of the steps that have been recently added at the same time?

It looks to me like this has already been done

Cawllec (approved):

PR #967, opened 2021-01-12
Build status badge: replace Travis with BuildKite

nickdowell:
## Goal

We no longer use Travis, so its build status badge no longer reflects the status of our builds.


twometresteve (approved):
LGTM

PR #970, opened 2021-01-14
[PLAT-5704] Fix "Update documentation page" build step on CI

nickdowell:
## Goal

The "Update documentation page" build step is currently not being run on BuildKite, because when the build triggers there is no Git tag associated with the branch (merges onto `master` happen via GitHub PRs, tags get created by the developer running `make release` some time afterwards.)

## Changeset

Removes the `build.tag` condition.

## Testing

Not tested. Will need to validate this during the next release.

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -77,7 +77,7 @@ steps:
           limit: 2
 
   - label: 'Update documentation page'
-    if: build.tag =~ /^v[2-9]\.[0-9]+\.[0-9]+\$/ && build.branch == "master"
+    if: build.branch == "master"
     agents:
       queue: opensource-mac-cocoa
     concurrency: 3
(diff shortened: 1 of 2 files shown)


twometresteve:
What's the reason for/benefit of this?  I'm not convinced it's what you want to be doing.... the Makefile appears to explicitly check that a tag is present:
```
ifeq ($(BUILDKITE_TAG),)
	@$(error Docs deployments should only occur on a tagged release)
endif
```

twometresteve:
> What's the reason for/benefit of this? I'm not convinced it's what you want to be doing.... the Makefile appears to explicitly check that a tag is present (and I can see other reasons for not doing it):
> 
> ```
> ifeq ($(BUILDKITE_TAG),)
> 	@$(error Docs deployments should only occur on a tagged release)
> endif
> ```

Ok, I've talked to @kattrali in the meantime and we think that pinning docs updates to `master` is reasonable given our workflow.  But the `Makefile` relies on `BUILDKITE_TAG`, which won't be present.  I'm not sure what the right change is, though, and a bit of investigation might be needed.

nickdowell:
@twometresteve the Makefile has now been updated to account for `BUILDKITE_TAG` not being set

kattrali (approved):

PR #973, opened 2021-01-15
build: fix warnings with -Wpedantic

keith:
This fixes a few cases of returning void functions which causes a
warning when building with -Wpedantic.

### DIFF
--- a/Bugsnag/BugsnagSystemState.m
+++ b/Bugsnag/BugsnagSystemState.m
@@ -224,7 +224,8 @@ static NSDictionary *copyDictionary(NSDictionary *launchState) {
 
 - (void)sessionUpdateNotification:(NSNotification *)notification {
     if (![BSGJSONSerialization isValidJSONObject:notification.object]) {
-        return bsg_log_err("Invalid session payload in notification");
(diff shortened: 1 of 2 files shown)


keith:
looks like the github actions failure is a danger permissions error

nickdowell:
> looks like the github actions failure is a danger permissions error

Correct; for forks `GITHUB_TOKEN` [does not grant write access](https://github.community/t/optional-read-write-permission-for-forked-repos/16330/2) so our Danger step is unable to update the PR.

nickdowell:
@keith thanks for the PR. Are these the only warnings you see when building with -Wpedantic ?

keith:
There are some more, I'm happy to submit a PR enabling these in the project if you'd like! Here's what I was using:

https://github.com/bugsnag/bugsnag-cocoa/pull/586#issuecomment-626995761

nickdowell (approved):

mattdyoung:
Released in [v6.6.1](https://github.com/bugsnag/bugsnag-cocoa/releases/tag/v6.6.1)

PR #976, opened 2021-01-19
Fix uninstalling signal handlers

blaztinn:
## Goal

Allow other signal handlers besides bugsnag's.

## Design

When trying bugsnag I noticed that my custom signal handlers weren't
triggered. After debugging I found out that bugsnag catches a mach
exception and uninstalls both mach and signal handlers
(`bsg_kscrashsentry_uninstall(BSG_KSCrashTypeAsyncSafe)`).

The problem is that when uninstalling bugsnag's signal handler it resets to the 
previously installed handler, this results in removing all the signal handlers
installed after bugsnag's (which was the case for my signal handler).

To solve this, disable signal handlers instead of uninstalling them. Do it this
way because:

1. Signal handlers installed after bugsnag's can take bugsnag's signal handling
function address and call it. In this case the function must always work and
chain calls forward.
2. If multiple signal handlers for same signal are installed after bugsnag's we
can't remove bugsnag's without removing all but the last installed one.

Because we don't uninstall bugsnag's signal handlers we now always directly
call previously installed signal handlers inside the bugsnag's signal handler.

## Changeset

Changed that signal handler is always installed once installed. On uninstall
it gets disabled, where it doesn't handle the crash but only forwards the call
to previous signal handlers.

## Testing

After the change both bugsnag and my custom handler handled the crash.

[Test-Bugsnag-iOS-2021.01.19_08-19-20-+0100.xcresult.zip](https://github.com/bugsnag/bugsnag-cocoa/files/5834271/Test-Bugsnag-iOS-2021.01.19_08-19-20-%2B0100.xcresult.zip)

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_Signal.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_Signal.c
@@ -50,6 +50,14 @@
  */
 static volatile sig_atomic_t bsg_g_installed = 0;
 
+/** Flag noting if we should handle signals.
+ * When other signal handlers are registered after ours we can't remove our
+ * signal handlers without removing the others.
+ * It's not fully thread safe, but it's safer than locking and slightly better
+ * than nothing.
+ */
+static volatile sig_atomic_t bsg_g_enabled = 0;
+
 #if !TARGET_OS_TV
 /** Our custom signal stack. The signal handler will use this as its stack. */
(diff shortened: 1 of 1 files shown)


xander-jones:
Thanks for raising this PR, we like the idea here. We'll review this as time allows, and we'll keep you updated on this thread.

nickdowell (approved):

yousif-bugsnag:
Fixed in [v6.6.3](https://github.com/bugsnag/bugsnag-cocoa/releases/tag/v6.6.3)

PR #978, opened 2021-01-20
Make OOMScenario less flakey

nickdowell:
## Goal

A [flake](https://buildkite.com/bugsnag/bugsnag-cocoa/builds/1720#53de7be7-2e6d-44c8-a464-2a79af6c7949) was noticed in the new `OOMScenario` - caused by the scenario triggering the OOM so quickly that the notifier does not have time to write breadcrumbs and metadata to disk.

## Design

Investigated approaches based on checking amount of free memory, but did not find an approach that would work on all OS versions.

## Changeset

The memory consumption is temporarily paused when a memory warning is received.

## Testing

Manually tested.


kattrali (approved):

twometresteve (approved):
Concept LGTM.

PR #980, opened 2021-01-20
[PLAT-5731] Improve Header docs readability

nickdowell:
## Goal

The [Header docs](http://bugsnag.github.io/bugsnag-cocoa/Bugsnag_h/index.html) for bugsnag-cocoa have readability issues when it comes to Objective-C properties.

The tool being used to generate them no longer appears to be maintained so it is necessary to find a better alternative.

## Design

[Jazzy](https://github.com/realm/jazzy) is a documentation generation tool created and actively maintained by Realm. It is used by several popular open-source libraries, including for documenting [Realm's own library](https://realm.io/docs/swift/latest/api/).

The documentation it generates can shows Objective-C and Swift versions of each symbol, similar to Apple's documentation.

## Other tools considered

* [appledoc](https://github.com/tomaz/appledoc) is no longer actively maintained
* [Doxygen](https://www.doxygen.nl/index.html) is popular but is not focused on Objective-C / Swift and would not be able to show Swift generated interfaces. Its default theme is also not as smart as Jazzy's.

## Changeset

Some of the header files were updated to add missing documentation for classes, and fix the formatting of some existing ones to be parsed more correctly by Jazzy.

### Note

The "Header docs" link on https://docs.bugsnag.com/platforms/ios/ will need to be updated after deploying new documentation, since Jazzy uses a different file naming convention and `Bugsnag_h/index.html` will no longer exist.

## Testing

Tested by running `make docs` locally and opening `docs/index.html`.

Note that a `bundle install` is required before running `make docs`.

Generated output attached below for convenience.

## Demo

Screenshot 2021-01-20 at 15 24 59

[docs.zip](https://github.com/bugsnag/bugsnag-cocoa/files/5843452/docs.zip)

### DIFF
--- a/.github/workflows/update_docs.yml
+++ b/.github/workflows/update_docs.yml
@@ -1,8 +1,7 @@
 name: "Update Docs"
 on:
-  push:
-    branches:
-      - master
+  release:
+    types: [published]
 
 jobs:
   build:
--- a/.github/workflows/update_docs.yml
+++ b/.github/workflows/update_docs.yml
@@ -1,0 +1,45 @@
+name: "Update Docs"
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: macos-latest
+    steps:
+      - name: Checkout bugsnag-cocoa
+        uses: actions/checkout@v2
+
+      - name: Checkout docs branch
+        uses: actions/checkout@v2
+        with:
+          ref: gh-pages
+          path: docs
+
+      - name: Configure docs branch
+        working-directory: docs
+        run: |
+          git config user.name "Bugsnag Bot"
+          git config user.email notifiers@bugsnag.com
+
+      - uses: actions/cache@v2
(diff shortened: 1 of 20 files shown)


kstenerud (approved):

nickdowell:
## Update

The docs update CI job has been moved to GitHub Actions, because our BuildKite agent does not have write access to the repos.

.github/workflows/update_docs.yml
```diff
@@ -0,0 +1,45 @@
+name: "Update Docs"
+on:
+  push:
+    branches:
+      - master
```
kattrali, line 5: Since we no longer have to worry about BK restriction, this could be on [release events](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release) now

nickdowell:
Rebased to fix merge conflicts

kattrali (approved):

PR #981, opened 2021-01-21
[PLAT-5055] Support macOS E2E tests

Cawllec:
## Goal

Part 1 of the addition of MacOS e2e tests to CI by @nickdowell

Enables standard e2e test features to be run on MacOS through an appium server.

Tests that won't run in this format are tagged with `@skip_macos`, and will be skipped using `--tags='not @skip_macos'` in test invocations.

Part 2 will be a basic WIP pipeline running these tests using our existing infrastructure

PR #982, opened 2021-01-21
[PLAT-5055] Adds basic pipeline support for running Cocoa tests on MacOS

Cawllec:
## Goal

Will retarget `next` once #981 is merged

Adds:
- MacOS 10.15 barebones tests to the barebones pipeline
- MacOS 10.15 full tests to the quick pipeline
- MacOS 10.14 full tests to the full pipeline

TODO in this PR:
- Update to Maze-runner v4 once #981 is updated to use it
- Target tests at MacOS 11 (requires infrastructure changes)

TODO elsewhere:
- Target tests at MacOS 10.13 (currently failing due to issues that need debugging separately)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -179,7 +179,7 @@ steps:
         download: ["features/fixtures/macos/output/macOSTestApp.zip"]
     commands:
       - bundle install
-      - bundle exec bugsnag-maze-runner
+      - bundle exec maze-runner
         features/barebone_tests.feature
         --farm=local
         --os=macos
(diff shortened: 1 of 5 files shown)


twometresteve (approved):
Change LGTM, would just like to see a full build before it gets merged I think.

.buildkite/pipeline.yml
```diff
@@ -168,5 +168,25 @@ steps:
...
+        --farm=local
+        --os=macos
+        --os-version=10.15
+        --app=macOSTestApp
+        --tags='not @skip_macos'
+        --fail-fast
+
   - label: 'Conditionally trigger full set of tests'
-    command: sh -c .buildkite/pipeline_trigger.sh
+    command: sh -c .buildkite/pipeline_trigger.sh
```
nickdowell, line 192: Quite a few tools (linters etc.) complain about files that don't end with a newline.. probably better to leave them in 🙃

nickdowell (approved):

PR #983, opened 2021-01-22
[PLAT-5721] Run OCLint static analysis tool

nickdowell:
## Goal

Integrates the [OCLint](http://oclint.org/) static analytics tool into the CI workflow.

## Design

Integrated into the same GitHub Action that runs Infer and reports results via Danger, so that all results are shown in the same PR comment.

## Changeset

The GH push_request action has been refactored to generate `compile_commands.json` before running Infer and OCLint (which requires that file, unlike Infer where it's optional.)

A number of rules have been disabled due to false positives or flagging relatively harmless issues whose fixes would create a lot of whitespace changes that would reduce the effectiveness of `git blame` 👉 see `.oclint` for details.

Warnings raised for remaining rules have been fixed in this PR.

## Testing

Tested by running locally - `make oclint` and verifying the results reported in the PR.

To see the issues reported by OCLint, view the older version of the Danger PR comment

Screenshot 2021-01-25 at 09 27 53

### DIFF
--- a/Bugsnag/Client/BugsnagClient+Private.h
+++ b/Bugsnag/Client/BugsnagClient+Private.h
@@ -60,7 +60,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (strong, nonatomic) BugsnagSessionTracker *sessionTracker; // Used in BugsnagReactNative
 
-@property (readonly, nonatomic) BOOL started;
+@property (nonatomic) BOOL started;
 
 @property (strong, nonatomic) BugsnagMetadata *state;
 
--- a/Bugsnag/Client/BugsnagClient+Private.h
+++ b/Bugsnag/Client/BugsnagClient+Private.h
@@ -60,7 +60,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (strong, nonatomic) BugsnagSessionTracker *sessionTracker; // Used in BugsnagReactNative
 
-@property (nonatomic) BOOL started;
+@property (readonly, nonatomic) BOOL started;
 
 @property (strong, nonatomic) BugsnagMetadata *state;
 
--- a/Bugsnag/Configuration/BugsnagConfiguration+Private.h
+++ b/Bugsnag/Configuration/BugsnagConfiguration+Private.h
@@ -25,7 +25,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (readonly) NSDictionary<NSString *, id> *errorApiHeaders;
 
-@property (readonly, nonatomic) BugsnagMetadata *metadata;
+@property (readonly, copy, nonatomic) BugsnagMetadata *metadata;
 
 @property (nullable, nonatomic) NSURL *notifyURL;
 
--- a/Bugsnag/Configuration/BugsnagConfiguration+Private.h
+++ b/Bugsnag/Configuration/BugsnagConfiguration+Private.h
@@ -27,7 +27,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (readonly, nonatomic) BugsnagMetadata *metadata;
 
-@property (readonly, nullable, nonatomic) NSURL *notifyURL;
+@property (nullable, nonatomic) NSURL *notifyURL;
 
 @property (nonatomic) NSMutableArray<BugsnagOnBreadcrumbBlock> *onBreadcrumbBlocks;
 
@@ -42,7 +42,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (readonly) NSDictionary<NSString *, id> *sessionApiHeaders;
 
-@property (readonly, nullable, nonatomic) NSURL *sessionURL;
+@property (nullable, nonatomic) NSURL *sessionURL;
+
+@property (readwrite, retain, nonnull, nonatomic) BugsnagUser *user;
 
 #pragma mark Methods
 
--- a/Bugsnag/Configuration/BugsnagConfiguration+Private.h
+++ b/Bugsnag/Configuration/BugsnagConfiguration+Private.h
@@ -27,7 +27,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (readonly, copy, nonatomic) BugsnagMetadata *metadata;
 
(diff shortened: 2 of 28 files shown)


Bugsnag/Client/BugsnagClient+Private.h
```diff
@@ -60,7 +60,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (strong, nonatomic) BugsnagSessionTracker *sessionTracker; // Used in BugsnagReactNative
 
-@property (readonly, nonatomic) BOOL started;
```
kattrali, line 63: I get that it mutes the warning, but is it an improvement to make this property (and others) mutable to the rest of the library to avoid ivar access?

nickdowell: In this case, maybe not; setting it to NO would be very bad.

nickdowell: Resolved 👍

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSString.c
```diff
@@ -132,7 +132,7 @@ bool bsg_ksstring_extractHexValue(const char *string, size_t stringLength,
                 accum += nybble;
             }
             *result = accum;
-            return true;
+            return true; //!OCLINT
```
kattrali, line 135: Is this a rule about early returns or something? Is that something that should be generally adopted?

Bugsnag/Configuration/BugsnagConfiguration+Private.h
```diff
@@ -42,7 +42,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (readonly) NSDictionary<NSString *, id> *sessionApiHeaders;
 
-@property (readonly, nullable, nonatomic) NSURL *sessionURL;
+@property (nullable, nonatomic) NSURL *sessionURL;
```
kattrali, line 45: Wouldn't the URLs also be very bad to change?

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSString.c
kstenerud: Actually the code itself is kind of ugly; I don't think it's flagged just because of the early return. It could probably be written in a much clearer form that also wouldn't trigger a lint warning (in this case it's probably the return at the end of a for loop block).

nickdowell: The message from oclint in this case was "avoid branching statement as last in loop"

Bugsnag/Configuration/BugsnagConfiguration.m
```diff
@@ -92,7 +92,7 @@ - (nonnull id)copyWithZone:(nullable NSZone *)zone {
     [copy setMaxPersistedEvents:self.maxPersistedEvents];
     [copy setMaxPersistedSessions:self.maxPersistedSessions];
     [copy setMaxBreadcrumbs:self.maxBreadcrumbs];
-    copy->_metadata = [[BugsnagMetadata alloc] initWithDictionary:[[self.metadata toDictionary] mutableCopy]];
+    copy->_metadata = [[BugsnagMetadata alloc] initWithDictionary:[[self.metadata toDictionary] mutableCopy]]; //!OCLINT
```
kstenerud, line 95: Is this being triggered because of a direct ivar access?

nickdowell: Yes; fixing it "properly" would require a bigger change

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSString.c
kstenerud: Yeah, that's definitely from my weird-ass for loop: https://github.com/bugsnag/bugsnag-cocoa/blob/7295d66f49217de2ff639f013d984a0736a0da9b/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSString.c#L117

It's basically using breaks and continues and then does a return if it reaches the end of the block. It could definitely be written in a clearer, more standard way.

Bugsnag/Configuration/BugsnagConfiguration.m
kstenerud: I haven't looked deeply, but wouldn't this be a matter of an anonymous `BugsnagMetadata` category that grants write access to the property?

Maybe these things that require bigger changes could be put as a ticket so we don't lose track?

nickdowell: Because `BugsnagConfiguration+Private.h` contains the class extension (`@interface BugsnagConfiguration ()`) it's not possible to override that declaration in `BugsnagConfiguration.m`; the compiler will bail with `error: property has a previous declaration`

I think the clean fix it to add an `initWithConfiguration:` method that initializes by copying, and make `copyWithZone:` call into that.

nickdowell: Or simply implement `setMetadata:` 😅

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSString.c
nickdowell: This function has now been refactored

kstenerud (approved):

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSString.c
```diff
@@ -111,31 +111,34 @@ static const unsigned int bsg_g_hexConversion[] = {
...
-                accum += nybble;
-            }
-            *result = accum;
-            return true;
+    if (!stringLength) {
+        return false;
+    }
+    
+    const char *value = strnstr(string, "0x", stringLength);
+    if (!value) {
```
kattrali, line 119: this was previously hinted as an `unlikely_if`.

PR #984, opened 2021-01-26
Improve reliability of OOM scenario

nickdowell:
## Goal

Observed another [OOM flake](https://buildkite.com/bugsnag/bugsnag-cocoa/builds/1803#f501b03a-a674-4e39-94fb-5d83c654cffa) where it appears the low memory warning was not received / processed despite the previous changes. This PR aims to improve reliability of the test.

## Design

The scenario will now consume a large amount of memory (enough to trigger a warning) very quickly, and then consume the remainder in a slower fashion.

Everything now happens on the main thread, using a timer for the second phase of allocations to ensure that the run loop has a chance to run and process any memory warnings.

## Testing

Tested locally with 3 different devices, and by [repeatedly running the barebones tests](https://buildkite.com/bugsnag/bugsnag-cocoa/builds?branch=nickdowell%2Fimprove-oom-scenario-reliability) (which run OOMScenario)

### DIFF
--- a/examples/objective-c-ios/Bugsnag Test App/OutOfMemoryController.m
+++ b/examples/objective-c-ios/Bugsnag Test App/OutOfMemoryController.m
@@ -1,6 +1,18 @@
 #import "OutOfMemoryController.h"
 
-@implementation OutOfMemoryController
+#define PRINT_STATS 0
+
+#if PRINT_STATS
+#import <mach/mach_init.h>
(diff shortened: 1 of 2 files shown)


kstenerud (approved):

twometresteve (approved):
Theory sounds good to me, without understanding all the code.  Thanks for caring.  :-)

PR #992, opened 2021-02-05
[PLAT-5923] Improved thread recording

nickdowell:
## Goals

* Enable recording the main thread's stack trace (required for some upcoming functionality.)
* Reduce overhead when logging unhandled errors.

## Changeset

Added 2 new (private) methods on `BugsnagThread`:
* `+allThreads:callStackReturnAddresses:` (replaces `-[BSG_KSCrash captureThreads:depth:recordAllThreads:]`)
* `+mainThread`

Other threads are now suspended for a shorter period of time while their backtraces are being captured, and resumed before symbolication starts.

## Testing

Manually tested to verify correctness and measure performance improvement.

Using a sample app, performance was measured by calling `notify()` 100 times in a loop to obtain reproducible figures for the execution time. The results were:

Thread send policy | Before | After
-- | -- | --
`.always` | 48ms | 7ms
`.unhandledOnly` | 43ms | 6ms

Unit test cases have been added to verify the new functionality.


Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMach.c
```diff
@@ -196,8 +196,11 @@ bool bsg_ksmachfillState(const thread_t thread, const thread_state_t state,
     mach_msg_type_number_t stateCountBuff = stateCount;
     kern_return_t kr;
 
+    // For unknown reasons, thread_get_state returns this value on Rosetta,
+    // but still succeeds.
+    const int ROSETTA_SUCCESS = 268435459;
```
kstenerud, line 201: Maybe write it as hex: 0x10000003?

Since `thread_get_state()` returns type `kern_return_t`, I think it's actually returning `MACH_SEND_INVALID_DEST`?

https://opensource.apple.com/source/xnu/xnu-792/osfmk/mach/message.h

kstenerud (changes requested):

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMach.c
nickdowell: I copied this verbatim [from Crashlytics](https://github.com/firebase/firebase-ios-sdk/blob/739d21a6234d20e28cfd124d1759035cfe0d1a2b/Crashlytics/Crashlytics/Components/FIRCLSProcess.c#L220-L222) and left it in its original form, but happy to change if we think it's better to do so.

kstenerud: The Crashlytics code and explanation doesn't fill me with confidence... I'd feel a lot better if we knew more about what was going on.

At the very least, it looks like it's getting an invalid destination port when sending this particular mach message over Rosetta, which is entirely plausible considering the emulation layers involved (perhaps the virtual port is getting deallocated early due to a bug...). Are we able to verify this behaviour?

nickdowell: I agree it would be best to verify this behaviour. It's not clear from the comment whether this relates to Rosetta from 2006 or 2020!

nickdowell: I have checked this behaviour on the Developer Transition Kit and confirmed that `thread_get_state` returns `268435459` when running Intel apps under Rosetta 2 and that despite this we are able to capture stack traces for the threads.

Screen Sharing Picture 15 February 2021 at 08 44 23 GMT copy

nickdowell: Update: after further investigation I have determined that valid data is **not** returned under Rosetta 2 when we encounter `MACH_SEND_INVALID_DEST`.

In my testing using the "Caught Exception Notify" function in the objective-c-osx sample app, it's consistently only the first two non-main threads that fail.

kstenerud: 👍

kstenerud (approved):

PR #995, opened 2021-02-09
Review Appium version use on iOS 10 and 11

twometresteve:
## Goal

Use an earlier version of Appium (1.7.0) on iOS 10 and 11 to avoid a bug in 1.15/1.16 that causes session instability.

## Changeset

During testing some devices showed errors closing the keyboard, so I've relaxed the behaviour to ignore any cases of the keyboard not being present in the first place.

I've removed use of the resilient driver on iOS 10/11 to allow us more readily form a picture of how the change performs over a series of scheduled overnight builds.

## Testing

Covered by CI.  Running with Appium 1.7.0 against all 11 iOS devices available on BrowserStack yielded good results, although we will continue to monitor the situation,

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -1,5 +1,5 @@
 steps:
-  - label: ':ios: iOS 13 end-to-end tests'
+  - label: ':ios: iOS 13 full end-to-end tests'
     depends_on:
       - cocoa_fixture
     timeout_in_minutes: 120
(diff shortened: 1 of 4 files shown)


nickdowell (approved):

twometresteve:
@nickdowell I've now split the iOS 10 e2e run into two batches by using the Cucumber `-e` (exclude pattern) option.  I'm not sure why, but the test device seemed to be timing out at the same point each time (approx 55 mins in).  There doesn't seem to be anything wrong with the scenario it was failing on, as it passes when run locally and in the separate batches.

PR #1002, opened 2021-02-15
Fix index into previous signal handlers table

kstenerud:
## Goal & design

A previous PR incorrectly used `sigNum` as an index into the `bsg_g_previousSignalHandlers` table. This PR derives the correct index.

## Changeset

Added `get_previous_sigaction()`, which gets the correct index.

## Testing

Testing must be done manually. Before this fix, the "Mach exception" button in the example app will cause the app to freeze after recording the event. With this fix in place, it will correctly terminate the app after recording the event.

Tested on real devices running iOS 12.5.1 and 13.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_Signal.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_Signal.c
@@ -73,6 +73,17 @@ static BSG_KSCrash_SentryContext *bsg_g_context;
 #pragma mark - Callbacks -
 // ============================================================================
 
+static struct sigaction *get_previous_sigaction(int sigNum) {
+    const int *fatalSignals = bsg_kssignal_fatalSignals();
+    int fatalSignalsCount = bsg_kssignal_numFatalSignals();
+    for (int i = 0; i < fatalSignalsCount; i++) {
+        if(fatalSignals[i] == sigNum) {
+            return &bsg_g_previousSignalHandlers[i];
+        }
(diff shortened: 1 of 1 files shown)


Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_Signal.c
```diff
@@ -126,15 +137,19 @@ void bsg_kssighndl_i_handleSignal(int sigNum, siginfo_t *signalInfo,
 
     BSG_KSLOG_DEBUG(
         "Re-raising or chaining signal for regular handlers to catch.");
-    struct sigaction previous = bsg_g_previousSignalHandlers[sigNum];
-    if (previous.sa_flags & SA_SIGINFO) {
-        previous.sa_sigaction(sigNum, signalInfo, userContext);
-    } else if (previous.sa_handler == SIG_DFL) {
+    struct sigaction *previous = get_previous_sigaction(sigNum);
+    if(previous == NULL) {
+        BSG_KSLOG_ERROR("BUG: Could not find handler for signal %d", sigNum);
```
nickdowell, line 142: Why is it a bug if we couldn't find the previous handler?

kstenerud: We fill out the entire table when we start up, so this could only happen if we somehow got a signal number we weren't supposed to. Technically this should be impossible, but I'd rather put in a log msg :)

nickdowell (approved):

PR #1013, opened 2021-02-25
[PLAT-6113] Fix possible deadlock in bsg_ksmachfreeMemory

nickdowell:
## Goal

Fixes a deadlock that can occur because the implementation of `__builtin_available()` uses `malloc()` and is therefore not async-signal-safe.

This was observed when investigating potential causes of deadlocks in the crash handling code.

Screenshot 2021-02-22 at 16 38 27

## Design / Changeset

The availability of `os_proc_available_memory()` is now determined at initialization time rather than in the implementation of `bsg_ksmachfreeMemory()`

A function pointer is used to avoid the need for `#pragma clang diagnostic` statements in `bsg_ksmachfreeMemory()`

## Testing

Manually verified by running example apps and adding log statements to confirm that `os_proc_available_memory()` is used on the relevant platforms.

Added an E2E scenario that could consistently produce a deadlock in `bsg_ksmachfreeMemory()` prior to the fix, and confirmed that it passes with the fix in place.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMach.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMach.c
@@ -37,8 +37,9 @@
 #include <mach-o/arch.h>
 #include <mach/mach_time.h>
 #include <sys/sysctl.h>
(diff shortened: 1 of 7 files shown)


kattrali (approved):
nicely done 🎉

PR #1017, opened 2021-03-01
[PLAT-6079] Fix flake in testMainThreadFromBackground

nickdowell:
## Goal

After the [previous fix](https://github.com/bugsnag/bugsnag-cocoa/pull/1016) a further failure was observed on CI.

Some additional logging of the main thread's stack trace revealed that it was still running some GCD functions:

```
__workq_kernreturn
_dispatch_global_queue_poke_slow
-[BugsnagThreadTests testMainThreadFromBackground]
```

## Changeset

Added additional synchronization to ensure the stack trace is not captured before the main thread has returned from the call to `dispatch_async`

## Testing

Created a separate [testing branch](https://buildkite.com/bugsnag/bugsnag-cocoa/builds?branch=nickdowell/soak-testing) to run unit tests on all available platforms / versions, have seen no failures after implementing this fix.

### DIFF
--- a/Tests/BugsnagThreadTests.m
+++ b/Tests/BugsnagThreadTests.m
@@ -254,12 +254,16 @@
 - (void)testMainThreadFromBackground {
     NSParameterAssert(NSThread.currentThread.isMainThread);
     __block BugsnagThread *thread = nil;
-    __block atomic_bool done = false;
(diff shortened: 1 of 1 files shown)


kattrali (approved):
lol this is a fun one. took a minute for it to all click.

PR #1021, opened 2021-03-04
[PLAT-6132] Fix incorrect freeMemory for handled errors

nickdowell:
## Goal

Fixes a bug where handled errors (those reported via `notify()`) included an incorrect value for `freeMemory`

## Changeset

The root cause was `-[BSG_KSSystemInfo systemInfo]` not including `bsg_ksmachfreeMemory()` and `BugsnagDeviceWithState` incorrectly falling back to the "usable" value instead.

## Testing

Added E2E and unit test cases that replicate the problem, and verified they pass after the fix.

Manually verified that handled errors now contain `freeMemory < totalMemory`

### DIFF
--- a/Bugsnag/Payload/BugsnagDeviceWithState.m
+++ b/Bugsnag/Payload/BugsnagDeviceWithState.m
@@ -101,7 +101,7 @@ NSNumber *BSGDeviceFreeSpace(NSSearchPathDirectory directory) {
     BugsnagDeviceWithState *device = [BugsnagDeviceWithState new];
     [self populateFields:device dictionary:event];
     device.orientation = [event valueForKeyPath:@"user.state.deviceState.orientation"];
-    device.freeMemory = [event valueForKeyPath:@"system.memory.free"] ?: [event valueForKeyPath:@"system.memory.usable"];
+    device.freeMemory = [event valueForKeyPath:@"system.memory.free"];
     device.freeDisk = BSGDeviceFreeSpace(NSCachesDirectory);
 
     NSString *val = [event valueForKeyPath:@"report.timestamp"];
--- a/Bugsnag/Payload/BugsnagDeviceWithState.m
+++ b/Bugsnag/Payload/BugsnagDeviceWithState.m
@@ -97,7 +97,7 @@ NSNumber *BSGDeviceFreeSpace(NSSearchPathDirectory directory) {
     return device;
 }
 
-+ (BugsnagDeviceWithState *)deviceWithDictionary:(NSDictionary *)event {
++ (BugsnagDeviceWithState *)deviceWithKSCrashReport:(NSDictionary *)event {
     BugsnagDeviceWithState *device = [BugsnagDeviceWithState new];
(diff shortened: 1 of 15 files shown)


Bugsnag/Payload/BugsnagDeviceWithState.m
```diff
@@ -97,11 +97,11 @@ + (BugsnagDeviceWithState *)deviceWithOomData:(NSDictionary *)data {
     return device;
 }
 
-+ (BugsnagDeviceWithState *)deviceWithDictionary:(NSDictionary *)event {
++ (BugsnagDeviceWithState *)deviceWithKSCrashReport:(NSDictionary *)event {
     BugsnagDeviceWithState *device = [BugsnagDeviceWithState new];
     [self populateFields:device dictionary:event];
     device.orientation = [event valueForKeyPath:@"user.state.deviceState.orientation"];
-    device.freeMemory = [event valueForKeyPath:@"system.memory.free"] ?: [event valueForKeyPath:@"system.memory.usable"];
+    device.freeMemory = [event valueForKeyPath:@"system.memory.free"];
```
kstenerud, line 104: There are cases where "free" memory gets reported as 0. This was an attempt to have some value even if not totally correct.

kstenerud (changes requested):

Bugsnag/Payload/BugsnagDeviceWithState.m
nickdowell: I think it's better to omit freeMemory than include a misleading value

kstenerud: Fair enough, but we'll need to warn support about this change since it will be user facing...

nickdowell: I tracked down the [original source of this logic](https://github.com/bugsnag/bugsnag-cocoa/commit/8d132cb6#diff-fbaad0246273de8e162c006dcbc278624d7595cf6a99d950435b86f4f3464cda) and it appears to be a workaround for the fact that [`systemInfo.memory` did not contain `"free"`](https://github.com/bugsnag/bugsnag-cocoa/blob/8d132cb637e74b7821f63ae3dcc7f650dd367317/Source/KSCrash/Source/KSCrash/Recording/BSG_KSSystemInfo.m#L427-L431)

With the changes in this PR, `[event valueForKeyPath:@"system.memory.free"]` will always return a value - for unhandled / crashes this has always been the case since [`bsg_kscrw_i_writeMemoryInfo()` does include `"free"`](https://github.com/bugsnag/bugsnag-cocoa/blob/nickdowell/fix-free-memory/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c#L1178-L1194). Therefore there will not be any user facing changes here.

Note: the previous logic would not have prevented `freeMemory` being reported as `0` due to an NSNumber containing 0 not evaluating as false.

kattrali (approved):

PR #1024, opened 2021-03-08
[PLAT-6000] Fatal app hang reporting

nickdowell:
## Goal

Implements the persistence of the `app_hang.json` file and reporting of fatal app hangs (where the app terminates while hung.)

## Design & Changeset

* Added `-[BugsnagEvent initWithJson:]` to allow loading BugsnagEvents from JSON.
* `BugsnagClient` (in its AppHangs) category implements saving to, deletion of, and reading from `app_hang.json`
* Improved the accuracy of hang detections as I observed some undetected app hangs in E2E tests.
* Moved file scanning to separate operation queue to mitigate a timing issue that could cause an unwanted immediate retry of app hang upload upon failure in E2E tests.

## Testing

* Tested manually using sample app.
* Added unit test case to cover loading `BugsnagEvent` from JSON.
* Added E2E scenarios to cover fatal app hang detection, delivery, and retry handling.


Bugsnag/Helpers/BugsnagCollections.m
```diff
@@ -23,6 +23,17 @@
 
 #import "BSGJSONSerialization.h"
 
+NSArray * BSGArrayMap(NSArray *array, id (^ transform)(id)) {
+    NSMutableArray *mappedArray = [NSMutableArray array];
+    for (id object in array) {
+        id mapped = transform(object);
```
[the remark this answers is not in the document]
nickdowell: This lint issue is now resolved

Bugsnag/Client/BugsnagClient+AppHangs.m
```diff
@@ -62,4 +73,34 @@ - (void)appHangEnded {
     self.appHangEvent = nil;
 }
 
+- (BOOL)lastRunEndedWithAppHang {
```
kstenerud, line 76: Does this need to be idempotent in case it gets called more than once? (short-circuiting if `self.appHangEvent` != nil)

Bugsnag/Client/BugsnagClient.m
```diff
@@ -559,7 +563,12 @@ - (void)computeDidCrashLastLaunch {
             unlink(crashSentinelPath);
         }
     }
-    BOOL didCrash = handledCrashLastLaunch || didOOMLastLaunch;
+    
+    BOOL shouldReportOOM = NO;
+    BOOL didCrash = (handledCrashLastLaunch ||
+                     [self lastRunEndedWithAppHang] ||
+                     (shouldReportOOM = [self shouldReportOOM]));
```
kstenerud, line 570: This changes the behaviour of this section due to the assignment side-effect. If either `handledCrashLastLaunch` or `[self lastRunEndedWithAppHang]` is true, `shouldReportOOM = [self shouldReportOOM]` won't be evaluated.

kstenerud (changes requested):

Bugsnag/Client/BugsnagClient.m
nickdowell: That was intentional - we should not report an OOM if a crash or app hang was detected.

Maybe this needs needs more refactoring... I was trying to minimise the number of changes but the code could be made clearer.

Bugsnag/Client/BugsnagClient+AppHangs.m
nickdowell: Not strictly, since it shouldn't be called more than once (like `computeDidCrashLastLaunch`)

Bugsnag/Client/BugsnagClient.m
kstenerud: If it's just temporary and will go away within this batch of changes, I'm OK with it, but otherwise expressions with side effects are just bugs waiting to pounce once someone unfamiliar with this section (or forgets to look for it) attempts to make a change.

Bugsnag/Client/BugsnagClient+AppHangs.m
nickdowell: This has now been made idempotent 👍

Bugsnag/Client/BugsnagClient.m
nickdowell: This logic has been revised to (hopefully) make it more explicit and obvious.

kstenerud (approved):

PR #1030, opened 2021-03-10
[PLAT-6191] Add BugsnagStressTest (macOS)

nickdowell:
## Goal

The goal is to help detect deadlocks, concurrency related crashes, or memory usage issues in the use of `Bugsnag.notify()`

## Design

It is designed as a command line tool so that a successful exit will result in CI passing.

If the tool deadlocks, it will exceed the configured timeout and fail the build.

If the tool crashes, a non-zero exit code will be returned and the fail the build.

## Testing

Tested manually by introducing malloc / objc code in an area of code where threads had been suspended to verify that the tool deadlocked.


features/fixtures/macos-stress-test/BugsnagStressTest/main.m
```diff
@@ -0,0 +1,96 @@
...
+        for (int i = 0; i < kNumberOfIterations; i++) {
+            [notifyQueue addOperationWithBlock:^{
+                NSError *error = [NSError errorWithDomain:@"BugsnagStressTest" code:random() userInfo:nil];
+                [Bugsnag notifyError:error block:^BOOL(BugsnagEvent *event) {
+                    return YES;
+                }];
+            }];
+        }
+        
+        [notifyQueue addBarrierBlock:^{
```
kattrali, line 68: In the case that this test needs to run on macOS less than 10.15, we'll need some other syntax here since [`addBarrierBlock()` is 10.15+](https://developer.apple.com/documentation/foundation/nsoperationqueue/3172534-addbarrierblock).

kattrali (changes requested):

features/fixtures/macos-stress-test/BugsnagStressTest/main.m
nickdowell: Updated the implementation to be compatible back to macOS 10.11

kattrali (approved):

PR #1031, opened 2021-03-10
Improve reliability of OOM scenario

nickdowell:
## Goal

Improve the OOM scenario in the Objective-C iOS sample app so that it works on iPhone 5s.

## Changeset

Fixed iPhone 5s case by ignoring the size reported by `NSPageSize()` and instead dirty each 4096th byte.

The iPhone 5s also has a quirk where its memory limit is a lower percentage (65%) than other devices with similar amount of memory (70%) which has been addressed.

Improved the reliability of receiving a low memory warning after determining that it's important to _cross_ the memory limit boundary in a gradual way to trigger it.

Also updates the Swift iOS sample app to use the same method of provoking an OOM.

## Testing

Tested locally on iPhone 5s, iPhone 8, iPhone 12 mini and iPad 3rd generation.

### DIFF
--- a/examples/objective-c-ios/Bugsnag Test App/OutOfMemoryController.m
+++ b/examples/objective-c-ios/Bugsnag Test App/OutOfMemoryController.m
@@ -1,18 +1,16 @@
 #import "OutOfMemoryController.h"
 
-#define PRINT_STATS 0
+#include <mach/mach_init.h>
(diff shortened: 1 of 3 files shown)


twometresteve (commented):
No comments from my perspective.

kstenerud (approved):

PR #1034, opened 2021-03-15
Replace sprintf in JSON encoder

kstenerud:
## Goal

sprintf can call malloc, which can cause lockups in a crash scenario. Replace the sprintf calls in the JSON encoder with async-safe code. The new code is a simplified algorithm that always outputs in exponential form when the exponent is not 0. The result is the same effective value as before when ingested by a machine, but it may look a little strange to a human.

## Testing

* Added more unit tests specifically targeting the integer and float encoding in the JSON encoder.
* Re-ran unit and e2e tests.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSJSONCodec.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSJSONCodec.c
@@ -26,8 +26,9 @@
 
 #include "BSG_KSJSONCodec.h"
 
-#include <ctype.h>
-#include <stdlib.h>
+#include <math.h>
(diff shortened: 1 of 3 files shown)


nickdowell:
What effect does this have on the generated output? (if any)

There does not appear to be much test coverage of the floating point output - is it worth adding some test cases for the various possible inputs?

kstenerud:
I'm still working on the tests. This was supposed to be a draft.

nickdowell (approved):
Looks good to me.

Successfully ran `features/crashprobe.feature:159` 20 times on my Mac without a failure, whereas the `next` branch failed on the 5th attempt.

kattrali (approved):

PR #1035, opened 2021-03-15
Fix test flakes

nickdowell:
## Goal

Fixes some E2E test flakes that have been plaguing the `integration/app-hangs` branch.

## Changeset

* macOS test fixture will now accept class names that are not correctly capitalized - this fixes flakes caused by AppiumForMac not reliably pressing the shift key 🤦 
* Fixed flakes caused by the test fixture being terminated by the test framework before the scenario run method had executed
* Improved logging in the macOS test fixture, to aid diagnosis of these issues
* Rolled back iOS & tvOS 10 unit tests to run on macOS 10.14, because the iOS 10 simulator is not available on macOS 11

## Testing

Was able to reproduce the flakes locally, and verified fixes locally and on CI

### DIFF
--- a/.buildkite/pipeline.quick.yml
+++ b/.buildkite/pipeline.quick.yml
@@ -81,7 +81,7 @@ steps:
   - label: iOS 10 unit tests
     timeout_in_minutes: 10
     agents:
-      queue: opensource-mac-cocoa-11
+      queue: opensource-mac-cocoa-10.14
     env:
       LANG: "en_GB.UTF-8"
       TEST_CONFIGURATION: "Debug"
@@ -148,7 +148,7 @@ steps:
   - label: tvOS 10 unit tests
     timeout_in_minutes: 10
     agents:
-      queue: opensource-mac-cocoa-11
+      queue: opensource-mac-cocoa-10.14
     env:
       LANG: "en_GB.UTF-8"
       TEST_CONFIGURATION: "Debug"
--- a/features/fixtures/macos/macOSTestApp/main.m
+++ b/features/fixtures/macos/macOSTestApp/main.m
@@ -8,6 +8,11 @@
 
 #import <Cocoa/Cocoa.h>
 
+void sigterm(int signum) {
+    NSLog(@"Received SIGTERM");
+    exit(0);
+}
+
 int main(int argc, const char * argv[]) {
     [[NSUserDefaults standardUserDefaults] registerDefaults:@{
         // Disable state restoration to prevent the following dialog being shown after crashes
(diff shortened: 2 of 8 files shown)


features/fixtures/macos/macOSTestApp/main.m
```diff
@@ -8,6 +8,11 @@
 
 #import <Cocoa/Cocoa.h>
 
+void sigterm(int signum) {
+    NSLog(@"Recevied SIGTERM");
```
twometresteve, line 12: nit:
```suggestion
    NSLog(@"Received SIGTERM");
```

features/fixtures/shared/scenarios/Scenario.m
```diff
@@ -31,20 +31,31 @@ + (Scenario *)createScenarioNamed:(NSString *)className
...
+#elif TARGET_OS_OSX
+    NSString *swiftPrefix = @"macOSTestApp.";
 #endif
-    }
-
-    NSAssert(clz != nil, @"Failed to find class named '%@'", className);
 
-    BOOL implementsRun = method_getImplementation(class_getInstanceMethod([Scenario class], @selector(run))) !=
-    method_getImplementation(class_getInstanceMethod(clz, @selector(run)));
+    if (!clz) { // Case-intensitive class lookup because AppiumForMac is a bit unreliable at entering uppercase characters.
```
twometresteve, line 40:
```suggestion
    if (!clz) { // Case-insensitive class lookup because AppiumForMac is a bit unreliable at entering uppercase characters.
```

.buildkite/pipeline.quick.yml
```diff
@@ -81,7 +81,7 @@ steps:
   - label: iOS 10 unit tests
     timeout_in_minutes: 10
     agents:
-      queue: opensource-mac-cocoa-11
+      queue: opensource-mac-cocoa-10.14
```
twometresteve, line 84: Any particular need for 10.14 rather than 10.15?  We only have one 10.14 build server, but a few 10.15s.

twometresteve (approved):
Looks good, just a couple of nits and a question.

twometresteve:
Could it have been better to target this at `next` and merge up into the integration branch?

.buildkite/pipeline.quick.yml
nickdowell: I tried 10.15 - those boxes don't have the iOS 10 simulator installed so the tests fail.

twometresteve: OK, I'll get the simulators installed on them so we can bump the pipeline on at some point.

PR #1038, opened 2021-03-16
Test for malloc calls in more low-level crashes

kstenerud:
## Goal

Per PLAT-6097, hidden malloc calls can cause lockups in the reporter. This PR deliberately poisons malloc in certain low-level e2e tests such that it locks up the app if called from inside of the crash handler's async-safe code.

## Testing

Re-ran e2e tests.


nickdowell:
Looks good; will kick off a full CI run after office hours

nickdowell (approved):
Full E2E test run passed last night 🎉

PR #1048, opened 2021-03-22
[PLAT-6165] Bypass KSCrashReport for user-reported errors

nickdowell:
## Goal

Simplifies the reporting of user-reported (and React Native, Unity) errors by bypassing KSCrashReport storage.

## Changeset

- `-[BugsnagClient notifyInternal:block:]` now bypasses KSCrash and sends events directly to `BSGEventUploader`.
- Removed support for user-reported exceptions from KSCrash.
- Replaced the "overrides" mechanism used to attach Javascript stacktraces, to be compatible with storage as BugsnagEvent JSON.
- Implemented throttling to avoid send queue becoming too big and consuming too much memory.
- Updated E2E tests that trigger a crash immediately after calling `notify()`.

Unhandled errors are written immediately to disk, as this is essential for reliably reporting unhandled Javascript exceptions in React Native apps.

## Testing

- Manually tested with React Native
- Manually tested with sample apps
- Ran [full suite of E2E tests](https://buildkite.com/bugsnag/bugsnag-cocoa/builds/2347)
- Ran [React Native E2E tests against this branch](https://buildkite.com/bugsnag/bugsnag-js-react-native/builds/2005#e62f128e-4cd6-4930-a217-b9102d63aa38)

## Performance

The average time taken for `Bugsnag.notifyError()` has decreased from 6.5 to 4.9 milliseconds 🎉 (measured on an iPhone SE)


Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1261,7 +1246,7 @@ void bsg_kscrw_i_writeError(const BSG_KSCrashReportWriter *const writer,
     writer->beginObject(writer, key);
     {
 
-        if (BSG_KSCrashTypeUserReported != crash->crashType) {
+        if (crash->crashType) {
```
kattrali, line 1249: I suspect this condition is no longer needed at all?

features/support/env.rb
```diff
@@ -17,6 +17,10 @@
 
 # Additional require MacOS configuration
 if Maze.config.os == 'macos'
+  # The default macOS Crash Reporter "#{app_name} quit unexpectedly" alert grabs focus which can cause tests to flake.
+  # This option, which appears to have been introduced in macOS 10.11, displays a notification instead of the alert.
+  `defaults write com.apple.CrashReporter UseUNC 1`
```
kattrali, line 22: There's also `defaults write com.apple.CrashReporter DialogType none`, if that's clearer.

kattrali (approved):
looks good, added some nits.

hit submit too soon! Remaining question: what happens to kscrash-ish notify reports that are already on disk when developers upgrade?

nickdowell:
> what happens to kscrash-ish notify reports that are already on disk when developers upgrade?

They will be continue to be delivered; [-[BugsnagEvent initWithUserData:]](https://github.com/bugsnag/bugsnag-cocoa/blob/nickdowell/bypass-kscrashreport-user-reported/Bugsnag/Payload/BugsnagEvent.m#L354-L413) has been left untouched and will continue to convert kscrash-ish notify reports for delivery.

features/support/env.rb
nickdowell: I felt this option was a bit safer than disabling it totally.

It could be confusing to have it completely disabled after running the tests on ones laptop, and miss important crash reports thereafter.

kattrali: Ah, good point 👍

PR #1053, opened 2021-03-26
[PLAT-6261] Detect uncaught objc exceptions on Mac Catalyst and iOSAppOnMac

nickdowell:
## Goal

iOS apps running on Mac Catalyst or on Macs with Apple Silicon were not having uncaught exceptions reported when thrown from a UI event handler.

The normal mechanisms we use to detect exceptions (`NSUncaughtExceptionHandler` or `std::terminate_handler`) are not triggered due to the way AppKit has historically dealt with exceptions.

## Changeset

Bugsnag now hooks into `-[NSApplication reportException:]` to detect uncaught exceptions.

## Testing

Tested using a sample project targeting Mac Catalyst and on a Mac mini with Apple silicon.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_NSException.m
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_NSException.m
@@ -32,6 +32,8 @@
 //#define BSG_KSLogger_LocalLevel TRACE
 #import "BSG_KSLogger.h"
 
+#import <objc/runtime.h>
+
 // ============================================================================
 #pragma mark - Globals -
 // ============================================================================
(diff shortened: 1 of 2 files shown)


Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_NSException.m
```diff
@@ -130,9 +132,53 @@ void bsg_recordException(NSException *exception) {
...
 }
 
+// ============================================================================
+#pragma mark - iOS apps on macOS -
+// ============================================================================
+
+// iOS apps behave a little differently when running on macOS via Catalyst or
+// on Apple Silicon. Uncaught NSExceptions raised while handling UI events get
+// caught by AppKit and are not propagated to NSUncaughtExceptionHandler or
+// std::terminate_handler, therefore we need another way to detect them...
```
kattrali, line 147: I'd add the feedback number here as a reference for future travelers, both to note that its related to an ongoing issue and to make it easier to know when to remove this code if the issue is ever addressed. (maybe also in the changelog?)

kattrali (approved):

PR #1055, opened 2021-03-30
[ROAD-1067] Report macOS version and device model for iOS / Catalyst apps

nickdowell:
## Goal

Resolves PLAT-6100 and  PLAT-6246

Makes `device.model`, `osName` and `osVersion` reflect the Mac when running iOS apps via Catalyst or natively on Apple Silicon Macs.

## Changeset

Fetches the underlying OS version from `/System/Library/CoreServices/SystemVersion.plist` as APIs generally report the simulated iOS version. Had to bypass the usual APIs for reading this because Big Sur has a shim in `libsyscall` that redirects access to this file for iOS apps not built against Catalyst 🤯

Switches how `hw.machine` and `hw.model` are interpreted according to the underlying OS type so that the "Model Identifier" (e.g. `iPhone1,1` or `MacBookPro1,1`) ends up in the expected field.

## Testing

Tested manually using sample app on M1 Mac mini, Mac Catalyst, iOS devices running iOS 9 and 14, and tvOS.


Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSSystemInfo.m
```diff
@@ -63,6 +63,48 @@ static inline bool is_jailbroken() {
...
+/**
+ * Returns the content of /System/Library/CoreServices/SystemVersion.plist
+ * bypassing the open syscall shim that would normally redirect access to this
+ * file for iOS apps running on macOS.
+ *
+ * https://opensource.apple.com/source/xnu/xnu-7195.81.3/libsyscall/wrappers/system-version-compat.c.auto.html
+ */
+static NSDictionary * bsg_systemversion() {
+    static NSDictionary *systemVersion;
+    if (!systemVersion) {
```
kattrali, line 75: Should this block be wrapped in a `dispatch_once` block? presumably it only needs to be set once.

kattrali (approved):

PR #1059, opened 2021-04-01
Fix react native E2E test failure

nickdowell:
## Goal

Fix a React Native E2E test failure when reporting unhandled native promise rejections.

## Changeset

The symptom was the unhandled event not being sent.

The root cause was the use of [`performSelector:withObject:afterDelay:`](https://developer.apple.com/documentation/objectivec/nsobject/1416176-performselector) to upload the stored event(s). This API sets up a timer to perform the selector on the **current** thread. This is not guaranteed to fire since the user thread may terminate or not be running a runloop.

Fixed by using a global dispatch queue to perform the delayed upload.

## Testing

Verified fix on a testing branch - https://buildkite.com/bugsnag/bugsnag-js-react-native/builds/2039

Test failures for reference - https://buildkite.com/bugsnag/bugsnag-js-react-native/builds/2031


kattrali (approved):
This had to be a fun one to find, nice work.

PR #1060, opened 2021-04-02
Fix unused function warning

keith:
This function is only used in some compiler conditional branches, this
makes sure it doesn't warn in the other configurations.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSSystemInfo.m
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSSystemInfo.m
@@ -70,6 +70,7 @@ static inline bool is_jailbroken() {
  *
  * https://opensource.apple.com/source/xnu/xnu-7195.81.3/libsyscall/wrappers/system-version-compat.c.auto.html
  */
+#if !BSG_PLATFORM_SIMULATOR
 static NSDictionary * bsg_systemversion() {
     static NSDictionary *systemVersion;
     static dispatch_once_t onceToken;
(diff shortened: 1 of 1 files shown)


keith:
danger failure looks like credential issues

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSSystemInfo.m
```diff
@@ -70,7 +70,7 @@ static inline bool is_jailbroken() {
  *
  * https://opensource.apple.com/source/xnu/xnu-7195.81.3/libsyscall/wrappers/system-version-compat.c.auto.html
  */
-static NSDictionary * bsg_systemversion() {
+__attribute__((unused)) static NSDictionary * bsg_systemversion() {
```
nickdowell, line 73: Marking this function as unused is misleading.

Would be clearer to surround in `#if !TARGET_OS_SIMULATOR`

nickdowell (changes requested):

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSSystemInfo.m
keith: good point, updated!

nickdowell: Thanks @keith 😃

nickdowell (approved):

keith:
Thanks!

PR #1062, opened 2021-04-06
[PLAT-6298] Improve accuracy of KSCrashReport timestamp

nickdowell:
## Goal

KSCrashReport timestamps were accurate only to the nearest second.

This mean that some breadcrumbs could appear to occur after the error, e.g.

breadcrumbs

## Changeset

Add a new `BSG_KSCrashField_Timestamp_Millis` field to KSCrashReports containing a millisecond-accurate timestamp.

Based on the `gettimeofday()` API which although not documented to be async-signal safe, is [used by `times()`](https://opensource.apple.com/source/Libc/Libc-320.1.3/gen/FreeBSD/times.c.auto.html) which is...

## Testing

Was able to reliably detect the issue in E2E tests. Verified that they now pass.


nickdowell:
Ah it appears there's a flake for unhandled errors...

kattrali (approved):

PR #1065, opened 2021-04-07
[PLAT-6316] Fix SIGSEGV in bsg_ksmachgetThreadQueueName

nickdowell:
## Goal

Fixes a crash that can occur in `bsg_ksmachgetThreadQueueName`

## Changeset

The `thread_identifier_info.dispatch_qaddr` can sometimes contain an invalid address, which will cause a crash if dereferenced. This is something that has been noted [in the Crashlytics source code](https://github.com/firebase/firebase-ios-sdk/blob/Crashlytics-4.6.2/Crashlytics/Crashlytics/Components/FIRCLSProcess.c#L261-L270).

`bsg_ksmachcopyMem()` is now used to read the contents of the memory without causing a crash if invalid.

## Testing

I was not able to reproduce the crash, but manual testing and the E2E tests confirm that the function continues to return the expected queue name.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMach.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMach.c
@@ -307,19 +307,20 @@ bool bsg_ksmachgetThreadQueueName(const thread_t thread, char *const buffer,
     }
 
     thread_identifier_info_t idInfo = (thread_identifier_info_t)info;
-    dispatch_queue_t *dispatch_queue_ptr =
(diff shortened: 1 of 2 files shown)


kattrali (approved):
I don't entirely see how this would happen, at least based on the available [xnu sources](https://github.com/apple/darwin-xnu/blob/8f02f2a044b9bb1ad951987ef5bab20ec9486310/osfmk/kern/thread.c#L1806), where the addr is non-zero but also invalid, but a little extra safety is always welcome.

sethfri:
The timing of this is wonderful. I just noticed we've seen 1500 instances of this crash in production and was about to open an issue 😅

PR #1088, opened 2021-05-05
PLAT-6370 unity support

kstenerud:
## Goal

Expose `BSGBreadcrumbTypeValue` and `BSGBreadcrumbTypeFromString` via BugsnagBreadcrumb+Private.h

This is needed to support https://github.com/bugsnag/bugsnag-unity/pull/234

## Testing

Re-run unit and e2e tests

### DIFF
--- a/Bugsnag/Payload/BugsnagBreadcrumb+Private.h
+++ b/Bugsnag/Payload/BugsnagBreadcrumb+Private.h
@@ -20,4 +20,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+FOUNDATION_EXPORT NSString *BSGBreadcrumbTypeValue(BSGBreadcrumbType type);
+FOUNDATION_EXPORT BSGBreadcrumbType BSGBreadcrumbTypeFromString( NSString * _Nullable value);
+
 NS_ASSUME_NONNULL_END


nickdowell:
This has introduced a lint warning, and I'm not sure that the introduction of an "unknown" string is desirable.

Maybe the Unity layer could deal with the NSDictionary representation which can be accessed via `-[BugsnagBreadcrumb objectValue]` ?

kstenerud:
Technically "unknown" will never happen given our code, but I needed a non-null in all possibilities. The old code just did an implicit return of nil (which would also cause problems if that possibility occurred). That worked before because there was no declaration for the function, and so the compiler's nonnull checks were bypassed.

I could access it via the dict, but then we're back to losing all type checking and name binding, and we'd still have a nonnull being filled with a nullable.

kstenerud:
I'd rather avoid using dictionary lookups to get internal values and instead make use of the type checker and type bindings. Otherwise we have to manually keep everything in sync across multiple projects, and that's asking for trouble.

Bugsnag/Payload/BugsnagBreadcrumb+Private.h
```diff
@@ -20,4 +20,15 @@ NS_ASSUME_NONNULL_BEGIN
...
+extern "C" {
+#endif
+
+NSString *BSGBreadcrumbTypeValue(BSGBreadcrumbType type);
+BSGBreadcrumbType BSGBreadcrumbTypeFromString( NSString * _Nullable value);
+
+#ifdef __cplusplus
+}
+#endif
+
```
nickdowell, line 33: We could use the `FOUNDATION_EXPORT` macro to take care of the `extern "C"` dance thereby tidying up the code a bit:

```suggestion
FOUNDATION_EXPORT NSString *BSGBreadcrumbTypeValue(BSGBreadcrumbType type);
FOUNDATION_EXPORT BSGBreadcrumbType BSGBreadcrumbTypeFromString(NSString * _Nullable value);
```

nickdowell (approved):

PR #1089, opened 2021-05-07
[PLAT-6482] Internal error reporting

nickdowell:
## Goal

Enables errors within the notifier to be reported to Bugsnag for further investigation.

Implements sending an internal error if crash reports fail to parse correctly.

## Design

See engineering design document.

## Changeset

`BSGInternalErrorReporter` implements the formatting and sending of internal error event.

`BSGEventUploadKSCrashReportOperation` has been updated to send error events for various different parsing errors.

## Testing

Tested manually using a sample app and creating an invalid JSON file in the KSCrashReports directory. Reported errors look like this in the dashboard:

Screenshot 2021-05-07 at 08 58 37

Unit tests have been added to verify the BugsnagEvent and NSURLRequest objects are populated as expected.

Added E2E scenario to verify that an internal error is reported if an invalid KSCrashReport is encountered ([results](https://buildkite.com/bugsnag/bugsnag-cocoa/builds/2601#28b5e7e8-f901-46ee-bdb2-26f297ff00a2/366-370)).


Bugsnag/Helpers/BSGInternalErrorReporter.m
```diff
@@ -0,0 +1,177 @@
...
+    }
+}
+
+// MARK: Private API
+
+- (nullable BugsnagEvent *)eventWithErrorClass:(NSString *)errorClass
+                                       message:(nullable NSString *)message
+                                   diagnostics:(nullable NSDictionary<NSString *, id> *)diagnostics {
+    id<BSGInternalErrorReporterDataSource> dataSource = self.dataSource;
+    if (!dataSource) {
```
kstenerud, line 92: Should we even allow dataSource to be nil?

kstenerud (changes requested):

Bugsnag/Helpers/BSGInternalErrorReporter.m
nickdowell: It's a weak reference so cannot be guaranteed to be nonnull.

Bugsnag/Delivery/BSGEventUploadKSCrashReportOperation.m
```diff
@@ -22,17 +23,43 @@
 @implementation BSGEventUploadKSCrashReportOperation
 
 - (BugsnagEvent *)loadEventAndReturnError:(NSError * __autoreleasing *)errorPtr {
-    id json = [BSGJSONSerialization JSONObjectWithContentsOfFile:self.file options:0 error:errorPtr];
-    if (!json) {
+    NSError *error = nil;
+    
+    NSData *data = [NSData dataWithContentsOfFile:self.file options:0 error:&error];
+    if (!data) {
+        if (!(error.domain == NSCocoaErrorDomain && error.code == NSFileReadNoSuchFileError)) {
```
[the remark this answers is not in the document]
nickdowell: It's possible for the file to get deleted when enforcing `maxPersistedEvents` (if new files are created before this operation gets its turn on the queue) but this should be a rare event, so I'll remove the condition and we can reintroduce it if we get too much noise 👍

PR #1090, opened 2021-05-10
Attach appium server logs to MacOS test runs

Cawllec:
## Goal

To enable easier debugging, use the newest version of maze-runner to log the appium output to a file and attach it to the buildkite step.

Note: There appears to be an error/potential flake at the moment, and this PR will be updated once resolved.


twometresteve:
I would revert the MazeRunner upgrade.  There's a good chance of it impacting the iOS e2e tests and we'd want to perform a full build for it up-front, but it's not a problem we need to solve right now - I'd just raise a separate ticket for the upgrade.

PR #1091, opened 2021-05-11
fix: remove failure msg when disabling crash handling

kattrali:
## Goal

Remove a potentially concerning message indicating crash handling "failed" to be installed when its actually configured not to install anything.

### DIFF
--- a/Bugsnag/BugsnagCrashSentry.m
+++ b/Bugsnag/BugsnagCrashSentry.m
@@ -21,6 +21,9 @@
 
 - (void)install:(BugsnagConfiguration *)config onCrash:(BSGReportCallback)onCrash
 {
+    if (!config.autoDetectErrors) {
+      return;
+    }
     BSG_KSCrash *ksCrash = [BSG_KSCrash sharedInstance];
     ksCrash.introspectMemory = NO;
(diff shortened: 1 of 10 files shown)


Bugsnag/BugsnagCrashSentry.m
```diff
@@ -21,6 +21,9 @@ @implementation BugsnagCrashSentry
 
 - (void)install:(BugsnagConfiguration *)config onCrash:(BSGReportCallback)onCrash
 {
+    if (!config.autoDetectErrors) {
+      return;
```
nickdowell, line 25:
```suggestion
        return;
```

nickdowell (approved):

nickdowell:
@kattrali I have pushed some changes to
* make it easier to diagnose crashes in the stress test and prevent the scenario from timing out upon failure
* disable `autoDetectErrors` in the handled barebones test to detect any potential issues related to that
* fix the crash that was occurring when `autoDetectErrors` was disabled, while still silencing the error message
* remove the unused `handlingCrashTypes` property on `BSG_KSCrash`

kattrali:
Aha! It makes sense, thank you.

PR #1093, opened 2021-05-13
Add autoNotify API for internal use by bugsnag-unity

fractalwrench:
## Goal

Alters bugsnag-cocoa so that it is capable of supporting [Unity's AutoNotify](https://docs.bugsnag.com/platforms/unity/configuration-options/#autonotify) API, which allows automatic error detection to be enabled/disabled after Bugsnag has initialized. This is not exposed as part of the Cocoa interface and is intended for internal use by Unity only.

## Changeset

- Added `autoNotify` as a private property of `BugsnagClient`
- When `autoNotify` is updated `bsg_kscrash_setHandlingCrashTypes` is invoked with the appropriate error types to capture automatically

## Testing

Added E2E tests to verify that `autoNotify` is respected for NSException/Signals, and that error capture behaviour is unaltered if the flag is toggled off then on.

### DIFF
--- a/Bugsnag/Client/BugsnagClient+Private.h
+++ b/Bugsnag/Client/BugsnagClient+Private.h
@@ -35,6 +35,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nullable, nonatomic) BugsnagEvent *appHangEvent;
 
+/// Alters whether error detection should be enabled or not after Bugsnag has been initialized.
+/// Intended for internal use only by Unity.
+@property (nonatomic) BOOL autoNotify;
+
 @property (nullable, retain, nonatomic) BugsnagBreadcrumbs *breadcrumbs;
 
 @property (nullable, nonatomic) NSString *codeBundleId;
--- a/Bugsnag/Client/BugsnagClient.m
+++ b/Bugsnag/Client/BugsnagClient.m
@@ -1201,4 +1201,38 @@ NSString *_lastOrientation = nil;
     [self.state addMetadata:self.extraRuntimeInfo withKey:BSGKeyExtraRuntimeInfo toSection:BSGKeyDevice];
 }
 
+// =============================================================================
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## TBD
+
+### Bug fixes
+
+* Add autoNotify API for internal use by bugsnag-unity
```
nickdowell, line 8: I don't think we should add anything about this change in the changelog as it's internal API only.

Bugsnag/Client/BugsnagClient+Private.h
```diff
@@ -138,6 +138,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)start;
 
+/// Alters whether error detection should be enabled or not after Bugsnag has been initialized.
+/// Intended for internal use only by React Native/Unity.
+@property (nonatomic) BOOL autoNotify;
+
```
nickdowell: The properties are defined before the methods in this header, and alphabetically - please move this up to line 38

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
...
+  [#1093](https://github.com/bugsnag/bugsnag-cocoa/pull/1093)
+
```
nickdowell, line 10:
```suggestion
```

nickdowell (changes requested):
The new scenarios need to be added to the macOSTestApp Xcode project, too 👍

Bugsnag/Client/BugsnagClient+Private.h
fractalwrench: Moved property into alphabetical order.

fractalwrench:
Added scenarios to macOS test fixture and addressed review comments.

nickdowell (approved):

PR #1100, opened 2021-05-27
Split E2E runs into batches and randomize order

nickdowell:
## Goal

We have observed test flakes that appear to be Appium related occur around the 1hr mark during an E2E test run.

## Changeset

Split all E2E runs into batches to reduce the duration of each Appium / BS session. Note that batch 1 takes longer than batch 2 because there are a large number of scenarios in files beginning with `e` that prevent an even split.

Enabled `--order=random` to make it easier to differentiate between flakey scenarios and time-related issues.

Made the step naming more consistent and concise.

## Testing

For a sample full run see https://buildkite.com/bugsnag/bugsnag-cocoa/builds/2709

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -4,7 +4,7 @@ env:
 steps:
   ##############################################################################
   #
-  # Barebones E2E tests
+  # E2E tests
   #
 
   - label: 'iOS 13 E2E tests batch 1'
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -2,10 +2,15 @@ env:
   LANG: "en_GB.UTF-8"
 
 steps:
-  - label: ':ios: iOS 13 full end-to-end tests'
+  ##############################################################################
+  #
+  # Barebones E2E tests
+  #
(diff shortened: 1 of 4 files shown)


.buildkite/pipeline.full.yml
```diff
@@ -2,10 +2,42 @@ env:
   LANG: "en_GB.UTF-8"
 
 steps:
-  - label: ':ios: iOS 13 full end-to-end tests'
+  ##############################################################################
+  #
+  # Barebones E2E tests
```
twometresteve, line 7: Barebones?

twometresteve (commented):
Comment retracted.

nickdowell:
> > there are a large number of scenarios in files beginning with e that prevent an even split.
> 
> I don't understand - the split is on a-n and o-z. Can you just push m and n into batch 2? (I wouldn't get too hung up, though, I found they can be variable).

The split has been changed to to `a-g` and `h-z` but batch 1 is still the heaviest.

If we push `e` into batch 2 then we'll just reverse the problem...

twometresteve (approved):
LGTM - just one comment on the pipeline comment.

PR #1103, opened 2021-05-28
[PLAT-6632] Fix KSCrash state storage for apps with no CFBundleName

nickdowell:
## Goal

Some apps (e.g. macOS command line tools and background daemons) may not have an Info.plist or CFBundleName, and this was causing the KSCrash state persistence to fail, resulting in failing to detect that the last run ended with a crash.

## Changeset

When constructing the stateFilePath, the notifier now falls back to the process's name if it the app has no bundle name.

The bundle name is also no longer included in the KSCrashReport file names.

## Testing

Reproduced the reported issue using the macOS stress test app and verified the fix.

E2E tests ensure that crash reporting continues to work as expected.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrash.m
@@ -62,7 +62,6 @@
 
 @interface BSG_KSCrash ()
 
-@property(nonatomic, readwrite, retain) NSString *bundleName;
 @property(nonatomic, readwrite, retain) NSString *nextCrashID;
 
 // Mirrored from BSG_KSCrashAdvanced.h to provide ivars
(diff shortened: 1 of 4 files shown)


kattrali:
What does this change mean for people who upgrade and have existing data in the "{bundle name}" directory?

kstenerud (commented):
Will this handle migration from the old filenames to the new ones?

nickdowell:
The KSCrash state file path remains `KSCrashReports/${bundleName}-CrashState.json` for apps with a bundle name.

Apps without a bundle name were never able to save a state file due to the invalid generated path, so there's nothing to migrate. They will now use `KSCrashReports/${processName}-CrashState.json`

The other change contained here is to strip the bundle name from the name of each KSCrashReport file. This does not require migration because BSGEventUploader will [pick up any json file that does not end with -CrashState.json](https://github.com/bugsnag/bugsnag-cocoa/blob/nickdowell/fix-state-file-path/Bugsnag/Delivery/BSGEventUploader.m#L130-L139).

kstenerud (approved):

PR #1104, opened 2021-06-01
[PLAT-6628] Cache systemInfo values to improve performance

nickdowell:
## Goal

Reduce overhead of calling `Bugsnag.notify()`

## Changeset

The majority of values are now computed once and stored in a static NSDictionary.

Removed values that are not used anywhere.

## Testing

Tested via unit and E2E tests.


Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSSystemInfo.m
```diff
@@ -313,96 +301,19 @@ + (NSString *)currentCPUArch {
...
 // ============================================================================
 
-+ (NSDictionary *)systemInfo {
-    NSMutableDictionary *sysInfo = [NSMutableDictionary dictionary];
+/**
+ * Returns a systemInfo dictionary containing all the nonvolatile unchanging values.
+ */
++ (NSDictionary *)systemInfoStatic {
+    static NSMutableDictionary *sysInfo;
+    if (sysInfo) {
```
kattrali, line 313: Should this `dispatch_once` to avoid weirdness in the case that the first couple of invocations happen at about the same time on different threads?

nickdowell: I didn't think any weirdness was possible because the first invocation is predictable but have now wrapped it with `dispatch_once` anyway 👍

kattrali (approved):

PR #1108, opened 2021-06-03
[PLAT-6614] Stop session being reported if app is launched in the background

nickdowell:
## Goal

A session should not be reported if an app was started in the background (for example to perform a background fetch.)

Only once a user opens the app (and it moves to the foreground) should a session be recorded.

## Changeset

`BugsnagClient` now checks whether the app is in the foreground before starting a session.

`BugsnagSessionTracker` now starts a session as soon as the app moves into the foreground if there was no previous session.

## Testing

Manual testing only because it's not possible to trigger a background launch via Appium.

In one of the example Xcode projects, add the background fetch capability and then in the scheme's run options enable the Background Fetch option. Note: this only works when debugging an app.

E2E tests verify that automatic sessions reporting still works when an app is launched in the foreground.

### DIFF
--- a/Bugsnag/BugsnagSessionTracker.m
+++ b/Bugsnag/BugsnagSessionTracker.m
@@ -195,8 +195,8 @@ NSString *const BSGSessionUpdateNotification = @"BugsnagSessionChanged";
 }
 
 - (void)handleAppForegroundEvent {
-    if (self.backgroundStartTime
-        && [[NSDate date] timeIntervalSinceDate:self.backgroundStartTime] >= BSGNewSessionBackgroundDuration) {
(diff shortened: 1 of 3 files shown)


kattrali (approved):

sethfri:
Does this mean if the app crashes during a background fetch, we'll no longer get crash reports? I'm not sure that's desirable. It's still useful to know about app crashes, since iOS can stop calling you for background fetches if the app is misbehaving

nickdowell:
@sethfri this does not stop crashes being reported, but it does mean they will no longer count against the app's stability score.

sethfri:
@nickdowell Thanks for clarifying!

PR #1010, opened 2021-02-23
[PLAT-6075] Parse Swift 5.4 fatal error messages

nickdowell:
## Goal

In Swift 5.4 (shipping with iOS 14.5) the format of Swift fatal error messages is changing: https://github.com/apple/swift/commit/1a051719e3b1b7c37a856684dd037d482fef8e59

This PR updates the notifier's parsing code added in https://github.com/bugsnag/bugsnag-cocoa/pull/948 to respond to this change.

## Changeset

Added new regular expression patterns to cope with the change between versions.

## Testing

* Extended unit tests to cover Swift 5.4 format
* Verified manually with iOS 14.5 simulator

PR #1113, opened 2021-06-08
Adjust features/crashprobe.feature to cater for change to Swift error message format

nickdowell:
## Goal

In Swift 5.4 the error message format changed from
`Fatal error: Unexpectedly found nil while unwrapping an Optional value: file foo.swift line 10`
to
`foo.swift:10: Fatal error: Unexpectedly found nil while unwrapping an Optional value`

This PR updates `features/crashprobe.feature` so that it accepts either format.

Relates to #1010


twometresteve (approved):
LGTM

PR #1122, opened 2021-06-17
[PLAT-6800] App hang detection improvements

nickdowell:
## Goal

To prevent false positive detection of app hangs and improve performance.

### False positive mitigation

App hangs will now be ignored if the detector thread wakes up too long after the specified deadline. This condition could occur if an app was suspended while the detector thread was waiting on the (`processingFinished`) semaphore and subsequently woke up before the main thread was able to signal.

The app hang deadline is now advanced whenever `kCFRunLoopBeforeSources` is encountered. This caters for cases where a busy run loop runs multiple timers / sources iteration before sleeping.

### Performance improvement

The detector has been reworked to use a long-lived thread controlled by signalling two semaphores. This is more efficient - profiling revealed that `dispatch_after` is a roughly 7x slower than `dispatch_semaphore_signal`.

### Other improvements

App hangs are now symbolicated in-app (this was broken by the recent change to symbolicate asynchronously)

## Testing

Automated testing via [E2E tests running the app hang scenarios](https://buildkite.com/bugsnag/bugsnag-cocoa/builds?branch=nickdowell%2Fapp-hang-detection-improvements).

Updated the example apps to include a fatal app hang sample.

Tested manually using example apps.

### DIFF
--- a/Bugsnag/Helpers/BSGAppHangDetector.m
+++ b/Bugsnag/Helpers/BSGAppHangDetector.m
@@ -18,25 +18,21 @@
 #import "BugsnagLogger.h"
 #import "BugsnagThread+Private.h"
 
-#if TARGET_OS_IOS
-#import "BSGUIKit.h"
-#endif
-
 
 @interface BSGAppHangDetector ()
 
+@property (weak, nonatomic) id<BSGAppHangDetectorDelegate> delegate;
+@property (nonatomic) BOOL recordAllThreads;
 @property (nonatomic) CFRunLoopObserverRef observer;
+@property (nonatomic) dispatch_time_t processingDeadline;
+@property (nonatomic) dispatch_semaphore_t processingStarted;
+@property (nonatomic) dispatch_semaphore_t processingFinished;
+@property (weak, nonatomic) NSThread *thread;
 
 @end
 
 
 @implementation BSGAppHangDetector
-
-- (void)dealloc {
-    if (_observer) {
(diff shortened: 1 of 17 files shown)


Bugsnag/Helpers/BSGAppHangDetector.m
```diff
@@ -57,98 +53,147 @@ - (void)startWithDelegate:(id<BSGAppHangDetectorDelegate>)delegate {
...
-        // If we are already in the run loop (e.g. in app delegate) start monitoring immediately so that app hangs during app launch are detected.
-        observerBlock(self.observer, kCFRunLoopAfterWaiting);
-        CFRelease(runLoopMode);
-    }
+    CFRunLoopActivity activities = kCFRunLoopAfterWaiting | kCFRunLoopBeforeSources | kCFRunLoopBeforeWaiting;
+    self.observer = CFRunLoopObserverCreateWithHandler(NULL, activities, true, order, observerBlock);
+    
+    // Start monitoring immediately so that app hangs during launch can be detected.
+    // Note that because scene-based apps start in UIApplicationStateBackground, hangs in
+    // -[AppDelegate application:didFinishLaunchingWithOptions:] will be ignored.
```
kstenerud, line 100: Is there any way we can detect this in future?

nickdowell: We could temporarily allow background hangs to be acted upon, we'd need to find a suitable point at which to reenable the usual logic though.

Briefly discussed this with Tom and agreed it was better to address it as a separate issue to be discussed with the product team.

kstenerud (approved):

PR #1123, opened 2021-06-18
Fix unit test flake in -[BSG_KSMachHeadersTests testMainImage]

nickdowell:
## Goal

Fixes a unit test failure that occurs on iOS 15 Simulator / Xcode 13 beta

`bsg_mach_headers_get_main_image()` was sometimes returning `libMainThreadChecker.dylib` instead of `xctest`.

## Changeset

The root cause was `BSG_Mach_Header_Info.isMain` being left uninitialised for non-main images.

The `BSG_Mach_Header_Info` struct is now initialised to 0 by using `calloc` to fix this issue and prevent similar ones in the future.

## Testing

Was able to reproduce with Xcode 13's new "Run Repeatedly..." function and verify the fix.

GitHub Unit Tests action now passing ✅


kattrali (approved):
"Run Repeatedly" is very handy!

PR #1130, opened 2021-06-23
Fix flake in OOM scenario

nickdowell:
## Goal

To improve reliability of the OOM test scenario.

## Changeset

The test now waits for the app to stop running before relaunching the app.

## Testing

Tested using maze runner locally.


twometresteve (approved):
LGTM

Issue #1138, opened 2021-06-27
NSInvalidArgumentException BSGEventUploadOperation.m:112


tshedor:
### Describe the bug

After upgrading from 6.2 to 6.9, we've been seeing unhandled crashes from a narrow set of users with the following stacktrace:

```
NSInvalidArgumentException: -[NSNull countByEnumeratingWithState:objects:count:]: unrecognized selector sent to instance 0x20a654a00

0  CoreFoundation          ___exceptionPreprocess
1  libobjc.A.dylib         _objc_exception_throw
2  CoreFoundation          -[NSObject(NSObject) doesNotRecognizeSelector:]
3  CoreFoundation          ____forwarding___
4  CoreFoundation          ___forwarding_prep_0___
5  Runner                  +[BugsnagError errorFromJson:] (BugsnagError.m:119:9)
6  Runner                  -[BugsnagEvent stacktraceTypes] (BugsnagEvent.m:729:31)
7  Runner                  -[BSGEventUploadOperation runWithDelegate:completionHandler:] (BSGEventUploadOperation.m:112:67)
8  Runner                  -[BSGEventUploadOperation start] (BSGEventUploadOperation.m:177:5)
9  Foundation              ___NSOPERATIONQUEUE_IS_STARTING_AN_OPERATION__
10 Foundation              ___NSOQSchedule_f
11 libdispatch.dylib       __dispatch_block_async_invoke2
12 libdispatch.dylib       __dispatch_client_callout
13 libdispatch.dylib       __dispatch_continuation_pop
14 libdispatch.dylib       __dispatch_async_redirect_invoke
15 libdispatch.dylib       __dispatch_root_queue_drain
16 libdispatch.dylib       __dispatch_worker_thread2
17 libsystem_pthread.dylib __pthread_wqthread
18 libsystem_pthread.dylib _start_wqthread
```

This is strange because [the line in BugsnagError.m](https://github.com/bugsnag/bugsnag-cocoa/blob/master/Bugsnag/Payload/BugsnagError.m#L118-L119) checks for nil values:

```objective-c
    if (trace != nil) {
        for (NSDictionary *dict in trace) {
            BugsnagStackframe *frame = [BugsnagStackframe frameFromJson:dict];

            if (frame != nil) {
                [data addObject:frame];
            }
        }
    }
```

[...]

PR #1141, opened 2021-06-30
strictly compare for NSNull values #1138

tshedor:
## Goal

When attaching a custom stack trace, the stack trace types is sometimes an `NSNull` object. This causes an unhandled error detailed in #1138 

## Design

This strictly checks for `nil` and `NSNull` values instead of exclusively `nil` values. It's a bit of a hack. 

The unhandled errors are universally reported immediately after app start, which would indicate that custom stack trace errors that are stored before the next app launch (like in the event of a simultaneous OOM) are not stored within `BSGKeyStacktrace`. I'm a little lost in this code base so please forgive me not going deeper than ensuring `NSNull` values aren't iterated. 

## Changeset

The comparison for `trace` in `BugsnagError.m` changed from `!= nil` to `!(trace == nil || trace == NSNull)`.

## Testing

This bug has been difficult to reproduce, but the change is additive and so I believe it's a low risk solution with substantial gain (the gain being the Bugsnag error collector does not produce errors itself).

### DIFF
--- a/Bugsnag/Payload/BugsnagError.m
+++ b/Bugsnag/Payload/BugsnagError.m
@@ -114,8 +114,9 @@ NSString *BSGParseErrorMessage(NSDictionary *report, NSDictionary *error, NSStri
 + (BugsnagError *)errorFromJson:(NSDictionary *)json {
     NSArray *trace = json[BSGKeyStacktrace];
     NSMutableArray *data = [NSMutableArray new];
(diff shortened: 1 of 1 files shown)


nickdowell:
Hi @tshedor I'm closing this as we've addressed the issue with https://github.com/bugsnag/bugsnag-cocoa/pull/1143

Thanks for looking into it 👍

tshedor:
Thanks @nickdowell fo the quick response, I appreciate it

PR #1144, opened 2021-07-01
Retry launch_app to fix OOM flakes

nickdowell:
## Goal

To fix flakes in the OOM scenario.

## Changeset

The Appium logs revealed that the launch_app command sometimes fails when run after the fixture app has stopped running / been terminated.

This change will attempt a retry for the `I relaunch the app` step used by the OOM scenario.

## Testing

Was not able to reproduce the flake despite over 100 runs of the scenario on iOS 12.

### DIFF
--- a/features/steps/ios_steps.rb
+++ b/features/steps/ios_steps.rb
@@ -69,7 +69,12 @@ When('I relaunch the app') do
     system("killall #{app} > /dev/null && sleep 1")
     Maze.driver.get(app)
   else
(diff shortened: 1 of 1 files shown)


twometresteve (approved):
I suspect it will continue to flake, but this is nice and simple and worth a shot.  
We'll keep our ticket in play to investigate further - I'd look to put a mechanism into MR for retrying any action in Maze Runner a set number of times, waiting for a set time between.

PR #1146, opened 2021-07-06
Increase iOS E2E timeout to 60 minutes

nickdowell:
Increases the timeout on the iOS E2E tests because recently we have seen failures due to reaching the 45 minute limit.


twometresteve (approved):
LGTM and I think makes sense overall, but I'll also do some investigation to try and work out where the slowdown has come from.

PR #1150, opened 2021-07-07
Increase timeout for iOS 11 E2E tests

nickdowell:
## Goal

Bring iOS E2E timeout in line with other iOS versions (my search & replace missed this in #1146)


twometresteve (approved):
Ha - I even did a search for "45" too!

PR #1169, opened 2021-08-18
Localized rescue for crashy button clicks

twometresteve:
## Goal

Restrict the rescuing of errors on Appium button clicks to just those that occur for `run_scenario` (and on macOS).

## Design

Until now such errors were handled by Maze Runner for all button clicks, but this will be removed in [its next release](https://github.com/bugsnag/maze-runner/pull/285) as it is too generic.

## Testing

Covered by CI.

### DIFF
--- a/features/steps/ios_steps.rb
+++ b/features/steps/ios_steps.rb
@@ -3,8 +3,15 @@ When('I run {string}') do |event_type|
     Given the element "scenario_name" is present
     When I send the keys "#{event_type}" to the element "scenario_name"
     And I close the keyboard
-    And I click the element "run_scenario"
   )
+  begin
+    step('I click the element "run_scenario"')
+  rescue StandardError
+    # AppiumForMac raises an to run a scenario that crashes the app
+    raise unless Maze.driver.capabilities['platformName'].eql?('Mac')
+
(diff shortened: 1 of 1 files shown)


nickdowell:
💭 In theory only the `When("I run {string} and relaunch the app")` step should ever trigger crashes - so we might be able to restrict the rescue to that step and be more specific

twometresteve:
> 💭 In theory only the `When("I run {string} and relaunch the app")` step should ever trigger crashes - so we might be able to restrict the rescue to that step and be more specific

Yes, agreed - now reworked.

nickdowell (approved):

features/steps/ios_steps.rb
```diff
@@ -16,10 +16,15 @@
...
 
 When("I run {string} and relaunch the app") do |event_type|
-  steps %Q{
-    When I run "#{event_type}"
-    And I relaunch the app after a crash
-  }
+  begin
+    step("I run \"#{event_type}\"")
+  rescue StandardError
+    # AppiumForMac raises an to run a scenario that crashes the app
```
nickdowell, line 22: I'm finding it difficult to understand this comment 😅

twometresteve: Seems picky 😂 , but reworded.

PR #1173, opened 2021-08-25
Increase default max breadcrumbs

kstenerud:
## Goal

PLAT-7130: Increase default max breadcrumbs from 25 to 50.

## Testing

Re-ran all tests


twometresteve:
Should this be targeting the Network Breadcrumbs integration branch?

kstenerud:
Quite right! Targeting integration branch now.

nickdowell (approved):

PR #1181, opened 2021-09-07
Implement network breadcrumbs plugin logic

kstenerud:
## Goal

Implement logic and unit tests for ROAD-740.

## Design

Follows the design in ROAD-740.

## Testing

Added new unit tests to cover the requirements of ROAD-740.

### DIFF
--- a/BugsnagRequestMonitor/BugsnagRequestMonitor/BSGURLSessionTracingDelegate.m
+++ b/BugsnagRequestMonitor/BugsnagRequestMonitor/BSGURLSessionTracingDelegate.m
@@ -1,0 +1,100 @@
+//
+//  BSGURLSessionTracingDelegate.m
+//  
+//
+//  Created by Karl Stenerud on 07.09.21.
+//
+
+#import "BSGURLSessionTracingDelegate.h"
+#import <Bugsnag/Bugsnag.h>
+
+
+@implementation BSGURLSessionTracingDelegate
+
+// All proxies should be talking to the same sink.
+static id<BSGBreadcrumbSink> g_sink;
+
++ (BSGURLSessionTracingDelegate *_Nonnull)sharedDelegate {
+    static dispatch_once_t onceToken;
+    static BSGURLSessionTracingDelegate *delegate;
+    dispatch_once(&onceToken, ^{
+        delegate = [BSGURLSessionTracingDelegate new];
+    });
+
+    return delegate;
+}
+
++ (void)setSink:(nullable id<BSGBreadcrumbSink>) sink {
+    g_sink = sink;
+}
+
+- (BOOL)canTrace {
+    return g_sink != nil;
+}
+
+static NSString *responseConclusions[10] = {
+    @"NSURLSession error",
+    @"NSURLSession succeeded",
+    @"NSURLSession succeeded",
+    @"NSURLSession succeeded",
+    @"NSURLSession failed",
+    @"NSURLSession error",
+    @"NSURLSession error",
+    @"NSURLSession error",
+    @"NSURLSession error",
+    @"NSURLSession error",
+};
+
+- (NSString *)conclusionForResponseCode:(NSInteger)responseCode {
+    return responseConclusions[(responseCode / 100) % 10];
+}
+
+static NSDictionary *_Nonnull queryItemsAsDict(NSArray<NSURLQueryItem *> *_Nullable queryItems) {
+    NSMutableDictionary *result = [NSMutableDictionary new];
+    for(NSURLQueryItem *item in queryItems)
+    {
+        result[item.name] = item.value;
+    }
+    return result;
+}
+
+- (void)URLSession:(NSURLSession *)session
+              task:(NSURLSessionTask *)task
+didFinishCollectingMetrics:(NSURLSessionTaskMetrics *)metrics
+API_AVAILABLE(macosx(10.12), ios(10.0), watchos(3.0), tvos(10.0)) {
+    if (g_sink != nil) {
+        // Note: Cannot use metrics transaction request because it might have a 0 length HTTP body.
+        NSURLRequest *req = task.originalRequest;
+        NSURLComponents *urlComponents = [NSURLComponents componentsWithURL:req.URL resolvingAgainstBaseURL:YES];
+        // Note: Cannot use metrics transaction response because it will be nil if a custom NSURLProtocol is present.
+        // Note: If there was an error, task.response will be nil, and the following values will be set accordingly.
+        NSURLResponse *resp = task.response;
+        NSHTTPURLResponse *httpResp = [resp isKindOfClass:NSHTTPURLResponse.class] ? (NSHTTPURLResponse *)resp : nil;
+        NSString *message = [self conclusionForResponseCode:httpResp.statusCode];
+        int64_t requestContentLength = (int64_t)req.HTTPBody.length;
+        int64_t responseContentLength = resp.expectedContentLength;
+
+        if (@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)) {
+            NSURLSessionTaskTransactionMetrics *transaction = [metrics.transactionMetrics lastObject];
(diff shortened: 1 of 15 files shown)


BugsnagRequestMonitor/BugsnagRequestMonitor/BSGURLSessionTracingProxy.m
```diff
@@ -0,0 +1,66 @@
...
+
+
+@implementation BSGURLSessionTracingProxy
+
+#define METRICS_SELECTOR @selector(URLSession:task:didFinishCollectingMetrics:)
+
+- (instancetype)initWithDelegate:(nonnull id<NSURLSessionDelegate>)delegate
+                 tracingDelegate:(nonnull BSGURLSessionTracingDelegate *) tracingDelegate {
+    self.delegate = delegate;
+    self.tracingDelegate = tracingDelegate;
```
nickdowell, line 27: Property accessors should be avoided in `init` methods

```suggestion
    _delegate = delegate;
    _tracingDelegate = tracingDelegate;
```

BugsnagRequestMonitor/BugsnagRequestMonitor/NSURLSession+Tracing.m
```diff
@@ -0,0 +1,67 @@
...
+#import "NSURLSession+Tracing.h"
+#import "BSGURLSessionTracingDelegate.h"
+#import "BSGURLSessionTracingProxy.h"
+#import <objc/runtime.h>
+
+
+static IMP get_class_imp(Class _Nonnull class, SEL selector) {
+    Method method = class_getClassMethod(class, selector);
+    return method_getImplementation(method);
+}
```
nickdowell, line 17: This function is unnecessary because `method_setImplementation` returns the previous implementation.

BugsnagRequestMonitor/BugsnagRequestMonitor/NSURLSession+Tracing.m
```diff
@@ -0,0 +1,67 @@
...
+
+static void set_class_imp(Class _Nonnull class, SEL selector, id _Nonnull implementationBlock) {
+    Method method = class_getClassMethod(class, selector);
+    assert(method);
```
nickdowell, line 21: Do we really want to crash in the case the method is not found?

Tests/Swift Tests/BugsnagSwiftPublicAPITests.swift
```diff
@@ -257,7 +257,7 @@ class BugsnagSwiftPublicAPITests: XCTestCase {
         client.notifyError(err) { (event) -> Bool in return false }
      
         client.leaveBreadcrumb(withMessage: "msg")
-        client.leaveBreadcrumb("msg", metadata: [:], type: .manual)
+        client.leaveBreadcrumb(withMessage: "msg", metadata: [:], andType: .manual)
```
nickdowell, line 260: Why has this changed; have we made a breaking change to the API?

BugsnagRequestMonitor/BugsnagRequestMonitor/BSGURLSessionTracingDelegate.m
```diff
@@ -0,0 +1,87 @@
...
+        NSString *message = [self conclusionForResponseCode:httpResp.statusCode];
+
+        [g_sink leaveBreadcrumbWithMessage:message metadata:@{
+            @"status": @(httpResp.statusCode),
+            @"method": req.HTTPMethod,
+            @"url": urlComponents.string,
+            @"urlParams": queryItemsAsDict(urlComponents.queryItems),
+            @"duration": @(metrics.taskInterval.duration),
+            @"requestContentLength": @(req.HTTPBody.length),
+            @"responseContentLength": @(resp.expectedContentLength)
```
kattrali, line 82: This should use the metrics request content length where available, or else expected content length when its not. Most usage is from iOS 13+ anyway.

BugsnagRequestMonitor/BugsnagRequestMonitorTests/BugsnagRequestMonitorTests.m
```diff
@@ -5,15 +5,535 @@
...
++ (void)setNextStatusCode:(NSInteger)statusCode data:(NSData *)data {
+    mock_nextStatusCode = statusCode;
+    mock_nextData = data;
+}
+
+@end
+
+#pragma mark BugsnagRequestMonitorTests
+
+@interface BugsnagRequestMonitorTests : XCTestCase <BSGBreadcrumbSink>
```
nickdowell, line 53: There don't seem to be any tests cases that cover a non-zero reqLength?

BugsnagRequestMonitor/BugsnagRequestMonitor/NSURLSession+Tracing.m
```diff
@@ -0,0 +1,67 @@
...
+            // The shared session uses the shared NSURLCache, NSHTTPCookieStorage,
+            // and NSURLCredentialStorage objects, uses a shared custom networking
+            // protocol list (configured with registerClass: and unregisterClass:),
+            // and is based on a default configuration.
+            // https://developer.apple.com/documentation/foundation/nsurlsession/1409000-sharedsession
+            session = [NSURLSession sessionWithConfiguration:[NSURLSessionConfiguration defaultSessionConfiguration]
+                                                    delegate:nil delegateQueue:nil];
+        });
+
+        return session;
```
nickdowell, line 60: Note: this uses a different approach than some other APM frameworks, in that it creates an entirely new `sharedSession` rather than proxying the original / default one. Not aware of this being a problem, but thought it worth explicitly pointing out.

nickdowell (commented):
Mostly LGTM, just a few comments!

Tests/Swift Tests/BugsnagSwiftPublicAPITests.swift
kstenerud: Xcode refused to compile if I didn't change this.

`Incorrect argument labels in call (have '_:metadata:type:', expected 'withMessage:metadata:andType:')`

kattrali: Since the test passes on master, some change in this branch broke the public API, probably around making the sink public without `NS_SWIFT_NAME`, as we have here: https://github.com/bugsnag/bugsnag-cocoa/blob/b83847c74669f64a9afc0e7e7ade595b620c09e8/Bugsnag/include/Bugsnag/BugsnagClient.h#L115-L118

BugsnagRequestMonitor/BugsnagRequestMonitorTests/BugsnagRequestMonitorTests.m
kstenerud: There was, but it wasn't triggering because `transaction.request` seems to come with a zero-length body and the test hadn't had its length requirement updated. Test now requires the correct length, and intercept code now uses `task.originalRequest`.

https://github.com/bugsnag/bugsnag-cocoa/pull/1181/files#diff-ab26956c0f7f778219f804a6f357de4ebb1e1aa58890b84ac1721024b9b928fdR466

BugsnagRequestMonitor/BugsnagRequestMonitor/BSGURLSessionTracingDelegate.m
```diff
@@ -0,0 +1,100 @@
...
+            }
+            if (transaction.countOfResponseBodyBytesReceived != 0) {
+                responseContentLength = transaction.countOfResponseBodyBytesReceived;
+            }
+        }
+
+        [g_sink leaveBreadcrumbWithMessage:message metadata:@{
+            @"status": @(httpResp.statusCode),
+            @"method": req.HTTPMethod,
+            @"url": urlComponents.string,
```
nickdowell, line 91: This will crash if either `req` or `urlComponents` is nil.

nickdowell (approved):

PR #1188, opened 2021-09-16
Update example apps for network breadcrumbs

kstenerud:
## Goal

Updates the example apps to have (commented out) code for using the network breadcrumbs plugin

## Design

This PR modifies the podfiles and app delegates to add example code for network breadcrumbs. Only the Cocoapods projects have been modified at this time because Swift package manager must point to the master branch, and it would break CI to have the SPM example apps in a broken state in next or the integration branch (where the plugin code is not yet on the master branch).

## Testing

There are no new tests as this only updates the example apps.


examples/objective-c-ios/objective-c-ios/AppDelegate.m
```diff
@@ -59,6 +65,11 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 //    config.enabledErrorTypes.unhandledExceptions = YES;
 //    config.enabledErrorTypes.machExceptions = YES;
     
+    /**
+     * To enable network breadcrumbs, add the BugsnagRequestMonitor plugin to your config.
+     */
+//    [config addPlugin:[BugsnagRequestMonitor new]];
```
nickdowell, line 71: In general I recommend avoiding `+new`, and using `+alloc` and `-init`; that's what Apple do in their example / sample code.

nickdowell (approved):

PR #1193, opened 2021-09-22
Network breadcrumbs fixes

nickdowell:
## Goal

Fix the following issues:
* PLAT-7288 Omit urlParams from network breadcrumbs if no params in request
* PLAT-7320 Status and response length should be omitted on error
* PLAT-7321 Strip query params from URL
* PLAT-7322 Reported request length varies depending on status code

## Changeset

`urlParams` are now omitted (rather than including `{}`) if there are no query parameters.

If there are multiple query items with the same name, the values will now be combined into an array.

`status` and `responseContentLength` are now omitted if no response was received - i.e. in the case of a network error.

`requestContentLength` now represents the length of the request's body **excluding** protocol-specific framing, transfer encoding, and content encoding. The NSURLSessionTasks's `countOfBytesSent` is used when available, falling back to `HTTPBody.length` otherwise (which is necessary during unit testing with the custom NSURLProtocol.)

`responseContentLength` now represents the length of the **body** of the response (i.e. the data provided to the completion handler / delegate) rather than the length of the whole encoded request including headers, and is now fetched from the `NSURLSessionTask` object (which reports the correct data when an `NSURLProtocol` is in use, unlike `NSURLSessionTaskTransactionMetrics`.)

`url` no longer includes the query string.

`conclusionForResponseCode:` was renamed to `messageForResponse:` and the logic changed to ensure that status codes outside of the normal range would not produce unexpected results (e.g. the original logic would have returned "NSURLSession succeeded" with 1400 as its input.)

## Testing

Updated existing unit tests to very new behaviour.

Added test cases to verify stripping of query params from URL.

### DIFF
--- a/BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
+++ b/BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
@@ -88,7 +88,10 @@ API_AVAILABLE(macosx(10.12), ios(10.0), watchos(3.0), tvos(10.0)) {
         metadata[@"method"] = req.HTTPMethod;
         metadata[@"url"] = [BSGURLSessionTracingDelegate URLStringWithoutQueryForComponents:urlComponents];
         metadata[@"urlParams"] = [BSGURLSessionTracingDelegate urlParamsForQueryItems:urlComponents.queryItems];
-        if (req.HTTPBody) {
+        if (task.countOfBytesSent) {
+            metadata[@"requestContentLength"] = @(task.countOfBytesSent);
+        } else if (req.HTTPBody) {
+            // Fall back because task.countOfBytesSent is 0 when a custom NSURLProtocol is used
             metadata[@"requestContentLength"] = @(req.HTTPBody.length);
         }
         if (httpResp) {
--- a/BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
+++ b/BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
@@ -51,12 +51,20 @@ static id<BSGBreadcrumbSink> g_sink;
     }
     NSMutableDictionary *result = [NSMutableDictionary new];
     for (NSURLQueryItem *item in queryItems) {
+        // - note: If a NSURLQueryItem name-value pair is empty (i.e. the query string starts with '&', ends
+        // with '&', or has "&&" within it), you get a NSURLQueryItem with a zero-length name and a nil value.
+        // If a NSURLQueryItem name-value pair has nothing before the equals sign, you get a zero-length name.
+        // If a NSURLQueryItem name-value pair has nothing after the equals sign, you get a zero-length value.
+        // If a NSURLQueryItem name-value pair has no equals sign, the NSURLQueryItem name-value pair string
+        // is the name and you get a nil value.
+        id value = item.value ? item.value : [NSNull null];
+        
         if ([result[item.name] isKindOfClass:[NSMutableArray class]]) {
-            [result[item.name] addObject:item.value];
+            [result[item.name] addObject:value];
         } else if (result[item.name]) {
-            result[item.name] = [NSMutableArray arrayWithObjects:result[item.name], item.value, nil];
+            result[item.name] = [NSMutableArray arrayWithObjects:result[item.name], value, nil];
         } else {
-            result[item.name] = item.value;
+            result[item.name] = value;
         }
     }
     return result;
--- a/BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
+++ b/BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
@@ -45,13 +45,19 @@ static id<BSGBreadcrumbSink> g_sink;
     return @"NSURLSession error";
 }
 
-+ (nullable NSDictionary<NSString *, NSString *> *)urlParamsForQueryItems:(nullable NSArray<NSURLQueryItem *> *)queryItems {
++ (nullable NSDictionary<NSString *, id> *)urlParamsForQueryItems:(nullable NSArray<NSURLQueryItem *> *)queryItems {
     if (!queryItems) {
         return nil;
     }
     NSMutableDictionary *result = [NSMutableDictionary new];
     for (NSURLQueryItem *item in queryItems) {
-        result[item.name] = item.value;
+        if ([result[item.name] isKindOfClass:[NSMutableArray class]]) {
+            [result[item.name] addObject:item.value];
+        } else if (result[item.name]) {
+            result[item.name] = [NSMutableArray arrayWithObjects:result[item.name], item.value, nil];
+        } else {
+            result[item.name] = item.value;
+        }
     }
     return result;
 }
--- a/BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
+++ b/BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
@@ -6,6 +6,7 @@
 //
 
 #import "BSGURLSessionTracingDelegate.h"
+
 #import <Bugsnag/Bugsnag.h>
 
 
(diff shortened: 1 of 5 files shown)


BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
```diff
@@ -32,34 +33,40 @@ - (BOOL)canTrace {
...
-static NSDictionary *_Nonnull queryItemsAsDict(NSArray<NSURLQueryItem *> *_Nullable queryItems) {
++ (nullable NSDictionary<NSString *, NSString *> *)urlParamsForQueryItems:(nullable NSArray<NSURLQueryItem *> *)queryItems {
+    if (!queryItems) {
+        return nil;
+    }
     NSMutableDictionary *result = [NSMutableDictionary new];
-    for(NSURLQueryItem *item in queryItems)
-    {
+    for (NSURLQueryItem *item in queryItems) {
         result[item.name] = item.value;
```
kattrali, line 54: If a key already exists, convert the value into an array of values. Something like:

```suggestion
        if ([result[item.name] isKindOfClass:[NSArray class]]) {
            [result[item.name] addObject:item.value];
        } else if (result[item.name]) {
            result[item.name] = @[result[item.name], item.value].mutableCopy;
        } else {
            result[item.name] = item.value;
        }
```

though its not exactly that simple, since [value can be nil](https://developer.apple.com/documentation/foundation/nsurlcomponents/1407752-queryitems):
> If a name-value pair has no equals sign, the value property of the corresponding query item is nil.

BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
```diff
@@ -72,32 +79,24 @@ - (void)URLSession:(NSURLSession *)session
...
-        NSHTTPURLResponse *httpResp = [resp isKindOfClass:NSHTTPURLResponse.class] ? (NSHTTPURLResponse *)resp : nil;
-        NSString *message = [self conclusionForResponseCode:httpResp.statusCode];
-        int64_t requestContentLength = (int64_t)req.HTTPBody.length;
-        int64_t responseContentLength = resp.expectedContentLength;
-
-        if (@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)) {
-            NSURLSessionTaskTransactionMetrics *transaction = [metrics.transactionMetrics lastObject];
-            // Note: Must check for zero because these somtimes lie.
-            if (transaction.countOfRequestBodyBytesSent != 0) {
-                requestContentLength= transaction.countOfRequestBodyBytesSent;
```
kattrali, line 85: Is this different from `req.HTTPBody.length`? Seems odd with body in the name

BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
```diff
@@ -72,32 +79,24 @@ - (void)URLSession:(NSURLSession *)session
...
+
+        NSString *message = [BSGURLSessionTracingDelegate messageForResponse:httpResp];
+
+        NSMutableDictionary *metadata = [NSMutableDictionary dictionary];
+        metadata[@"duration"] = @((unsigned)(metrics.taskInterval.duration * 1000));
+        metadata[@"method"] = req.HTTPMethod;
+        metadata[@"url"] = [BSGURLSessionTracingDelegate URLStringWithoutQueryForComponents:urlComponents];
+        metadata[@"urlParams"] = [BSGURLSessionTracingDelegate urlParamsForQueryItems:urlComponents.queryItems];
+        if (req.HTTPBody) {
+            metadata[@"requestContentLength"] = @(req.HTTPBody.length);
```
kattrali, line 92: Why httpbody.length and not `task.countOfBytesSent`?

kattrali (commented):
Looks good, left one query around URL parameters related to earlier discussion.

BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
nickdowell: The documentation says

```
/*
 * countOfRequestBodyBytesSent is the number of bytes transferred for request body.
 * It includes protocol-specific framing, transfer encoding, and content encoding.
 */
@property (readonly) int64_t countOfRequestBodyBytesSent API_AVAILABLE(macos(10.15), ios(13.0), watchos(6.0), tvos(13.0));
```

so due to including encoding etc it's not the same as the size of the HTTPBody passed in by the app.

BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
nickdowell: `task.countOfBytesSent` returns erroneous values when an NSURLProtocol is in use.

nickdowell: `task.countOfBytesSent` is `0` when a custom NSURLProtocol is used, which breaks our unit tests.

nickdowell: But `HTTPBody.length` is broken because there isn't always a body - `HTTPBodyStream` may have been set instead.

BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
nickdowell: Name-value pairs with no equals sign are represented as an `NSURLQueryItem` with `item.value = @""`

BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
nickdowell: Now using `task.countOfBytesSent` when available

BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPluginTests/BSGURLSessionTracingDelegateTests.m
```diff
@@ -0,0 +1,68 @@
...
+    
+    TEST(@"http://example.com?foo=bar", @{@"foo": @"bar"});
+    
+    TEST(@"http://example.com?foo=bar&bar=baz", (@{@"foo": @"bar", @"bar": @"baz"}));
+    
+    // Multiple query items with the same name should be represented as arrays.
+    TEST(@"http://example.com?foo=bar&foo=baz", (@{@"foo": @[@"bar", @"baz"]}));
+    
+    // Query items with no value should be represented as empty string.
+    TEST(@"http://example.com?foo=bar&foo=baz&foo=&sort=name", (@{@"foo": @[@"bar", @"baz", @""], @"sort": @"name"}));
```
kattrali, line 34: This is missing a few cases.

1. without an equals sign, like

   ```objc
   TEST(@"http://example.com?bar", @{@"bar":@""});
   ```
   where the query item's value would be nil

2. unterminated `&`

   ```objc
   TEST(@"http://example.com?foo=bar&", @{@"foo": @"bar", @"":@""});
   ```

3. leading `=`

   ```objc
   TEST(@"http://example.com?=bar", @{@"": @"bar"});
   ```

nickdowell: Added these cases in https://github.com/bugsnag/bugsnag-cocoa/pull/1193/commits/5883a9e1d975430eaa90e7e6ec6ea2fb2608a0fc, though the results don't quite match your expectations 🙃

kattrali: I'm not sure I follow, previously you mentioned: 

> Name-value pairs with no equals sign are represented as an NSURLQueryItem with `item.value = @""`

Though if that was the case, [this test](
https://github.com/bugsnag/bugsnag-cocoa/commit/5883a9e1d975430eaa90e7e6ec6ea2fb2608a0fc#diff-6fad19de5bea41e743cf47d78cfe028632b8e00fe3a8f675e6819704ac1c16f7R36) would have a value?

```objc
TEST(@"http://example.com?foo", @{});
```

It looks like the value is nil, which is resulting in the key being "unset" [here](https://github.com/bugsnag/bugsnag-cocoa/pull/1193/files#diff-9184f3f6165422fee0871d7b405793441bee81d6be485fa7101129696675094bR59), so you can't tell the original query params contained key `foo`.

nickdowell: My mistake, I had not realized we would get an NSURLQueryItem with a nil value... should have read the documentation!

```C
// - note: If a NSURLQueryItem name-value pair is empty (i.e. the query string starts with '&', ends
// with '&', or has "&&" within it), you get a NSURLQueryItem with a zero-length name and a nil value.
// If a NSURLQueryItem name-value pair has nothing before the equals sign, you get a zero-length name.
// If a NSURLQueryItem name-value pair has nothing after the equals sign, you get a zero-length value.
// If a NSURLQueryItem name-value pair has no equals sign, the NSURLQueryItem name-value pair string
// is the name and you get a nil value.
```

nickdowell: Ok https://github.com/bugsnag/bugsnag-cocoa/pull/1193/commits/3833e06503de156e0b9846deeb9139385162370f should hopefully resolve this

kattrali (approved):

PR #1195, opened 2021-09-27
Add network breadcrumbs e2e tests

kstenerud:
## Goal

This adds an E2E test that checks for network breadcrumbs.

All commits from https://github.com/bugsnag/bugsnag-cocoa/pull/1184 have been rebased onto the current integration branch and merged into a single commit (separate commits would have been too messy due to the file renaming on the integration branch).

The metrics interception code now uses `NSURLSessionTaskTransactionMetrics` whenever it's available (see BSGURLSessionTracingDelegate.m).


nickdowell:
This PR appears to undo the fixes that were merged in last week

PR #1198, opened 2021-09-29
[PLAT-7263] Disable app hang detection in app extensions

nickdowell:
## Goal

App extensions have a different life cycle and environment that make the hang detection mechanism unsuitable.
* Depending on the type of extension, the run loop is not necessarily dedicated to UI.
* The host app or other extensions run by it may trigger false positives.
* The system may kill app extensions without any notification.

## Changeset

Disables app hang detection in app extensions.

Improves the logic of `+[BSG_KSSystemInfo isRunningInAppExtension]` by querying the Info.plist's [NSExtensionPointIdentifier](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/AppExtensionKeys.html#//apple_ref/doc/uid/TP40014212-SW15) rather than `CFBundlePackageType` (which would erroneously cause YES to be returned for some versions of Xcode's xctest unit test runner.)

## Testing

Manually verified in a Widget app extension.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## TBD
+
+### Bug fixes
+
+* Disable app hang detection for app extensions.
+  [#1198](https://github.com/bugsnag/bugsnag-cocoa/pull/1198)
+
 ## 6.13.0 (2021-09-29)
 
 ### Enhancements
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## TBD
+
+### Bug fixes
+
+* Disable app hang detection for app extensions.
+  [#2000](https://github.com/bugsnag/bugsnag-cocoa/pull/2000)
```
nickdowell, line 9:
```suggestion
  [#1198](https://github.com/bugsnag/bugsnag-cocoa/pull/1198)
```

kattrali (approved):

PR #1200, opened 2021-09-30
Add support for getting thread run state

kstenerud:
## Goal

PLAT-7316: Record thread run state in cocoa.

## Design

Per ROAD-1307

## Testing

* Updated unit tests and e2e tests to also check for thread run state.
* Added new unit test for run state to string conversion code and thread list manipulation.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashC.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashC.c
@@ -42,12 +42,14 @@
 #pragma mark - Globals -
 // ============================================================================
 
+/** True if BSG_KSCrash has been initialised. */
+static volatile sig_atomic_t bsg_g_initialised = 0;
+
 /** True if BSG_KSCrash has been installed. */
 static volatile sig_atomic_t bsg_g_installed = 0;
 
 /** Single, global crash context. */
-static BSG_KSCrash_Context bsg_g_crashReportContext = {
-    .config = {.handlingCrashTypes = BSG_KSCrashTypeProductionSafe}};
+static BSG_KSCrash_Context bsg_g_crashReportContext;
 
 /** Path to store the state file. */
 static char *bsg_g_stateFilePath;
@@ -55,6 +57,7 @@ static char *bsg_g_stateFilePath;
 // ============================================================================
 #pragma mark - Utility -
 // ============================================================================
+
 BSG_KSCrash_Context *crashContext(void) {
     return &bsg_g_crashReportContext;
 }
@@ -89,6 +92,13 @@ void bsg_kscrash_i_onCrash(BSG_KSCrash_Context *context) {
 // ============================================================================
 #pragma mark - API -
 // ============================================================================
+
+void bsg_kscrash_init(void) {
+    if (!bsg_g_initialised) {
+        bsg_g_initialised = true;
+        bsg_g_crashReportContext.config.handlingCrashTypes = BSG_KSCrashTypeProductionSafe;
+    }
+}
 
 BSG_KSCrashType bsg_kscrash_install(const char *const crashReportFilePath,
                                     const char *const recrashReportFilePath,
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashNames.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashNames.c
@@ -1,0 +1,28 @@
+//
+//  BSG_KSCrashNames.c
+//  Bugsnag
+//
+//  Created by Karl Stenerud on 28.09.21.
+//  Copyright © 2021 Bugsnag Inc. All rights reserved.
+//
+
+#include "BSG_KSCrashNames.h"
+#include <mach/thread_info.h>
+
+static const char* thread_state_names[] = {
+    // Defined in mach/thread_info.h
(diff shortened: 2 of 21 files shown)


Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -1063,10 +1070,11 @@ void bsg_kscrw_i_writeAllThreads(const BSG_KSCrashReportWriter *const writer,
     // Fetch info for all threads.
     writer->beginArray(writer, key);
     {
-        for (mach_msg_type_number_t i = 0; i < numThreads; i++) {
-            thread_t thread = threads[i];
```
nickdowell, line 1067: Now we're no longer using `threads` can we remove lines 1060 - 1070 and 1082 - 1088?

nickdowell:
> **`Bugsnag.framework`** binary size increased by 34,272 bytes from 1,156,600 to 1,190,872

That's quite a big increase!

I guess it's because `bsg_g_crashReportContext` is in the `__data` section due to having an [initializer](https://github.com/bugsnag/bugsnag-cocoa/blob/7316-thread-state/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashC.c#L49-L50).

If we removed the initializer it would be moved to the `__bss` section which AFAIK does not impact the binary size (the loader maps in zero-initialized pages for `__bss`).

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
kstenerud: I've updated the rest of the codebase to use the new getAllThreads and freeThreads functions. There were actually a couple of places where memory could leak before :)

kstenerud:
The context is now initialised via an init function, so the total size should go down a bit.

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashC.c
```diff
@@ -90,6 +93,14 @@ void bsg_kscrash_i_onCrash(BSG_KSCrash_Context *context) {
 #pragma mark - API -
 // ============================================================================
 
+void bsg_kscrash_init(void) {
+    if (!bsg_g_initialised) {
+        bsg_g_initialised = true;
+        memset(&bsg_g_crashReportContext, 0, sizeof(bsg_g_crashReportContext));
```
nickdowell, line 99: A call to `memset` is not required, `bsg_g_crashReportContext` is already initialized to 0 according to the C standard §6.7.8 10:

> If an object that has static storage duration is not initialized explicitly, then:
— if it has pointer type, it is initialized to a null pointer;
— if it has arithmetic type, it is initialized to (positive or unsigned) zero;
— if it is an aggregate, every member is initialized (recursively) according to these rules;
— if it is a union, the first named member is initialized (recursively) according to these rules.

kstenerud: Quite right!

Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry.h
```diff
@@ -101,6 +105,22 @@ typedef struct BSG_KSCrash_SentryContext {
...
 
+    /** All threads at the time of the crash.
+     * Note: This will be a kernel-allocated array that must be manually kernel-deallocated.
+     */
+    thread_t *allThreads;
+    unsigned allThreadsCount;
+
+    // NOTE: Due to a compiler bug, this must NOT be placed after threadsToResume.
+    //       The mach struct is safe to be placed afterwards.
+    unsigned threadsToResumeCount;
```
tomlongridge, line 116: Can you be a bit more specific on the version of iOS affected and a brief explanation of what you saw? Just thinking that this will worrying to our future selves to maintain/refactor if they don't know in what way it was breaking.

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashNames.c
```diff
@@ -0,0 +1,28 @@
...
+#include <mach/thread_info.h>
+
+static const char* thread_state_names[] = {
+    // Defined in mach/thread_info.h
+    NULL,
+    "running", // TH_STATE_RUNNING = 1
+    "stopped", // TH_STATE_STOPPED = 2
+    "waiting", // TH_STATE_WAITING = 3
+    "uninterruptible", // TH_STATE_UNINTERRUPTIBLE = 4
+    "halted", // TH_STATE_HALTED = 5
```
tomlongridge, line 19: Would it be better to use same strings as the consts in `mach/thread_info.h`? They still convey the same meaning but I think make it more clear where they came from.

Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry.h
kstenerud: Nick found the problem. Another test was initialising a global variable from a local stack-allocated one, and later tests were overwriting its data when they reached that point in the stack.

nickdowell (approved):

PR #1206, opened 2021-10-12
Group existing steps into new files

twometresteve:
## Goal

Reduce the number of lines of code in `ios_steps.rb`, which was getting a little large.

## Design

`app_steps.rb` groups steps associated with the test app.
`reusable_steps.rb` contains candidates for absorbing into Maze Runner itself.

## Changeset

No tangible overall change - just moving some step implementations into separate files without changing them.

## Testing

Covered by CI.


twometresteve:
> .

Sure - renamed to `cocoa_steps.rb`.

PR #1216, opened 2021-10-25
Add iOS 15 to CI

twometresteve:
## Goal

Add iOS 15 to CI.

## Testing

Covered by a full CI run.


Cawllec:
I've made a minor change to the appium version and verified the full test set runs on iOS 15, as well as rebased against `next`.  This should be ready for review.

nickdowell (approved):

PR #1227, opened 2021-11-15
[PLAT-7591] Feature flags

nickdowell:
## Goal

Implements feature flags - a formal way to express which optional features / experiments have been enabled in an app session.

## Design

Follows the Engineering Design.

## Changeset

* Adds the `BugsnagFeatureFlag` class and `BugsnagFeatureFlagStore` protocol.
* `BSGFeatureFlagStore.m` implements the basic manipulation and serialization of feature flags.
* `BugsnagClient` persists feature flags using the existing "state" metadata mechanism (also used to persist app, context and user info.)

## Testing

Covered by new unit tests and amended E2E scenarios.


Bugsnag/Payload/BugsnagEvent.m
```diff
@@ -680,6 +687,28 @@ - (void)setUnhandled:(BOOL)unhandled {
     self.handledState.unhandled = unhandled;
 }
 
+// MARK: - <BugsnagFeatureFlagStore>
+
+- (void)addFeatureFlagWithName:(NSString *)name variant:(nullable NSString *)variant {
+    BSGFeatureFlagStoreAddFeatureFlag(self.featureFlagStore, name, variant);
```
kstenerud, line 693: Why do these need to be C functions?

nickdowell: They don't need to be, but C functions take up less space in the binary.

Reimplementing these in a `BSGFeatureFlagStore` Objective-C class adds an additional 2kB.

kstenerud (approved):

kattrali (approved):

PR #1230, opened 2021-11-17
[PLAT-7635] check nil parameter in network breadcrumbs

kstenerud:
Apple metrics via the Apple API might not have a request object, or the request might not have a URL. We need to check before trying to use them.

Also made the request search more robust since there can be many possible places for them, and not all of them will have an HTTP body.

## Testing

Unfortunately we don't have a repro case for the report that triggered this issue, but the existing tests all pass.

### DIFF
--- a/BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
+++ b/BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
@@ -86,31 +86,36 @@ static id<BSGBreadcrumbSink> g_sink;
 - (void)URLSession:(__unused NSURLSession *)session task:(NSURLSessionTask *)task didFinishCollectingMetrics:(NSURLSessionTaskMetrics *)metrics
 API_AVAILABLE(macosx(10.12), ios(10.0), watchos(3.0), tvos(10.0)) {
     if (g_sink != nil) {
-        // Note: Cannot use metrics transaction request because it might have a 0 length HTTP body.
-        NSURLRequest *req = task.originalRequest;
-        NSURLComponents *urlComponents = [NSURLComponents componentsWithURL:req.URL resolvingAgainstBaseURL:YES];
-        // Note: Cannot use metrics transaction response because it will be nil if a custom NSURLProtocol is present.
-        // Note: If there was an error, task.response will be nil, and the following values will be set accordingly.
-        NSHTTPURLResponse *httpResp = [task.response isKindOfClass:NSHTTPURLResponse.class] ? (NSHTTPURLResponse *)task.response : nil;
-
-        NSString *message = [BSGURLSessionTracingDelegate messageForResponse:httpResp];
+        NSURLRequest *req = task.originalRequest ? task.originalRequest : task.currentRequest;
+        if (req == nil) {
+            return;
+        }
 
         NSMutableDictionary *metadata = [NSMutableDictionary dictionary];
         metadata[@"duration"] = @((unsigned)(metrics.taskInterval.duration * 1000));
         metadata[@"method"] = req.HTTPMethod;
-        metadata[@"url"] = [BSGURLSessionTracingDelegate URLStringWithoutQueryForComponents:urlComponents];
-        metadata[@"urlParams"] = [BSGURLSessionTracingDelegate urlParamsForQueryItems:urlComponents.queryItems];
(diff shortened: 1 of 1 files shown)


BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
```diff
@@ -83,12 +83,32 @@ + (nonnull NSString *)URLStringWithoutQueryForComponents:(nonnull NSURLComponent
...
+                return req;
+            }
+        }
+        // Failing that, just use the last phase.
+        req = transactionMetrics[transactionMetrics.count - 1].request;
+    }
+
+    return req;
+}
+
```
nickdowell, line 108: Why does this logic need to be so complicated?

If we have no way to reproduce a nil `originalRequest`, is it even worth having untested logic?

kstenerud: I wanted to try harder to get a valid request. The `transactionMetrics` array can contain a whole bunch of different objects depending on what phase of the request it was in (including retries from a 3xx response). If you think it's too risky I can just go back to the old way.

nickdowell: The NSURLSession header says:

```
@interface NSURLSessionTask : NSObject <NSCopying, NSProgressReporting>

@property (readonly)                 NSUInteger    taskIdentifier;    /* an identifier for this task, assigned by and unique to the owning session */
@property (nullable, readonly, copy) NSURLRequest  *originalRequest;  /* may be nil if this is a stream task */
@property (nullable, readonly, copy) NSURLRequest  *currentRequest;   /* may differ from originalRequest due to http server redirection */
@property (nullable, readonly, copy) NSURLResponse *response;         /* may be nil if no response has been received */
```

So we may be getting a nil `originalRequest` if it was a "stream task" - does that give a lead to find a way to reproduce the crash?

If so and this logic can't find a request for that case, probably better to keep it simple like:

```
        NSURLRequest *req = task.originalRequest ?: task.currentRequest;
        if (!req) {
            return;
        }
```

kstenerud: OK, I've put the simplified version in. I couldn't repro the crash using a stream task unfortunately.

BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin.xcodeproj/project.pbxproj
```diff
@@ -757,7 +757,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_STYLE = Automatic;
-				DEVELOPMENT_TEAM = 372ZUL2ZB7;
+				DEVELOPMENT_TEAM = "";
```
nickdowell, line 760: These changes should not be committed

BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
```diff
@@ -97,9 +94,20 @@ - (void)URLSession:(__unused NSURLSession *)session task:(NSURLSessionTask *)tas
...
         metadata[@"duration"] = @((unsigned)(metrics.taskInterval.duration * 1000));
-        metadata[@"method"] = req.HTTPMethod;
-        metadata[@"url"] = [BSGURLSessionTracingDelegate URLStringWithoutQueryForComponents:urlComponents];
-        metadata[@"urlParams"] = [BSGURLSessionTracingDelegate urlParamsForQueryItems:urlComponents.queryItems];
+
+        NSURLRequest *req = task.originalRequest;
+        if (req == nil) {
+            req = task.currentRequest;
+        }
+        if (req) {
```
nickdowell, line 102: If there's no request (and hence no url) is there any point leaving a breadcrumb at all?

@kattrali

kstenerud: Better some info than none?

nickdowell (approved):

PR #1232, opened 2021-11-18
Only upload the e2e test app once

twometresteve:
## Goal

Upload the e2e test fixture app to BrowserStack just once for each build, instead of in every e2e test step.  Uploads can occasionally fail, causing test flakes, so doing fewer uploads should reduce the chance of that.

## Design

The IPA build step now also uploads the built app to BrowserStack (using code adapted from Maze Runner).  It captures the `bs://` URL that's returned and writes it to a file that's then shared as a build artifact.  Whilst it's a little convoluted, it's the easiest way to pass information like this between build steps and, critically, into Maze Runner.

## Changeset

- New Ruby script to control the build and upload to BrowserStack
- IPA build artifact replaced with text file containing URL of uploaded app

## Testing

Covered by a full CI run.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ gem 'xcpretty'
 
 # A reference to Maze Runner is only needed for running tests locally and if committed it must be
 # portable for CI, e.g. a specific release.  However, leaving it commented out would mean quicker CI.
-gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v6.3.0'
+gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v6.5.0'
 
 # Use a development branch
 #gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', branch: 'tms/update-cucumber'
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ gem 'xcpretty'
 
 # A reference to Maze Runner is only needed for running tests locally and if committed it must be
 # portable for CI, e.g. a specific release.  However, leaving it commented out would mean quicker CI.
-gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v6.0.0'
+gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v6.3.0'
 
 # Use a development branch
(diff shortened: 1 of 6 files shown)


features/scripts/build_and_upload_app.rb
```diff
@@ -0,0 +1,35 @@
...
+  response['app_url']
+end
+
+# Build the test fixture apps
+`make test-fixtures`
+
+# Upload the IPA to BrowserStack
+url = upload_app ENV['BROWSER_STACK_USERNAME'], ENV['BROWSER_STACK_ACCESS_KEY'], 'features/fixtures/ios/output/iOSTestApp.ipa'
+puts "BrowserStack URL is #{url}"
+File.write('features/fixtures/ios/output/ipa_url.txt', url)
```
nickdowell, line 35: Wouldn't it be simpler to use `curl`?

Gemfile
```diff
@@ -7,7 +7,7 @@ gem 'xcpretty'
 
 # A reference to Maze Runner is only needed for running tests locally and if committed it must be
 # portable for CI, e.g. a specific release.  However, leaving it commented out would mean quicker CI.
-gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v6.0.0'
+gem 'bugsnag-maze-runner', git: 'https://github.com/bugsnag/maze-runner', tag: 'v6.3.0'
```
nickdowell, line 10: Why not use the latest release - [6.5.0](https://github.com/bugsnag/maze-runner/releases/tag/v6.5.0)?

twometresteve: No reason, the branch was just sat there for a little while and I didn't update it after the new releases came out.

features/scripts/build_and_upload_app.rb
twometresteve: Possibly, but we'd still need some script in there to process the response.  Ultimately, though, the repo should really have direct knowledge of BrowserStack at all - we should add an "upload app only" function to Maze Runner and abstract it away.  We could mark this section with a TODO against a ticket to refactor it?  Or, if it's too offensive, we can park this PR for a bit and add that function now?

twometresteve: Now annotated with a ticket number to remove almost all of it.

Gemfile
twometresteve: Now updated

nickdowell (approved):

PR #1234, opened 2021-11-22
Gather more useful information for breadcrumbs about windows

kstenerud:
## Goal

PLAT-7627

## Design

Capture window name, window class, and window description (when available) and add the information to breadcrumbs about windows.

This gives us information such as:

Screenshot 2021-11-22 at 11 51 40
Screenshot 2021-11-22 at 11 51 32

## Testing

Manual testing + reran existing tests.


kattrali (approved):
nice!

PR #1236, opened 2021-11-22
Build e2e test fixtures on ARM

twometresteve:
## Goal

Build e2e test fixtures on ARM and add e2e tests for macOS 12 (on ARM).

## Design

The tracks were already laid in terms on basic/full CI builds, so the changes here result from the addition of ARM-based build servers.

## Changeset

- iOS/macOS test fixtures now built on ARM with XCode 13
- Barebones and full e2e tests enabled for macOS 12/ARM
- (Not directly related fix) delete any test fixture app found in `/Applications` before copying the new one

## Testing

Covered by CI.

### DIFF
--- a/features/fixtures/macos/Podfile
+++ b/features/fixtures/macos/Podfile
@@ -19,11 +19,5 @@ post_install do |installer|
         end
       end
     end
-
-    if target.name == "Bugsnag" || target.name == "BugsnagNetworkRequestPlugin"
-      target.build_configurations.each do |config|
-        config.build_settings['ONLY_ACTIVE_ARCH'] = 'NO'
-      end
-    end
   end
 end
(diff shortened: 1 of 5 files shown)


features/fixtures/macos/Podfile
```diff
@@ -19,5 +19,11 @@ post_install do |installer|
         end
       end
     end
+
+    if target.name == "Bugsnag" || target.name == "BugsnagNetworkRequestPlugin"
+      target.build_configurations.each do |config|
+        config.build_settings['ONLY_ACTIVE_ARCH'] = 'NO'
+      end
+    end
```
nickdowell, line 27: This is not necessary - passing `ONLY_ACTIVE_ARCH=NO` to `xcodebuild` (in `features/scripts/export_mac_app.sh`) does the same thing (confirmed locally) and also doesn't slow things down when building & testing directly from Xcode.

```suggestion
```

nickdowell (approved):

PR #1237, opened 2021-11-23
[PLAT-7270] Capture KSCrash logs for E2E runs

nickdowell:
## Goal

Capture the KSCrash logs when running E2E tests, to help diagnose failures.

## Design

Uses Appium's [file pulling functionality](https://appium.io/docs/en/writing-running-appium/ios/ios-xctest-file-movement/) to fetch KSCrash logs, which are now written to disk.

## Testing

Sample failure with logs: https://buildkite.com/bugsnag/bugsnag-cocoa/builds/3949#2a097acf-c586-447a-a9be-5e86aab3aa87

Note that `pull_file` doesn't work on BrowserStack's iOS 10 devices.

### DIFF
--- a/features/fixtures/ios/iOSTestApp/Info.plist
+++ b/features/fixtures/ios/iOSTestApp/Info.plist
@@ -25,6 +25,8 @@
 		<key>NSAllowsArbitraryLoads</key>
 		<true/>
 	</dict>
+	<key>UIFileSharingEnabled</key>
+	<true/>
 	<key>UILaunchStoryboardName</key>
 	<string>LaunchScreen</string>
 	<key>UIMainStoryboardFile</key>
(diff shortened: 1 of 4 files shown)


features/fixtures/ios/iOSTestApp/Info.plist
```diff
@@ -48,19 +50,19 @@
 	</array>
 	<key>bugsnag</key>
 	<dict>
+		<key>apiKey</key>
+		<string>0192837465afbecd0192837465afbecd</string>
```
kstenerud, line 54: Do the e2e tests need an API key? I thought they were talking to a mock server?

twometresteve: It's a good question.... There are various steps that expect this specific apiKey in headers etc, so Nick is following the existing pattern.  It's just curious that an obviously fake apiKey wasn't chosen in the first place.

kstenerud: OK, as long as it's supposed to be there.

nickdowell: I haven't added anything here!

Xcode decided to re-order some of the lines in the Info.plist when a new field (`UIFileSharingEnabled`) was added.

kstenerud (approved):

features/support/env.rb
```diff
@@ -57,3 +57,21 @@ def skip_below(os, version)
...
+    folder1 = File.join(Dir.pwd, 'maze_output')
+    folder2 = scenario.failed? ? 'failed' : 'passed'
+    folder3 = scenario.name.gsub(/[:"& ]/, "_").gsub(/_+/, "_")
+
+    path = File.join(folder1, folder2, folder3)
+
+    FileUtils.makedirs(path)
+
+    File.open(File.join(path, 'kscrash.log'), 'wb') { |file| file << data }
+  end
```
twometresteve, line 74: We should probably provide an abstraction in Maze to allow clients to add their own log files, but this seems fine for now.

nickdowell: Ideally Maze would expose a way to get the path to store log files in. One to consider if any other clients need this functionality!

PR #1247, opened 2021-12-06
Use Maze Runner helper routine

twometresteve:
## Goal

Use Maze Runner helper routine for determining the current platform.

## Design

Use the `Maze::Helper.get_current_platform` routine added in Maze Runner v6.6.2 to avoid exposing the Cocoa test pipeline to differences between between different device farms in use.

## Testing

Covered by a `[quick ci]` run - any full set of e2e tests.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -207,7 +207,7 @@ GEM
     nokogiri (1.12.5)
       mini_portile2 (~> 2.6.1)
       racc (~> 1.4)
-    nokogiri (1.12.5-arm64-darwin)
+    nokogiri (1.12.5-x86_64-darwin)
       racc (~> 1.4)
     octokit (4.21.0)
       faraday (>= 0.9)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,9 +1,9 @@
 GIT
   remote: https://github.com/bugsnag/maze-runner
-  revision: fd33f19db2d0b03be857fecb55f1aa4d4033de6c
-  tag: v6.5.0
+  revision: 9efc4f0c876f050d34e0bef54be2599598076818
+  tag: v6.6.2
   specs:
-    bugsnag-maze-runner (6.5.0)
+    bugsnag-maze-runner (6.6.2)
       appium_lib (~> 11.2.0)
       bugsnag (~> 6.24)
       cucumber (~> 7.1)
@@ -41,7 +41,7 @@ GEM
       faye-websocket (~> 0.11.0)
       selenium-webdriver (~> 3.14, >= 3.14.1)
     atomos (0.1.3)
-    bugsnag (6.24.0)
+    bugsnag (6.24.1)
       concurrent-ruby (~> 1.0)
     builder (3.2.4)
     childprocess (3.0.0)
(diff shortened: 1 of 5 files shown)


Gemfile.lock
```diff
@@ -207,7 +207,7 @@ GEM
     nokogiri (1.12.5)
       mini_portile2 (~> 2.6.1)
       racc (~> 1.4)
-    nokogiri (1.12.5-x86_64-darwin)
+    nokogiri (1.12.5-arm64-darwin)
```
nickdowell, line 210: This prevents `bundle install` from working on Intel Macs (with built-in Ruby) 😬

Please undo!

features/steps/app_steps.rb
```diff
@@ -17,8 +17,8 @@
   # This step should only be used when the app has crashed, but the notifier needs a little
   # time to write the crash report before being forced to reopen.  From trials, 2s was not enough.
   sleep(5)
-  case Maze.driver.capabilities['platformName']
-  when 'Mac'
+  case Maze::Helper.get_current_platform
```
nickdowell, line 20: I'm no Ruby expert, but shouldn't this just be `Maze::Helper.current_platform` ?

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/AccessorMethodName

Gemfile.lock
twometresteve: Hmm, nasty.  I wonder how it decided to do that when I ran it on Intel myself - must not be a factor of the platform.

features/steps/app_steps.rb
```diff
@@ -3,8 +3,8 @@
 end
 
 When('I relaunch the app') do
-  case Maze.driver.capabilities['platformName']
-  when 'Mac'
+  case Maze::Helper.get_current_platform
+  when 'macos'
```
nickdowell, line 7: Would it be better if these were symbols (i.e. `:macos`) instead of strings?

twometresteve: Yes, almost certainly - I realised that too late, after releasing Maze Runner 😞 .  We could fix that in the future.

features/steps/app_steps.rb
twometresteve: Maybe, I wasn't sure - it's not really an accessor in the getter/setter sense, more of a "work_out_the_current_platform" routine, if that works sense.  The whole `Helper` class hasn't reached a coherent state as yet - it's more a holding area for reusable routines and a step-up from things living outside of any module, but it could certainly be evolved more in the future.  Once it's gather a certain amount of momentum it would make sense to refactor it into a more targeted set of classes.

Gemfile.lock
nickdowell: FWIW I see the same issue when running it on my machine - it seems like a bug in the version of Ruby preinstalled on macOS :-/

features/steps/cocoa_steps.rb
```diff
@@ -193,49 +193,53 @@ def check_device_model(field, list)
 end
 
 Then('the error is valid for the error reporting API') do
-  case Maze.driver.capabilities['platformName']
-  when 'iOS'
+  platform = Maze::Helper.get_current_platform
+  STDOUT.puts "platform: #{platform}"
```
nickdowell, line 197: Is this a debugging log statement that should be removed?

twometresteve: Yes, you're right - good spot.

nickdowell (approved):

PR #1254, opened 2021-12-08
Remove unnecessary automation steps

twometresteve:
## Goal

Remove the unnecessary closing of the keyboard and checking that elements are present before interacting with them.  This saves a total of an hour device time per full builds.

## Design

Each Appium `driver.find_element` operation takes about a second and closing the keyboard was taking two.  Losing the steps saves a lot of time and also makes the Appium logs much cleaner.

If we do start getting failures due to elements "missing" (not loaded yet), we can add a simple, reactive, retry.

## Changeset

Removal of unnecessary Cucumber steps.

## Testing

Covered by a full CI run.

### DIFF
--- a/features/steps/app_steps.rb
+++ b/features/steps/app_steps.rb
@@ -61,7 +61,6 @@ end
 
 When('I set the app to {string} mode') do |mode|
   steps %(
-    Given the element "scenario_metadata" is present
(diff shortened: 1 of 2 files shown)


nickdowell (approved):
🎉

PR #1259, opened 2021-12-14
[PLAT-5956] Ensure intended copy of Mac test fixture is run

nickdowell:
## Goal

Prevent the wrong app bundle being run during E2E testing.

## Changeset

The macOS test fixture now accepts a command line argument `-register` which will cause it to register itself with Launch Services, and (crucially) unregister all other copies of the app using the private `_LSUnregisterURL` API.

`env.rb` no longer moves the test fixture into `/Applications` - it will unzip if needed and register the app if found in the standard location of `{repo}/features/fixtures/macos/output`.

The fixture now prints its path in the log, to aid debugging in the future.

To run tests locally against a fixture built via Xcode, simply ensure that `features/fixtures/macos/output` is empty or missing.

## Testing

Verified locally and via CI.

### DIFF
--- a/features/fixtures/macos/macOSTestApp/main.m
+++ b/features/fixtures/macos/macOSTestApp/main.m
@@ -8,6 +8,8 @@
 
 #import <Cocoa/Cocoa.h>
 
+extern void _LSUnregisterURL(CFURLRef url);
+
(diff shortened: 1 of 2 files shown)


twometresteve (approved):
When I run this locally, the initial launch of the app stays up and I have to close it manually for the scenarios to start in anger.

However, everything seems to be ok on CI (I triggered a separate full run), so if you're happy then I'm happy.

PR #1263, opened 2021-12-20
[PLAT-6816] Detect hangs during launch of scene based apps

nickdowell:
## Goal

Detect hangs during launch of scene based apps.

UIScene-based apps are launched with a background UIApplicationState - the state changes to foreground when the first scene is created.

## Changeset

Enables detection of hangs between `Bugsnag.start()` and the first turn of the run loop regardless of the reported application state.

This means that hangs during the launch of background sessions will also be detected - no way to determine between foreground and background launches was found for scene based apps.

Once the run loop has run we expect the app state to be accurate.

## Testing

Tested locally using a sample app.

### DIFF
--- a/Bugsnag/Helpers/BSGAppHangDetector.m
+++ b/Bugsnag/Helpers/BSGAppHangDetector.m
@@ -22,6 +22,7 @@
 @interface BSGAppHangDetector ()
 
 @property (weak, nonatomic) id<BSGAppHangDetectorDelegate> delegate;
+@property (nonatomic) BOOL runLoopIsRunning;
 @property (nonatomic) BOOL recordAllThreads;
 @property (nonatomic) CFRunLoopObserverRef observer;
 @property (atomic) dispatch_time_t processingDeadline;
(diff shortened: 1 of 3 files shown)


Bugsnag/Helpers/BSGAppHangDetector.m
```diff
@@ -81,6 +82,9 @@ - (void)startWithDelegate:(id<BSGAppHangDetectorDelegate>)delegate {
                 // Each iteration indicates a separate unit of work so the hang detection should be reset accordingly.
                 dispatch_semaphore_signal(self.processingFinished);
             }
+            if (!self.runLoopIsRunning) {
+                self.runLoopIsRunning = YES;
+            }
```
kstenerud, line 87: Can't we just set `runLoopIsRunning` without the if statement?

kstenerud (approved):

Bugsnag/Helpers/BSGAppHangDetector.m
nickdowell: Fixed 👍

PR #1264, opened 2021-12-20
[PLAT-7694] Try to at least get the function address if the top of the stack is missing

kstenerud:
## Goal

In certain rare cases, the top of the stack can be lost when making a call to a bad function pointer (such as in the example app). Depending on the optimization and stripping flags, you'll only get the parent function of the function that made the bad function call (because the top of the stack will be that bad pointer).

This PR looks in other registers on 64-bit systems that contain the address of the beginning of the function that made the bad call. Then at least one can see the function it was called from, even if not the exact line.

## Design

Check if the first stack entry is valid, and if not, look in the secondary register (`r11` on x86-64, `x17` on arm64) for something that's within the main image. This isn't perfect, but chances are much higher that a rogue pointer will get called in the main app rather than a library, and we don't want to send people on a wild goose chase. If it's not in the main image, it doesn't get used and the behaviour is as before (the original pointer gets used, and rejected by the backend).

## Changeset

Keep track of the low and high addresses of the image list, and then on crash check the top of the stack with that range. If it's out of range, start looking in the secondary registers.

**Note** This only works on 64-bit systems. 32-bit systems behave as they did before.

## Testing

Tested manually with the example app. I couldn't get it to repro in the mazerunner app, probably due to optimization settings.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSBacktrace.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSBacktrace.c
@@ -27,6 +27,8 @@
 #include "BSG_KSBacktrace_Private.h"
 
 #include "BSG_KSMach.h"
+#include "BSG_KSMachHeaders.h"
+#include "BSG_KSLogger.h"
 
 /**
  * Mask to strip pointer authentication codes from pointers on Arm64e
@@ -104,6 +106,28 @@ bool bsg_ksbt_isBacktraceTooLong(
     return true;
 }
 
+static uintptr_t getFirstStackTraceAddress(const BSG_STRUCT_MCONTEXT_L *const machineContext) {
+    uintptr_t instructionAddress = bsg_ksmachinstructionAddress(machineContext);
+
+    if (!bsg_is_address_in_instruction_range(instructionAddress)) {
+        // These aren't the right addresses, but they do point to the beginning of
+        // the function that made the bad pointer call (which is better than nothing).
+#if defined (__x86_64__)
+        const uintptr_t tryAddress = machineContext->__ss.__r11;
+        if (bsg_is_address_in_instruction_range(tryAddress)) {
+            instructionAddress = tryAddress;
+        }
+#elif defined(__arm64__)
+        const uintptr_t tryAddress = machineContext->__ss.__x[17];
+        if (bsg_is_address_in_instruction_range(tryAddress)) {
+            instructionAddress = tryAddress;
+        }
+#endif
+    }
+
+    return instructionAddress;
+}
+
 int bsg_ksbt_backtraceThreadState(
     const BSG_STRUCT_MCONTEXT_L *const machineContext,
     uintptr_t *const backtraceBuffer, const int skipEntries,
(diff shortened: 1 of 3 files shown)


Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSBacktrace.c
```diff
@@ -104,6 +106,41 @@ bool bsg_ksbt_isBacktraceTooLong(
...
+    uintptr_t instructionAddress = bsg_ksmachinstructionAddress(machineContext);
+
+    if (!bsg_is_address_in_instruction_range(instructionAddress)) {
+        // These aren't the right addresses, but they do point to the beginning of
+        // the function that made the bad pointer call, which is better than nothing.
+        // We only do this if the address is in the main image, otherwise it could
+        // lead to a wild goose chase.
+#if defined (__x86_64__)
+        const uintptr_t tryAddress = machineContext->__ss.__r11;
+        if (isInMainImage(tryAddress)) {
```
nickdowell, line 130: Maybe this should be `bsg_is_address_in_instruction_range` instead?

Many (most?!) apps these days are comprised of multiple modules / frameworks, so only considering the main image seems too restrictive.

nickdowell:
For unoptimized builds this logic can result in confusing looking stack traces (e.g. below) so we need to consider whether we're happy for this to occur or if there is a way to avoid it

<details>

```
"exceptions": [
    {
        "errorClass": "EXC_BAD_ACCESS",
        "message": "Attempted to dereference garbage pointer 0xdeadbeef.",
        "stacktrace": [
            {
                "frameAddress": "0x100b6f414",
                "isPC": true,
                "machoFile": "/private/var/containers/Bundle/Application/847F943A-1C27-41A5-9CE9-37E2D0532726/iOSTestApp.app/iOSTestApp",
                "machoLoadAddress": "0x100b68000",
                "machoUUID": "B57A492F-BBE2-35C3-8703-D9A073A77712",
                "machoVMAddress": "0x100000000",
                "method": "-[BadFunctionPointerScenario run]",
                "symbolAddress": "0x100b6f414"
            },
            {
                "frameAddress": "0x100b6f43c",
                "isLR": true,
                "machoFile": "/private/var/containers/Bundle/Application/847F943A-1C27-41A5-9CE9-37E2D0532726/iOSTestApp.app/iOSTestApp",
                "machoLoadAddress": "0x100b68000",
                "machoUUID": "B57A492F-BBE2-35C3-8703-D9A073A77712",
                "machoVMAddress": "0x100000000",
                "method": "-[BadFunctionPointerScenario run]",
                "symbolAddress": "0x100b6f414"
            },
            {
                "frameAddress": "0x100b79804",
                "machoFile": "/private/var/containers/Bundle/Application/847F943A-1C27-41A5-9CE9-37E2D0532726/iOSTestApp.app/iOSTestApp",
                "machoLoadAddress": "0x100b68000",
                "machoUUID": "B57A492F-BBE2-35C3-8703-D9A073A77712",
                "machoVMAddress": "0x100000000",
                "method": "$s10iOSTestApp14ViewControllerC15runTestScenarioyyF",
                "symbolAddress": "0x100b79484"
            },
```

nickdowell:
Superseded by #1329

PR #1266, opened 2021-12-21
Use Maze.check abstraction

twometresteve:
## Goal

Use the `Maze.check` abstraction added to Maze Runner for the checking of values in tests.

## Design

This avoids potential issues from using functions in the global namespace and allows the use of forthcoming features of Maze Runner such as deferred checking.

## Changeset

Also includes a tweak to the pipeline steps to ensure that the latest Maze Runner docker image is always used.

## Testing

Covered by a full CI run.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -206,7 +206,7 @@ GEM
     nokogiri (1.12.5)
       mini_portile2 (~> 2.6.1)
       racc (~> 1.4)
-    nokogiri (1.12.5-arm64-darwin)
+    nokogiri (1.12.5-x86_64-darwin)
       racc (~> 1.4)
     octokit (4.21.0)
       faraday (>= 0.9)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,21 +1,20 @@
 GIT
   remote: https://github.com/bugsnag/maze-runner
-  revision: ad4225e8bcfb9765049c8abc7ae8833177249d70
-  tag: v6.7.0
+  revision: fe12189f83aad154f54221ee0fcd41b483d3c0d1
+  tag: v6.8.0
   specs:
-    bugsnag-maze-runner (6.7.0)
+    bugsnag-maze-runner (6.8.0)
       appium_lib (~> 11.2.0)
       bugsnag (~> 6.24)
       cucumber (~> 7.1)
       cucumber-expressions (~> 6.0.0)
       curb (~> 0.9.6)
-      minitest (~> 5.0)
       optimist (~> 3.0.1)
       os (~> 1.0.0)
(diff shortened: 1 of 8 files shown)


Gemfile.lock
```diff
@@ -207,7 +206,7 @@ GEM
     nokogiri (1.12.5)
       mini_portile2 (~> 2.6.1)
       racc (~> 1.4)
-    nokogiri (1.12.5-x86_64-darwin)
+    nokogiri (1.12.5-arm64-darwin)
```
nickdowell, line 209:
```suggestion
    nokogiri (1.12.5-x86_64-darwin)
```

features/steps/cocoa_steps.rb
```diff
@@ -1,6 +1,7 @@
 def short_scenario_name(scenario)
   unless scenario.end_with? 'Scenario'
-    raise 'All scenario names must end with "Scenario".  There are removed by the test harness and re-added uniformly by the test fixture.'
+    raise 'All scenario names must end with "Scenario".  ' \
+          'There are removed by the test harness and re-added uniformly by the test fixture.'
```
nickdowell, line 4: 🤨

Gemfile.lock
twometresteve: Argh, so annoying!  I have updated my version of Ruby to no effect. 😞

twometresteve: I wonder if there's is some way of pining it so it stays correct....

features/steps/cocoa_steps.rb
twometresteve: Yup, nasty typo that my brain is frequently guilty of making.  I do know the difference between they/there/their/they're really!

features/steps/cocoa_steps.rb
```diff
@@ -1,6 +1,7 @@
 def short_scenario_name(scenario)
   unless scenario.end_with? 'Scenario'
-    raise 'All scenario names must end with "Scenario".  There are removed by the test harness and re-added uniformly by the test fixture.'
+    raise 'All scenario names must end with "Scenario".  ' \
+          'They are removed by the test harness and re-added uniformly by the test fixture.'
```
nickdowell, line 4: To me this looks uglier / less readable when split onto two lines with `\`

Maybe it doesn't need to be so verbose?

Also there's a typo - "There are removed ..." instead of "These are removed ..."

twometresteve: I wrapped it in anticipation of Rubocop complaining about line lengths at some point in the future, but have now shortened the message and added an explanation as a comment - to try and avoid someone taking it out one day.

nickdowell (approved):

PR #1267, opened 2021-12-22
[PLAT-5164] Prevent curl hanging when uploading dSYMs

nickdowell:
## Goal

Stop `curl` from getting stuck using 100% CPU when uploading dSYMs.

## Changeset

Calls `Dir[\"#{ENV[\"DWARF_DSYM_FOLDER_PATH\"]}/*/Contents/Resources/DWARF/*\"]` before `fork`.

```diff
@@ -1,3 +1,5 @@
+dsyms = Dir["#{ENV["DWARF_DSYM_FOLDER_PATH"]}/*/Contents/Resources/DWARF/*"]
+
 fork do
   Process.setsid
   STDIN.reopen("/dev/null")
@@ -6,7 +8,7 @@
 
   require 'shellwords'
 
-  Dir["#{ENV["DWARF_DSYM_FOLDER_PATH"]}/*/Contents/Resources/DWARF/*"].each do |dsym|
+  dsyms.each do |dsym|
     system("curl -F dsym=@#{Shellwords.escape(dsym)} -F projectRoot=#{Shellwords.escape(ENV["PROJECT_DIR"])} https://upload.bugsnag.com/")
   end
 end
```

This was found to prevent the issue occurring - it's not clear exactly why, but spindump revealed that `curl` was repeatedly calling `select()` so presumably something about the open file descriptors was causing a problem.

## Testing

Reproduced the issue and verified the fix by repeatedly modifying, building, and running the iOS test fixture.


kattrali (approved):
that makes sense!

PR #1269, opened 2021-12-22
Combine User e2e test scenarios

twometresteve:
## Goal

Combine a number of the User e2e test scenarios, reducing the number of times that the app has to be relaunched and saving approx 40s on the overall test run time.

## Design

This is a relatively simple refactor to combine scenarios that are in logical groups.  E.g. one scenario now sends 4 different requests with different combinations of user info and another checks both a session and error in the same scenario.

## Changeset

I've also dropped the scenario `Only User ID field set` entirely.  At this stage it feels unnecessary to check every possible combination and the request where just the email is set should be sufficient to guard against any regression.

## Testing

`user.feature` run several times locally during development and covered by a `[quick ci]` run (which includes full test runs for both macOS and iOS).


features/fixtures/shared/scenarios/UserInfoScenario.swift
```diff
@@ -0,0 +1,47 @@
...
+    override func startBugsnag() {
+      self.config.autoTrackSessions = false;
+      super.startBugsnag()
+    }
+
+    override func run() {
+        let error = NSError(domain: "UserDefaultInfo", code: 100, userInfo: nil)
+        Bugsnag.notifyError(error)
+
+        after(.seconds(1)) {
```
fractalwrench, line 25: The scenario should try and avoid these waits if possible - it's going to lead to flaky behaviour if it takes more than 1s to call `Bugsnag.notify`. I'm not sure off the top of my head that it's required for bugsnag-cocoa - my guess would have been that the oldest error should always be delivered first (this could be wrong)?

features/user.feature
```diff
@@ -3,70 +3,62 @@ Feature: Reporting User Information
...
-  Scenario: Only User ID field set
-    When I run "UserIdScenario"
-    And I wait to receive an error
-    Then the error is valid for the error reporting API
-    And the exception "message" equals "The operation couldn’t be completed. (UserIdScenario error 100.)"
-    And the event "user.id" equals "abc"
-    And the event "user.email" is null
-    And the event "user.name" is null
-
-  Scenario: Overriding the user in the Event callback
```
fractalwrench, line 51: This scenario needs to be reinstated IMO - it confirms that the user information on `Event` can be overwritten.

features/fixtures/shared/scenarios/UserInfoScenario.swift
nickdowell: Agreed 👍 there's no need for these delays.

Errors will be delivered in the order of the calls to notify().

twometresteve: Ah, ok - I had it in my mind that calls to notify were asynchronous. Perhaps they are, but there's still a queuing mechanism that means they will be delivered in order (or I'm just wrong about them being asynch!).

nickdowell: No, calls to notify are synchronous.

features/user.feature
twometresteve: I rolled it into the session scenario that follows - I should change the name of to make it clear though.

features/fixtures/shared/scenarios/UserInfoScenario.swift
twometresteve: Delays removed

features/user.feature
twometresteve: Name now changed

fractalwrench (approved):
LGTM

nickdowell (approved):

PR #1274, opened 2022-01-10
Fix missing user.id in session-less OOMs

nickdowell:
## Goal

Make OOM events include user information even if there was no active session.

## Changeset

* Ensures the correct configuration object (that includes the auto-generated user id) is serialised to disk at start-up
* Uses the serialized user info for OOM events if there is no `session.user`

## Testing

Added an E2E scenario that triggered the problem and verified the fix.

### DIFF
--- a/features/out_of_memory.feature
+++ b/features/out_of_memory.feature
@@ -95,3 +95,10 @@ Feature: Out of memory errors
     Then the error is valid for the error reporting API
     And the event "unhandled" is false
     And the exception "message" equals "OOMEnabledErrorTypesScenario"
+
+  Scenario: Out of memory errors with no session contain a user id
+    Given I run "OOMSessionlessScenario" and relaunch the crashed app
+    And I configure Bugsnag for "OOMSessionlessScenario"
+    And I wait to receive an error
+    Then the error is an OOM event
+    And the event "user.id" is not null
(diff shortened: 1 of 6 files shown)


features/out_of_memory.feature
```diff
@@ -95,3 +95,10 @@ Feature: Out of memory errors
     Then the error is valid for the error reporting API
     And the event "unhandled" is false
     And the exception "message" equals "OOMEnabledErrorTypesScenario"
+
+  Scenario: Out of memory errors with no session contain a user id
+    Given I run "OOMSessionlessScenario" and relaunch the crashed app
+    And I configure Bugsnag for "OOMSessionlessScenario"
+    And I wait to receive an error
+    Then the error is an OOM event
+    And the event "user.id" is not null
```
kstenerud, line 104: If user was null, it will be replaced with an empty dict, so wouldn't `user.id` be null?

nickdowell: If user.id is nil, `BugsnagClient` will set it to the device hash

https://github.com/bugsnag/bugsnag-cocoa/blob/nickdowell/fix-oom-user-id/Bugsnag/Client/BugsnagClient.m#L207-L209

kstenerud (approved):

PR #1281, opened 2022-01-12
[PLAT-7804] Improve crash report writing performance with buffered output

nickdowell:
## Goal

Improve the performance of writing crash reports, to reduce the likelihood of apps being terminated (by watchdog or user) before finishing.

## Design

Use write buffers to reduce the number of calls to `write()`.

⚠️ Buffers are allocated on the stack since they cannot be `malloc()`'d at run-time due to the constrained crash-time environment. Thoughts on the safety of this are welcome - perhaps static allocation should be used instead at the expense of thread safety (which is not something we support in the crash handler.)

## Changeset

* Implements write buffering in `BSG_KSFile.c`.
* Modifies `BSG_KSCrashReport.c` to use buffered output.
* Modifies `BSG_KSCrashState.m` to use buffered output.

## Testing

Tested by updating unit tests `BSG_KSCrashReportTests` and `KSCrashState_Tests`.

Also verified by manually inspecting crash report files and running full E2E suite on CI. Bugs in file writing would manifest as unreported crashes.

The performance unit test case shows improvements of 40-60% depending on the device.

Profiling shows that around 95% of `bsg_kscrashreport_writeStandardReport()`'s time is now spent in `bsg_symbolicate()`.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSFile.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSFile.c
@@ -1,0 +1,59 @@
+//
+//  BSG_KSFile.c
+//  Bugsnag
+//
+//  Created by Nick Dowell on 12/01/2022.
+//  Copyright © 2022 Bugsnag Inc. All rights reserved.
+//
+
+#include "BSG_KSFile.h"
+
+#include "BSG_KSFileUtils.h"
+
+#include <string.h>
+#include <sys/param.h>
+
+static inline bool bsg_write(const int fd, const char *bytes, size_t length) {
+    return bsg_ksfuwriteBytesToFD(fd, bytes, (ssize_t)length);
+}
+
+void BSG_KSFileInit(BSG_KSFile *file, int fd, char *buffer, size_t length) {
+    file->fd = fd;
+    file->buffer = buffer;
+    file->bufferSize = length;
+    file->bufferUsed = 0;
+}
+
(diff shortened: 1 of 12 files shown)


Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSFile.c
```diff
@@ -0,0 +1,62 @@
...
+    }
+    
+    memcpy(file->buffer, data, length);
+    file->bufferUsed = length;
+    return true;
+}
+
+bool BSG_KSFileFlush(BSG_KSFile *file) {
+    if (!file->bufferUsed) {
+        return 0;
```
kstenerud, line 55: If the buffer is empty, flushing it returns false?

kstenerud (changes requested):
I think we need some isolated tests of BSG_KSFile to check all edge cases...

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSFile.c
nickdowell: oops that's supposed to be `true` - well spotted!

kstenerud:
Allocations don't have to all happen on the stack. In a crashed context there will only be one thread that can run, so a globally pre-allocated buffer will suffice. On a recrash, you just reinit the buffering mechanism.

nickdowell:
> I think we need some isolated tests of BSG_KSFile to check all edge cases...

Added a unit test that covers all the non-error code.

kstenerud (approved):

PR #1282, opened 2022-01-13
Use Maze Runner command endpoint

nickdowell:
## Goal

Speed up E2E tests.

## Design

Command payloads, that include the scenario name and action to take, are queued up by the test framework and read by the fixture using `GET /command`.

This is much faster than sending scenario names as text, which Appium enters character by character.

* https://github.com/bugsnag/maze-runner/pull/332

## Changeset

* iOS and macOS fixtures now have an execute_command button that fetches and executes a command.
* The scenario name is shown on the UI to provide some visual feedback before performing the action.
* More logic has been moved to Scenario.m where it can be shared between iOS and macOS fixtures.

## Testing

Ran full E2E suite locally and [on CI](https://buildkite.com/bugsnag/bugsnag-cocoa/builds/4653).

### DIFF
--- a/features/steps/cocoa_steps.rb
+++ b/features/steps/cocoa_steps.rb
@@ -1,17 +1,5 @@
-def short_scenario_name(scenario)
-  # A "Scenario" suffix is removed by the test harness and re-added uniformly by
-  # the test fixture. This reduces the time that Appium spends entering text.
-  unless scenario.end_with? 'Scenario'
-    raise 'All scenario names must end with "Scenario".'
-  end
-  scenario.delete_suffix 'Scenario'
-end
-
-When('I run {string}') do |event_type|
-  steps %(
-    When I set the app to "#{short_scenario_name event_type}" scenario
-    And I click the element "run_scenario"
-  )
+When('I run {string}') do |scenario_name|
+  execute_command :run_scenario, scenario_name
 end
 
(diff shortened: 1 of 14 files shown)


features/steps/cocoa_steps.rb
```diff
@@ -274,8 +259,13 @@ def wait_for_true
...
-  element.set_value(text)
+def execute_command(action, scenario_name)
+  command = { action: action, scenario_name: scenario_name, scenario_mode: $scenario_mode }
+  Maze::Server.commands.add command
+  Maze.driver.click_element :execute_command
+  $scenario_mode = nil
+  # Ensure fixture has read the command
+  count = 100
+  sleep 0.1 until Maze::Server.commands.remaining.empty? || (count -= 1) < 1
+  raise 'Test fixture did not GET /command' unless Maze::Server.commands.remaining.empty?
```
twometresteve, line 270: Did you have problems without this, or is it just belt and braces?

twometresteve (approved):
LGTM - just one question out of interest.

features/steps/cocoa_steps.rb
nickdowell: This is easier to diagnose than the more generic "Assertion timed out" we'd otherwise get

PR #1287, opened 2022-01-25
Improve concurrent KSLogging

nickdowell:
The KSCrash logs were invaluable in investigating and debugging the improvements to concurrent crash handling.

However, the logs were often confusing to read due to overlapping writes from different threads.

These changes improve log readability by ensuring each log message is written with a single call to `write()`, thereby reducing the chances for overlaps.

To fully ensure log file consistency would require some form of locking, this was avoided to avoid altering behaviour of code when logging is enabled.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSLogger.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSLogger.c
@@ -33,25 +33,14 @@
 #include <sys/fcntl.h>
 #include <unistd.h>
 
-// ===========================================================================
-#pragma mark - Common -
-// ===========================================================================
-
 // Compiler hints for "if" statements
 #define unlikely_if(x) if (__builtin_expect(x, 0))
 
 /** The buffer size to use when writing log entries.
  *
- * If this value is > 0, any log entries that expand beyond this length will
- * be truncated.
- * If this value = 0, the logging system will dynamically allocate memory
- * and never truncate. However, the log functions won't be async-safe.
- *
- * Unless you're logging from within signal handlers, it's safe to set it to 0.
+ * Any log entries that exceed this length will be truncated.
  */
-#ifndef BSG_KSLOGGER_CBufferSize
 #define BSG_KSLOGGER_CBufferSize 1024
-#endif
 
 /** Interpret the path as a unix file path and return the last path entry.
  * e.g. "/some/path/to/a/file.txt" will result in "file.txt"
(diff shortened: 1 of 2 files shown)


Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSLogger.c
```diff
@@ -147,88 +78,73 @@ bool bsg_kslog_setLogFilename(const char *filename, bool overwrite) {
...
+        return;
     }
 
-    FILE *file = fopen(filename, overwrite ? "wb" : "ab");
-    unlikely_if(file == NULL) {
-        writeFmtToLog("KSLogger: Could not open %s: %s", filename,
-                      strerror(errno));
-        return false;
+    unlikely_if(!fmt) {
+        len = (int)(strncpy(buf + len, "(null)", size - len) - buf);
```
kstenerud, line 132: `strncpy` returns its `dst` parameter, so int((`buf` + `len`) - `buf`) == `len`, which means the later `buf[len++] = '\n'` will replace the `(` in `(null)` with a newline.

I'd feel better if we used less clever code. It took me awhile to puzzle this out ;-)

kstenerud (changes requested):

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSLogger.c
```diff
@@ -147,88 +78,73 @@ bool bsg_kslog_setLogFilename(const char *filename, bool overwrite) {
...
+__printflike(1, 2)
+void bsg_i_kslog_logCBasic(const char *const fmt, ...) {
+    char buf[BSG_KSLOGGER_CBufferSize];
+    const int size = sizeof(buf);
+
+    int len = 0;
+
+    unlikely_if(!fmt) {
+        len = (int)(strncpy(buf + len, "(null)", size - len) - buf);
+    } else if (len < size - 1) {
```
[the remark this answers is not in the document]
nickdowell: True - it's because I've used the exact same code in the two logging functions.

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSLogger.c
nickdowell: Fair point - I'd feel better if we simply did an early exit if `fmt` is NULL to keep the code simpler - would you be ok with that?

kstenerud: Yup that's totally fine. Anything to keep code this low level as simple and non-crash-prone as possible :)

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSLogger.c
```diff
@@ -147,88 +77,81 @@ bool bsg_kslog_setLogFilename(const char *filename, bool overwrite) {
...
+    }
+    if (len > size) {
+        len = size;
+    }
+
+    if (len < size - 1) {
+        va_list args;
+        va_start(args, fmt);
+        int max = size - len;
+        int msglen = vsnprintf(buf + len, max, fmt, args);
```
kstenerud, line 143: If msglen < 0 there was an error and we should just return without writing.

kstenerud (changes requested):

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSLogger.c
nickdowell: Done 👍

kstenerud (approved):

PR #1292, opened 2022-02-08
Add bloaty reports to PRs

nickdowell:
## Goal

Use [Bloaty](https://github.com/google/bloaty) to show details of changes in framework size.

## Changeset

Includes Bloaty's diff output in the Danger message. More detailed bloaty output is available in the action logs, e.g.

Screenshot 2022-02-09 at 11 15 40

Runs Infer and OCLint separately because they need to be run on different git shas (bloaty needs to be run on the merge commit rather than the pr branch) and combining all steps results in a longer delay before getting results.

Stops using `bundle` because that ends up pulling in many unneeded dependencies and slowing down CI.

## Testing

Since this PR contains no changes to the framework, the output isn't very useful, so here is a sample obtained by running locally:

**`Bugsnag.framework`** binary size increased by 224 bytes from 911,040 to 911,264
```
    FILE SIZE        VM SIZE    
 --------------  -------------- 
  +0.1%    +160  +0.1%    +160    Symbol Table
  +0.0%     +88  +0.0%     +88    __TEXT,__text
  +0.0%     +56  +0.0%     +56    String Table
  +0.4%      +8  +0.4%      +8    Function Start Addresses
  +0.0%      +7  +0.0%      +7    __TEXT,__cstring
  +0.1%      +4  +0.1%      +4    __TEXT,__unwind_info
  -0.4%     -99  -0.4%     -99    [__TEXT]
  [ = ]       0  -3.5%    -224    [__LINKEDIT]
  +0.0%    +224  [ = ]       0    TOTAL
```


kstenerud (approved):

kattrali (approved):
neat!

PR #1293, opened 2022-02-11
Fix deadlock in testCrashReportWriterConcurrency under Rosetta

nickdowell:
## Goal

Fixes an intermitted hang / deadlock in `-[BugsnagBreadcrumbsTest testCrashReportWriterConcurrency]` that can occur when running in an Intel-based Simulator on ARM Macs.

## Analysis

Reproduced the hang on an ARM Mac and captured [spindump.txt](https://github.com/bugsnag/bugsnag-cocoa/files/8049007/spindump.txt)

It shows some interesting threads like `com.apple.rosetta.exceptionserver` and `com.apple.rosetta.debugserver` that could cause trouble if suspended, I suspect the deadlock has something to do with these.

## Changeset

Modifies the test to only suspend the test's own threads.

## Testing

Re-ran the test on the ARM Mac (1000 repetitions, relaunching each time) and could no longer reproduce the hang 🎉

### DIFF
--- a/Tests/BugsnagTests/BugsnagBreadcrumbsTest.m
+++ b/Tests/BugsnagTests/BugsnagBreadcrumbsTest.m
@@ -7,7 +7,6 @@
 //
 
 #import "BSGUtils.h"
-#import "BSG_KSCrashSentry_Private.h"
 #import "BSG_KSJSONCodec.h"
 #import "Bugsnag.h"
 #import "BugsnagBreadcrumb+Private.h"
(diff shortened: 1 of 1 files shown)


kstenerud (approved):

kattrali (approved):

1964fayz (commented):
الجمود

1964fayz:
شكرا

في الجمعة، ١١ فبراير ٢٠٢٢ ٧:١١ م Nick Dowell ***@***.***> كتب:

PR #1295, opened 2022-02-14
Ignore non-fatal SIGPIPEs

nickdowell:
## Goal

Stop reporting false `SIGPIPE` unhandled errors when ignored with `SIG_IGN`.

If an app uses `sigignore(SIGPIPE)` or `signal(SIGPIPE, SIG_IGN)` then when a `SIGPIPE` is raised, the process will not be terminated.

Bugsnag's hander assumes that the signal will be fatal and writes a crash report.

## Changeset

Removes Bugsnag's signal handler for `SIGPIPE` if the previously configured handler was `SIG_IGN`.

## Testing

Adds an E2E scenario that was able to reproduce the problem and verify the fix.

Updates SIGPIPEScenario to use a more realistic method to trigger a SIGPIPE.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_Signal.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_Signal.c
@@ -212,6 +212,11 @@ bool bsg_kscrashsentry_installSignalHandler(
             }
             goto failed;
         }
+        if (fatalSignals[i] == SIGPIPE &&
+            bsg_g_previousSignalHandlers[i].sa_handler == SIG_IGN) {
+            BSG_KSLOG_DEBUG("Removing handler for signal %d", fatalSignals[i]);
+            sigaction(fatalSignals[i], &bsg_g_previousSignalHandlers[i], NULL);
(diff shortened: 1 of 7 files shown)


Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_Signal.c
```diff
@@ -212,6 +212,11 @@ bool bsg_kscrashsentry_installSignalHandler(
             }
             goto failed;
         }
+        if (fatalSignals[i] == SIGPIPE &&
```
kstenerud, line 215: I wonder if maybe we should check for SIG_IGN on all signals except KILL and STOP (which can't be ignored).

kstenerud (approved):

kattrali (approved):

Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_Signal.c
nickdowell: Yes that might be worth a follow-up 👍

PR #1297, opened 2022-02-16
Allow provisioning updates when building examples

twometresteve:
## Goal

Resolve build errors on newly setup servers.

## Testing

Tested on two specific servers prior to raising a PR - all examples built when they were persistently failing previously.  Also covered by a full CI run.


joshedney (approved):
LGTM!

PR #1300, opened 2022-02-17
Testing/macos remove appium

Cawllec:
## Goal

Remove the requirement for Appium and AppiumForMac when running MacOS E2E tests.

## Issues

Ensure that all tests currently failing are properly investigated and fixed


features/fixtures/macos/macOSTestApp/MainWindowController.m
```diff
@@ -34,8 +36,13 @@ - (void)windowDidLoad {
     [super windowDidLoad];
 
     self.apiKey = @"12312312312312312312312312312312";
-    self.notifyEndpoint = @"http://bs-local.com:9339/notify";
-    self.sessionEndpoint = @"http://bs-local.com:9339/sessions";
+    self.notifyEndpoint = @"http://localhost:9339/notify";
+    self.sessionEndpoint = @"http://localhost:9339/sessions";
```
Cawllec, line 40: These will need to be changed back
```suggestion
    self.notifyEndpoint = @"http://bs-local.com:9339/notify";
    self.sessionEndpoint = @"http://bs-local.com:9339/sessions";
```

features/fixtures/macos/macOSTestApp/MainWindowController.m
```diff
@@ -97,6 +105,24 @@ - (IBAction)executeMazeRunnerCommand:(id)sender {
...
+        [self clearPersistentData:nil];
+    }
+    NSString *action = (NSString *)[environment objectForKey:@"BUGSNAG_SCENARIO_ACTION"];
+    BSLog(@"Received action: %@ for scenario: %@ and metadata: %@", action, self.scenarioName, self.scenarioMetadata);
+    if ([action isEqualToString:@"run_scenario"]) {
+        [self runScenario:nil];
+    } else if ([action isEqualToString:@"start_bugsnag"]) {
+        [self startBugsnag:nil];
+    }
+}
```
Cawllec, line 124: I'm running the scenarios using environment variables rather than the command api as:
- It's non-trivial to contact the command endpoint using an Async data task and then run scenarios compared to this method which all stays within the startup thread.
- This is a lot faster, as all of the information required to run the scenario is present when the app is started

features/steps/cocoa_steps.rb
```diff
@@ -3,57 +3,158 @@
 end
 
 When("I run {string} and relaunch the crashed app") do |event_type|
-  begin
-    step("I run \"#{event_type}\"")
-  rescue StandardError
-    # Ignore on macOS - AppiumForMac raises an error when clicking a button causes the app to crash
-    raise unless Maze::Helper.get_current_platform.eql?('macos')
```
Cawllec, line 10: Since this capture is only here to ignore these on MacOS we can removed it and allow the error to raise naturally.

features/support/env.rb
```diff
@@ -23,10 +23,6 @@
     unless File.exist?(app_path) || !File.exist?(zip_path)
       system("cd #{fixture_dir} && unzip -q #{zip_name}", exception: true)
     end
-
-    if File.exist?(app_path)
```
Cawllec, line 27: This is no longer required as we're launching the app directly.

Cawllec (commented):
You'll see I've collated the app_steps and cocoa_steps files together.  I plan to undo this by having an "app" file with all the application control steps included, and an "assertions" file including all the custom assertion steps included.

nickdowell:
> You'll see I've collated the app_steps and cocoa_steps files together. I plan to undo this by having an "app" file with all the application control steps included, and an "assertions" file including all the custom assertion steps included.

This is actually making it harder to review this PR - so much change and hard to focus on what's important.

Cawllec (commented):
Hmm, github went wrong there.  I'll get that change pushed up now.

Cawllec:
> > You'll see I've collated the app_steps and cocoa_steps files together. I plan to undo this by having an "app" file with all the application control steps included, and an "assertions" file including all the custom assertion steps included.
> 
> This is actually making it harder to review this PR - so much change and hard to focus on what's important.

Ok, I've attempted to address this, let me know if that helps.

features/steps/app_steps.rb
```diff
@@ -54,10 +111,66 @@
...
+    attempts += 1
+    assertion_passed = yield
+    sleep 0.1
+  end
+  raise 'Assertion not passed in 30s' unless assertion_passed
+end
+
+def run_macos_app
+  Maze::Runner.kill_running_scripts if $reset_data
+  Maze::Runner.run_command("features/fixtures/macos/output/#{Maze.config.app}.app/Contents/MacOS/#{Maze.config.app}", blocking: false)
```
nickdowell, line 172: Given we always pass the same app name, is it even worth making this configurable?

The steps are so highly coupled to the behaviour of the app, this "flexibility" seems like an illusion...

features/steps/app_steps.rb
```diff
@@ -1,25 +1,82 @@
...
 end
 
 When("I relaunch the app after a crash") do
   # Wait for the app to stop running before relaunching
-  step 'the app is not running'
   case Maze::Helper.get_current_platform
   when 'macos'
-    Maze.driver.get(Maze.driver.capabilities['app'])
+    # Allow any operations to complete before the next step
+    sleep(2)
```
nickdowell, line 65: No - sleeping is not a reliable solution. We need to wait for the process to die.

twometresteve: I think I gave a bad steer on this one - I'd said it was ok on the basis that overall the approach is still quicker than Appium-based and doesn't come with that baggage.  I didn't think about the reliability of the approach and that sometimes it may not be long enough.

Cawllec: I did have a more concrete reason for this, but I've made other changes since then so I'll switch this back and re-test, and see if it's made a difference where we don't require the wait.

nickdowell: But the tests should also (essentially) assert that the app did die in a crashing scenario.

Not dying can indicate a deadlock in the crash reporter.

features/steps/app_steps.rb
```diff
@@ -54,10 +111,66 @@
...
+    sleep 0.1
+  end
+  raise 'Assertion not passed in 30s' unless assertion_passed
+end
+
+def run_macos_app
+  Maze::Runner.kill_running_scripts if $reset_data
+  Maze::Runner.run_command("features/fixtures/macos/output/#{Maze.config.app}.app/Contents/MacOS/#{Maze.config.app}", blocking: false)
+  # Required to allow the non-blocking app to fully start before exiting
+  sleep(2)
```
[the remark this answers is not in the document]
Cawllec: I've implemented this, but unfortunately there isn't a `blocking` option.

nickdowell:
I think the app should be killed after running each scenario, which can be done with

```
Maze.hooks.after do |scenario|
  Process.kill('KILL', $fixture_pid) if $fixture_pid
  $fixture_pid = nil
end
```

features/steps/app_steps.rb
```diff
@@ -54,10 +108,71 @@
...
+
+def run_macos_app
+  `kill -s SIGTERM #{$fixture_pid}` if $fixture_pid
+  $fixture_pid = Process.spawn(
+    Maze::Runner.environment,
+    'features/fixtures/macos/output/macOSTestApp.app/Contents/MacOS/macOSTestApp',
+    [:err, :out] => 'macOSTestApp.log'
+  )
+  # Required to allow the non-blocking app to fully start before exiting
+  sleep(2)
```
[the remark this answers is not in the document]
Cawllec: I've shortened this to 0.1s, but removing it completely breaks the tests.

features/fixtures/macos/macOSTestApp/MainWindowController.m
nickdowell: There is a reusable class method on `Scenario` (used by the iOS fixture) which would make it trivial to use the command endpoint - https://github.com/bugsnag/bugsnag-cocoa/blob/master/features/fixtures/shared/scenarios/Scenario.m#L157-L192

Nevertheless this approach is fine for now 👍

twometresteve: Thanks - I'll capture that in a ticket and we can look again at some point in the future.  Really happy to see the back of Appium for now though.

twometresteve (approved):
Excellent.

PR #1301, opened 2022-02-18
Don't check session data contents if they are nil

kstenerud:
## Goal

When a session is resumed, we send an update notification with a nil object to indicate that there's no new session info to update:

        [center addObserver:self selector:@selector(sessionUpdateNotification:) name:BSGSessionUpdateNotification object:nil];

In our update handler, it fails the JSON check so the data doesn't get updated (as we want), but it also prints out an ominous error message:

    if (![BSGJSONSerialization isValidJSONObject:notification.object]) {
        bsg_log_err("Invalid session payload in notification");
        return;
    }

This PR does an explicit nil check so that it doesn't print an error message on session resume.

## Testing

Reran e2e tests

### DIFF
--- a/Bugsnag/BugsnagSystemState.m
+++ b/Bugsnag/BugsnagSystemState.m
@@ -243,11 +243,7 @@ static NSDictionary *copyDictionary(NSDictionary *launchState) {
 }
 
 - (void)sessionUpdateNotification:(NSNotification *)notification {
-    if (notification.object == nil) {
-        // No data to update
-        return;
-    }
-    if (![BSGJSONSerialization isValidJSONObject:notification.object]) {
+    if (notification.object && ![BSGJSONSerialization isValidJSONObject:notification.object]) {
         bsg_log_err("Invalid session payload in notification");
         return;
     }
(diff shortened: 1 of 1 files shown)


Bugsnag/BugsnagSystemState.m
```diff
@@ -243,6 +243,10 @@ - (instancetype)initWithConfiguration:(BugsnagConfiguration *)config {
 }
 
 - (void)sessionUpdateNotification:(NSNotification *)notification {
+    if (notification.object == nil) {
+        // No data to update
+        return;
+    }
     if (![BSGJSONSerialization isValidJSONObject:notification.object]) {
```
nickdowell, line 250: Ah I think this is a genuine problem - when a session is stopped or paused, `BugsnagSystemState` should clear `state[BSGKeySession]` so that it no longer thinks there is an active session. Currently we would end up reporting a session when we shouldn't... this would affect OOM and thermal kill events.

```suggestion
    if (notification.object && ![BSGJSONSerialization isValidJSONObject:notification.object]) {
```

nickdowell (changes requested):

Bugsnag/BugsnagSystemState.m
nickdowell: Note: we have no test coverage for this edge case, calling `pauseSession` and causing an OOM / thermal kill.

nickdowell (approved):
Tested it locally, LGTM 👍

PR #1312, opened 2022-02-28
[PLAT-5601] Fix crash in CPPExceptionTerminate when handling null exception

nickdowell:
## Goal

Fixes a crash in Bugsnag's std::terminate_handler (`CPPExceptionTerminate()`) if `throw` is called without an exception or `std::unexpected()` is called, that manifests itself as:

```
EXC_BAD_ACCESS: Stack overflow in ...

CPPExceptionTerminate
std::__terminate
std::terminate
CPPExceptionTerminate
std::__terminate
std::terminate
CPPExceptionTerminate
std::__terminate
std::terminate
(etc.)
```

## Changeset

Skips the "rethrow" logic and gets the current backtrace if there is no current exception.

```
errorClass: "std::terminate"
message: "throw may have been called without an exception"
```

## Testing

Amended E2E scenarios and verified fix on iOS and macOS.

Props to @kattrali for finding the repro case and identifying the solution! 🎉


kattrali (approved):
looks good!

PR #1317, opened 2022-03-04
macOS Appium removal tweaks

nickdowell:
* Stops fixture's log being truncated when relaunched in a scenario.
* Captures fixture's log in maze_output.
* Increases and documents launch delay.


nickdowell:
Note: if the mac fixture did a `GET /command`, we would have a way to wait until we know it's finished launching.

Cawllec (approved):

PR #1333, opened 2022-04-07
Refactor session serialization

nickdowell:
## Goal

Simplify & clarify `BugsnagSession` serialization & related code.

## Changeset

* `BSGSessionToDictionary()` replaces `-[BugsnagSession toJson]`
  * This produces JSON that is not suitable for including directly in payloads, but is used if sessions need to be persisted to disk for retrying later.
* `BSGSessionFromDictionary()` replaces `-[BugsnagSession initWithDictionary:]` and parses the above format.
* `BSGSessionToEventJson()` produces JSON in the format used within an event.
* `BSGSessionFromEventJson()` replaces `+[BugsnagSession fromJson:]` and parses the above format.
* Removes `BSGSessionUpdateNotification` and uses the existing callback mechanism instead.
* Removes some unused code.

## Testing

Tested by running full E2E suite and manual testing through example app.


Bugsnag/BugsnagSessionTracker.h
```diff
@@ -48,17 +48,6 @@ static NSNotificationName const BSGSessionUpdateNotification = @"BugsnagSessionC
  */
 - (void)startNewSessionIfAutoCaptureEnabled;
 
-/**
- Update the details of the current session to account for externally reported
- session information. Current session details are included in subsequent crash
- reports.
- */
-- (void)registerExistingSession:(NSString *)sessionId
```
kattrali, line 56: This is not unused https://github.com/bugsnag/bugsnag-unity/blob/bdd771d30f635f5ead5225186a5196355b888c63/src/BugsnagUnity.m#L356

nickdowell: Thanks, well spotted

Bugsnag/Payload/BugsnagSession+Private.h
```diff
@@ -47,18 +32,20 @@ NS_ASSUME_NONNULL_BEGIN
...
+/// Produces a session dictionary that contains all the information to fully recreate it via BSGSessionFromDictionary().
+NSDictionary * BSGSessionToDictionary(BugsnagSession *session);
 
-// Representation used in report payloads.
-- (NSDictionary *)toJson;
+/// Parses a session dictionary produced by BSGSessionToDictionary() or added to a KSCrashReport by BSSerializeDataCrashHandler().
+BugsnagSession *_Nullable BSGSessionFromDictionary(NSDictionary *_Nullable json);
 
-/// Full representation of a session suitable for creating an identical session using -initWithDictionary:
-- (NSDictionary *)toDictionary;
```
nickdowell, line 60: `toDictionary` is used by `bugsnag-unity`

https://github.com/bugsnag/bugsnag-unity/blob/bdd771d30f635f5ead5225186a5196355b888c63/src/BugsnagUnity.m#L369

Bugsnag/BugsnagSessionTracker.h
nickdowell: Restored and documented

Bugsnag/BugsnagSessionTracker.h
```diff
@@ -49,9 +43,7 @@ static NSNotificationName const BSGSessionUpdateNotification = @"BugsnagSessionC
 - (void)startNewSessionIfAutoCaptureEnabled;
 
 /**
- Update the details of the current session to account for externally reported
- session information. Current session details are included in subsequent crash
- reports.
+ Used by bugsnag-unity
```
kattrali, line 46: This is useful, though the old comment explained what this method actually does. "externally reported" always meant "libraries built on top" so the comment could be expanded a bit to be more specific.

nickdowell: Hmm I did start rewriting this and then saw that I was essentially repeating the name of the method 🤔

kattrali: Yeah I could see that. Though it might be otherwise lost to time _why_ this would ever be done. What we probably need is a unified place for these kinds of hooks for separate libraries. I think @lemnik had some ideas on this since Android has a similar issue.

nickdowell: Yeah I was thinking of some kind of attribute we could use to tag APIs used be other notifiers.

Refactoring is so much harder when other code depends on internals 😬

Bugsnag/Payload/BugsnagSession+Private.h
nickdowell: https://github.com/bugsnag/bugsnag-unity/pull/547

kattrali (approved):
The `EventJson` vs `Dictionary` distinction is much clearer :+1:

Bugsnag/BugsnagSessionTracker.m
```diff
@@ -117,32 +115,22 @@ - (void)setCodeBundleId:(NSString *)codeBundleId {
 
 #pragma mark - Creating and sending a new session
 
-- (void)startNewSession {
-    [self startNewSessionWithAutoCaptureValue:NO];
-}
-
 - (void)pauseSession {
-    [[self currentSession] stop];
+    self.currentSession.stopped = YES;
```
kstenerud, line 119: It feels weird to reach into a session from another object to change its internal state rather than encapsulating start/stop/resume in BugsnagSession proper.

nickdowell: I can see the argument, but it's not unusual to set the value of an object's property. The `stopped` property is also not really "internal" since it's declared in the header. Furthermore the session object never examines it, only `BugsnagSessionTracker` does.

kstenerud: It's not the end of the world, but the fact that a property defining the session's state is writable, and is only set externally by an administrator object, is a code smell. Normally an object would manage its own state internally and present a view to the world for inspecting (not directly setting) that state.

PR #1339, opened 2022-04-14
[PLAT-8359] Add BSGRunContext

nickdowell:
## Goal

Establish new memory-mapped storage for changeable app / device / system info.

## Changeset

Adds `struct BSGRunContext` and basic routines to load existing data, set up mapping and initialise global pointers.

Settled on "runcontext" as it is a term not currently used in the codebase. Other rejected names considered include "runinfo", "runstate", "systemstate" and "systeminfo".

## Testing

Tested manually by debugging example app.

### DIFF
--- a/Bugsnag/Helpers/BSGRunContext.h
+++ b/Bugsnag/Helpers/BSGRunContext.h
@@ -5,6 +5,13 @@
 //  Copyright © 2022 Bugsnag Inc. All rights reserved.
 //
 
+//
+// The struct version should be incremented prior to a release if changes have
+// been made to BSGRunContext.
+//
+// During development this is not strictly necessary since last run's data will
+// not be loaded if the struct's size has changed.
+//
 #define BSGRUNCONTEXT_VERSION 1
 
 struct BSGRunContext {
--- a/Bugsnag/Helpers/BSGRunContext.h
+++ b/Bugsnag/Helpers/BSGRunContext.h
@@ -1,0 +1,24 @@
+//
+//  BSGRunContext.h
+//  Bugsnag
+//
+//  Copyright © 2022 Bugsnag Inc. All rights reserved.
+//
+
+#define BSGRUNCONTEXT_VERSION 1
+
+struct BSGRunContext {
+    long structVersion;
+};
+
+/// Information about the current run of the app / process.
+///
+/// This structure is mapped to a file so that changes will be persisted by the OS.
+///
+/// Guaranteed to be non-null once BSGRunContextInit() is called.
+extern struct BSGRunContext *_Nonnull bsg_runContext;
+
+/// Information about the last run of the app / process, if it could be loaded.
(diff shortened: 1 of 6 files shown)


Bugsnag/Helpers/BSGRunContext.h
```diff
@@ -0,0 +1,24 @@
+//
+//  BSGRunContext.h
+//  Bugsnag
+//
+//  Copyright © 2022 Bugsnag Inc. All rights reserved.
+//
+
+#define BSGRUNCONTEXT_VERSION 1
```
kstenerud, line 8: Can we put a comment here that discusses how/when this gets updated?

Bugsnag/Helpers/BSGRunContext.h
```diff
@@ -0,0 +1,24 @@
...
+
+/// Information about the current run of the app / process.
+///
+/// This structure is mapped to a file so that changes will be persisted by the OS.
+///
+/// Guaranteed to be non-null once BSGRunContextInit() is called.
+extern struct BSGRunContext *_Nonnull bsg_runContext;
+
+/// Information about the last run of the app / process, if it could be loaded.
+extern const struct BSGRunContext *_Nullable bsg_lastRunContext;
```
kstenerud, line 22: Do we have a way to ensure that the data loaded from disk is valid / safe to use? For example:

* App crashes immediately after memmapping - Does the file contain random data?
* App crashes immediately after memmapping, clearing, setting version - Is zeroed data safe to read from?

Bugsnag/Helpers/BSGRunContext.h
nickdowell: I've added some comments here, although it's made me question whether we really need a version field at this stage - checking the struct size is enough...

Bugsnag/Helpers/BSGRunContext.h
nickdowell: When the file is created, `ftruncate()` fills it with zeros. It will never contain random data.

`structVersion` will be set _last_ to prevent partially initialised data from being read.

Bugsnag/Helpers/BSGRunContext.h
kstenerud: I think we should keep it to stay on the safe side, in case a struct change results in a total size we've already used in the past.

Also, a "struct data is valid" flag would probably solve the other issue I mentioned. Set it to true once struct initialization is complete. Then we have three checks to make: Does the version match, does the size match, is the "valid" flag true.

nickdowell: Sure, it doesn't hurt to check the version.

That said, I think it makes sense to follow a rule of never removing a field from the struct.

The `structVersion` field will also act as a "struct data is valid" flag.

kstenerud: Sorry for the ninja edit :)

kstenerud (approved):

PR #1340, opened 2022-04-19
[PLAT-8360] Replace KV-store

nickdowell:
## Goal

Migrate data from KV-store to `BSGRunContext`.

## Changeset

Removes `BugsnagKVStore`.

Deletes the `kvstore` directory at startup.

Adds `BSGRunContext` fields for data previously stored in kvstore.

Refactors logic for setting these fields, moving the code into `BSGRunContext.m` for better code locality.

See individual commits for more logical diffs.

## Testing

Amended unit tests to work with `BSGRunContext` (e.g. some of them require a foreground state which is not true under xctest.)

Ran full E2E test suite.


Bugsnag/BugsnagSystemState.m
```diff
@@ -18,58 +18,39 @@
...
-#import "BugsnagSystemState.h"
 
 #import <stdatomic.h>
 
+#define SYSTEMSTATE_APP_VERSION @"version"
+#define SYSTEMSTATE_APP_BUNDLE_VERSION @"bundleVersion"
+
+#define SYSTEMSTATE_DEVICE_BOOT_TIME @"bootTime"
+
 static NSString * const ConsecutiveLaunchCrashesKey = @"consecutiveLaunchCrashes";
```
kstenerud, line 36: Not for this PR, but our styling and define vs const should be harmonized.

nickdowell: Agreed - this will be removed in the next PR btw

kstenerud (approved):

PR #1341, opened 2022-04-20
[PLAT-8361] Move kill detection to BSGRunContext

nickdowell:
## Goal

Move remaining data for detection of unexpected terminations (OOMs and Thermal Kills) to `BSGRunContext`.

## Changeset

Stores bootTime and the apps's Mach-O UUID in `BSGRunContext`.

Mach-O UUID is now compared instead of app version & bundle version to determine whether the app has been upgraded.

Moves logic from `-[BugsnagSystemState lastLaunchTerminatedUnexpectedly]` to `BSGRunContextWasKilled()` since it is tied to `BSGRunContext`.

## Testing

Amended existing tests to work with `BSGRunContext`

### DIFF
--- a/Bugsnag/Helpers/BSGRunContext.m
+++ b/Bugsnag/Helpers/BSGRunContext.m
@@ -188,6 +188,46 @@ CFNotificationSuspensionBehaviorDeliverImmediately)
 }
 
 
+#pragma mark - Kill detection
+
+bool BSGRunContextWasKilled() {
+    // App extensions have a different lifecycle and the heuristic used for
+    // finding app terminations rooted in fixable code does not apply
+    if ([BSG_KSSystemInfo isRunningInAppExtension]) {
+        return NO;
+    }
+    
+    if (!bsg_lastRunContext) {
+        return NO;
+    }
+    
+    if (bsg_lastRunContext->isTerminating) {
+        return NO; // The app terminated normally
+    }
+    
+    if (bsg_lastRunContext->isDebuggerAttached) {
+        return NO; // The debugger may have killed the app
+    }
+    
+    // Once the app is in the background we cannot determine between good (user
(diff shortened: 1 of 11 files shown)


Bugsnag/Helpers/BSGRunContext.m
```diff
@@ -26,6 +28,7 @@
 
 #pragma mark Forward declarations
 
+static uint64_t BSGRunContextGetBootTime(void);
```
kstenerud, line 31: Style nit: I feel that prefixing static functions in the same way as public functions makes it harder to tell which is API and which is not. We don't need to artificially namespace these since their names are not exported.

kstenerud (approved):

Bugsnag/Helpers/BSGRunContext.m
nickdowell: https://github.com/bugsnag/bugsnag-cocoa/pull/1342

features/fixtures/shared/scenarios/OOMScenario.m
```diff
@@ -20,9 +20,14 @@
 
 @implementation OOMScenario
 
+void onCrashHandler(const BSG_KSCrashReportWriter *writer) {
+    assert(!!"onCrashHandler should not be called for OOMs");
```
kattrali, line 24: Isn't this always true, so the `assert()` does nothing?

nickdowell: You're right, this should be

```suggestion
    assert(!"onCrashHandler should not be called for OOMs");
```

nickdowell: Fixed in d2ddf07beb4bf2d1b60a5ac08ffc0776c9c7b7d9

PR #1343, opened 2022-04-21
[PLAT-8362] Store session info in BSGRunContext

nickdowell:
## Goal

Reduce the overhead of persisting session information needed to build OOM and Thermal Kill payloads.

## Changeset

Moves (C-level) session info storage into `BSGRunContext` and removes session info from `BugsnagSystemState`.

Logic to convert sessions objects to and from BSGRunContext has been moved to `BugsnagSession.m` to make it easy to compare reading and writing code.

## Testing

Passed full E2E run, which includes scenarios that verify presence of session data in OOM and crash payloads.

### DIFF
--- a/Bugsnag/Payload/BugsnagSession+Private.h
+++ b/Bugsnag/Payload/BugsnagSession+Private.h
@@ -8,6 +8,10 @@
 
 #import <Bugsnag/BugsnagSession.h>
 
+#import "BSG_KSCrashReportWriter.h"
+
+#define BSG_PRIVATE __attribute__((visibility("hidden")))
+
 NS_ASSUME_NONNULL_BEGIN
 
 @class BugsnagUser;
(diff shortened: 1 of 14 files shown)


Bugsnag/Payload/BugsnagSession+Private.h
```diff
@@ -8,6 +8,10 @@
 
 #import <Bugsnag/BugsnagSession.h>
 
+#import "BSG_KSCrashReportWriter.h"
+
+#define BSG_PRIVATE __attribute__((visibility("hidden")))
```
kstenerud, line 13: Can we put this in a utility header so that we can use it elsewhere?

kstenerud (approved):

Bugsnag/Payload/BugsnagSession+Private.h
nickdowell: I agree we should, but wanted to leave that to a later PR

Maybe `BSGDefines.h` as a general purpose one?

I'm keen to avoid adding too many new headers, as we already have so many.

PR #1347, opened 2022-04-25
[PLAT-8363] Move changeable device data to BSGRunContext

nickdowell:
## Goal

Store changeable device data in `BSGRunContext` instead of writing to disk in response to notifications.

## Changeset

Adds `batteryLevel`, `batteryState`, `lastKnownOrientation` and memoryPressure to `BSGRunContext`.

Removes thermal state from `BugsnagClient.metadata` to avoid disk I/O.

Updates handled event generation to include the new data.

Updates OOM & Thermal Kill event generation to include the new data.

Updates `BSSerializeDataCrashHandler()` to include the new data in crash reports and `BugsnagEvent`'s corresponding crash report parsing code.

Moves breadcrumb related code to `BSGNotificationBreadcrumbs`.

## Testing

Updates E2E tests to perform more thorough checking. `metaData.device.thermalState` is not verified in most scenarios due to being unavailable on iOS 10.

Successfully passes full E2E run.


kstenerud (approved):
Liking how clean this is getting!

PR #1355, opened 2022-05-05
[PLAT-8410] Add `device.time` to OOM and Thermal Kill events

nickdowell:
## Goal

Provide a timestamp for OOM and Thermal Kill events so that the dashboard renders breadcrumbs with sensible relative timestamps.

## Changeset

Adds a `timestamp` field to `BSGRunContext` which is updated from a timer and in response to certain events (so that `device.time >= breadcrumbs.last.timestamp`)

Sets `device.time` for OOM and TK events based on the last run's context.

Uses `CFAbsoluteTimeGetCurrent()` as the source of timing information, which provides a wall-clock time equivalent to `[NSDate date].timeIntervalSinceReferenceDate`.

## Performance

`CFAbsoluteTime()` is very fast on recent OS versions (~70 nanoseconds on iPhone 5s with iOS 12) where it does not need to call into the kernel.

On iPad 3 running iOS 9 it takes ~1400 nanoseconds (1.4 µs) which while much slower is still reasonable for the frequency at which this code calls it. For comparison, a single character `write()` to stdout takes ~8 µs.

Considered using `mach_approximate_time()` which is _much_ faster (~45 nanoseconds on iPad 3) but since it would be difficult (perhaps impossible?) to convert to the required wall-clock time and only old devices / OSes would benefit, the extra complexity does not feel justified.

## Testing

Amends OOM barebone E2E scenario to verify presence of `device.time`.

Manually verified outcome on dashboard using example app.


nickdowell:
Have not changed `BSGRUNCONTEXT_VERSION` because we have not yet made a release that includes `BSGRunContext`

kattrali (approved):

PR #1357, opened 2022-05-09
[PLAT-8411] Add `device.freeMemory` to OOM and Thermal Kill events

nickdowell:
## Goal

Provide recent memory statistics to improve confidence in OOM errors.

## Changeset

Adds an `availableMemory` field to `BSGRunContext` and updates this in the timer callback (twice per second) and in response to a memory pressure notification.

Only on iOS/tvOS 13 and later where `os_proc_available_memory()` is available.

ℹ️ Does not update `BSGRUNCONTEXT_VERSION` since no release yet contains BSGRunContext.

## Performance

`os_proc_available_memory()` is much faster than calling `host_statistics()` or `task_info()`... ~150 vs ~700 nanoseconds on an iPhone SE (2020).

## Testing

Amends E2E scenarios to verify presence of `device.freeMemory`

### DIFF
--- a/Bugsnag/Helpers/BSGRunContext.m
+++ b/Bugsnag/Helpers/BSGRunContext.m
@@ -17,6 +17,10 @@
 #import <sys/stat.h>
 #import <sys/sysctl.h>
 
+#if __has_include(<os/proc.h>) && TARGET_OS_IPHONE && !TARGET_OS_MACCATALYST
+#include <os/proc.h>
+#endif
+
 #if TARGET_OS_IOS
 #import "BSGUIKit.h"
 #endif
@@ -31,6 +35,7 @@
 static uint64_t GetBootTime(void);
 static bool GetIsForeground(void);
 static void InstallTimer(void);
+static void UpdateAvailableMemory(void);
 
 
 #pragma mark - Initial setup
@@ -58,6 +63,8 @@ static void InitRunContext() {
     
     BSGRunContextUpdateTimestamp();
     InstallTimer();
+    
+    UpdateAvailableMemory();
     
     // Set `structVersion` last so that BSGRunContextLoadLast() will reject data
     // that is not fully initialised.
@@ -148,6 +155,7 @@ static void InstallTimer() {
     
     dispatch_source_set_event_handler(timer, ^{
         BSGRunContextUpdateTimestamp();
+        UpdateAvailableMemory();
     });
     
     dispatch_resume(timer);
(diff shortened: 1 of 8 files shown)


features/barebone_tests.feature
```diff
@@ -278,7 +279,5 @@ Feature: Barebone tests
     And the error payload field "events.0.app.duration" is null
     And the error payload field "events.0.app.durationInForeground" is null
     And the error payload field "events.0.device.freeDisk" is null
-    And the error payload field "events.0.device.freeMemory" is null
```
kstenerud, line 281: Can this be checked for an integer as well?

kattrali (approved):

Bugsnag/Helpers/BSGRunContext.m
```diff
@@ -272,12 +281,22 @@ static void AddObservers() {
...
-#pragma mark -
+#pragma mark - Misc
 
 void BSGRunContextUpdateTimestamp() {
     bsg_runContext->timestamp = CFAbsoluteTimeGetCurrent();
 }
 
+static void UpdateAvailableMemory() {
+    // Deliberately avoids use of bsg_ksmachfreeMemory() because that falls back
+    // to a much more expensive (~5x) system call on earlier releases.
```
tomlongridge, line 292: Just to be clear, what I understand from the comments and `bsg_ksmachfreeMemory` code, is that freeMemory will be reported from the quicker `os_proc_available_memory` API if available, but on simulators or older devices the crash report falls back to a slower API but the OOM report will omit the field.

I see the check in `BSG_KSMach.c` does a check for a 0 response to see if "it's working" - would we want to do this here?

nickdowell: On simulators or older devices OOMs will not include `freeMemory`.

`bsg_ksmachfreeMemory()` falls back to using the `free_count` from the `HOST_VM_INFO` flavor of `host_statistics()`, which returns system-wide VM statistics, whereas `os_proc_available_memory()` returns the amount remaining in the current process' dirty memory limit.

The two values could be wildly different given that memory limits are typically 50-80% of system memory for foreground apps, and much smaller for app extensions. Using the fallback could be confusing or misleading in addition to being much slower, since it's hitting the memory limit that causes apps to be killed.

The check for `0` is done when constructing the event - [here](https://github.com/bugsnag/bugsnag-cocoa/pull/1357/files#diff-9e40e69405013ebceec8e93a7bb9f77586d4fbecc564622b874b364e0b772d29R1129).

PR #1362, opened 2022-05-11
Delete sessions if they are more than 60 days old

nickdowell:


Bugsnag/Delivery/BSGSessionUploader.m
```diff
@@ -141,3 +188,27 @@ - (void)sendSession:(BugsnagSession *)session completionHandler:(nonnull void (^
...
+
+static NSArray * SortedFiles(NSFileManager *fileManager, NSMutableDictionary<NSString *, NSDate *> **outDates) {
+    NSString *dir = BSGFileLocations.current.sessions;
+    NSMutableDictionary<NSString *, NSDate *> *dates = [NSMutableDictionary dictionary];
+    
+    for (NSString *name in [fileManager contentsOfDirectoryAtPath:dir error:nil]) {
+        NSString *file = [dir stringByAppendingPathComponent:name];
+        NSDate *date = [fileManager attributesOfItemAtPath:file error:nil].fileCreationDate;
+        if (!date) {
+            bsg_log_debug(@"Could not determine fileCreationDate for session %@",
```
kstenerud, line 201: Should it delete the file here? Otherwise it will only get removed when pruning after enough of them pile up, which could lead to some edge cases where non-stale files get removed.

kstenerud (approved):

kstenerud (approved):

PR #1373, opened 2022-05-18
Add watchOS BugsnagNetworkRequestPlugin scheme

kstenerud:
Xcode sometimes lies about creating a public scheme for a target. You have to manually unselect "public" and then re-select it to generate the file :/

Also fixed up the watchOS versions for the network plugin targets. Minimum OS is 6.0 (where NSURLSessionTaskMetrics are introduced)


BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin.xcodeproj/project.pbxproj
```diff
@@ -1141,7 +1149,7 @@
 				TVOS_DEPLOYMENT_TARGET = 9.2;
 				VERSIONING_SYSTEM = "apple-generic";
 				VERSION_INFO_PREFIX = "";
-				WATCHOS_DEPLOYMENT_TARGET = 4.0;
+				WATCHOS_DEPLOYMENT_TARGET = 6.0;
```
nickdowell, line 1152: Aren't we targeting 2.0?

kstenerud: Yes, but this plugin won't work without `NSURLSessionTaskMetrics`, which requires watchos 6.

nickdowell: But the podspec says `2.0`

https://github.com/bugsnag/bugsnag-cocoa/blob/watchos-network-plugin-scheme/BugsnagNetworkRequestPlugin.podspec.json#L22

This would mean Cocoapods would happily add the BugsnagNetworkRequestPlugin but there'd then be a link error.

The iOS deployment target is set to 9.0 but the logic only kicks in on iOS 10 and above... this is the same pattern we should be following here.

nickdowell: (this makes it easier for devs to integrate into their projects)

kstenerud: I've updated the podspec to 6.0

nickdowell (approved):

PR #1376, opened 2022-05-19
Update minimum version to what we can actually test on

kstenerud:
As per our zoom conversation, bump the minimum watchOS version to the minimum we can run tests on.


BugsnagNetworkRequestPlugin/BugsnagNetworkRequestPlugin/BSGURLSessionTracingDelegate.m
```diff
@@ -84,7 +84,7 @@ + (nonnull NSString *)URLStringWithoutQueryForComponents:(nonnull NSURLComponent
 }
 
 - (void)URLSession:(__unused NSURLSession *)session task:(NSURLSessionTask *)task didFinishCollectingMetrics:(NSURLSessionTaskMetrics *)metrics
-API_AVAILABLE(macosx(10.12), ios(10.0), watchos(3.0), tvos(10.0)) {
+API_AVAILABLE(macosx(10.12), ios(10.0), watchos(6.0), tvos(10.0)) {
```
nickdowell, line 87: I don't think these need to change

kstenerud: They won't work or even link on watchos < 6.0

nickdowell: Why is that? The headers indicate that the underlying APIs are available as of watchOS 3.0

```
API_AVAILABLE(macosx(10.12), ios(10.0), watchos(3.0), tvos(10.0))
@interface NSURLSessionTaskMetrics : NSObject
```

kstenerud: Yes, they do, but when you actually try to compile it fails in the linker stage (some kind of bug maybe?). Moot point anyway since it fails to compile unless I leave it at 3.0 :P

nickdowell (approved):

PR #1379, opened 2022-05-19
[PLAT-8456] Fix stack trace capture on watchOS

kstenerud:
Get stack traces working properly on watchOS. The main thread is captured except for C++ exceptions.

Also add missing scheme for example app.


Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
```diff
@@ -180,6 +180,9 @@ static void CPPExceptionTerminate(void) {
 #if BSG_HAVE_MACH_THREADS
         BSG_KSLOG_DEBUG("Suspending all threads.");
         bsg_kscrashsentry_suspendThreads();
+#else
+        // We still need the threads list for other purposes.
```
nickdowell, line 184: What other purposes?

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -460,6 +460,13 @@ uintptr_t *bsg_kscrw_i_getBacktrace(
         return backtraceBuffer;
      }
 
+#if TARGET_OS_WATCH
+    // The C++ exception stack trace trick doesn't work on watchOS, so don't capture it.
```
nickdowell, line 464: i.e. our `__cxa_throw` function is not called?

kstenerud: That code is unchanged, so it will still attempt, and fail for reasons I haven't figured out yet.

A task for later is to see if this can be made to work again.

PR #1380, opened 2022-05-19
[PLAT-8477] Redact internal errors

nickdowell:
## Goal

Use a different device ID, stop including the unparseable JSON payload, and improve grouping of internal errors.

## Changeset

Adds a new function for getting a unique device ID for internal errors that differs from `+[BSG_KSSystemInfo deviceAndAppHash]`.

Removes `BugsnagDiagnostics.data`.

Implements a new mechanism for reporting parsing errors:
* `context` is now used for grouping (in combination with `errorClass`) rather than specifying a `groupingHash`.
* `context` is set to the type of payload error - e.g. empty file, missing `{`, etc.
* `BugsnagDiagnostics.keys` is set to a list of the expected keys that were found in the valid portion of the JSON payload, to give an indication of where the error was.

## Testing

`BSGEventUploadKSCrashReportOperationTests` has been updated to verify internal errors.

### DIFF
--- a/Bugsnag/Delivery/BSGEventUploadKSCrashReportOperation.m
+++ b/Bugsnag/Delivery/BSGEventUploadKSCrashReportOperation.m
@@ -20,12 +20,44 @@
 #import "BugsnagLogger.h"
 
 
+/// Returns a list of the crash report keys present in the valid portion of the JSON data
+static NSArray * CrashReportKeys(NSData *data, NSError *error) {
+    NSString *description = error.userInfo[NSDebugDescriptionErrorKey]; 
+    for (NSString *separator in @[@" around character ", @"around line 1, column "]) {
+        if ([description containsString:separator]) {
+            NSUInteger end = (NSUInteger)[description componentsSeparatedByString:separator].lastObject.intValue;
+            if (!end) {
+                return nil;
+            }
+            NSData *subdata = [data subdataWithRange:NSMakeRange(0, end)];
+            if (!subdata) {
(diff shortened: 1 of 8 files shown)


Bugsnag/Delivery/BSGEventUploadKSCrashReportOperation.m
```diff
@@ -59,32 +93,38 @@ - (BugsnagEvent *)loadEventAndReturnError:(NSError * __autoreleasing *)errorPtr
...
+            return nil;
+        }
+        
+        reportError(@"JSON parsing error", data);
         return nil;
     }
     
     NSDictionary *crashReport = [self fixupCrashReport:json];
     if (!crashReport) {
-        ReportInternalError(@"Unexpected JSON payload", @"-fixupCrashReport: returned nil", @{@"json": json});
```
nickdowell, line 83: Removed because none have ever been reported.

kstenerud (approved):
Would it make sense to include the file contents if the file is small enough?

Or maybe the first and last 30 bytes of the file, since the beginning and end are the most likely points of failure?

nickdowell:
> Would it make sense to include the file contents if the file is small enough?
> 
> Or maybe the first and last 30 bytes of the file, since the beginning and end are the most likely points of failure?

From a technical point of view, yes - but to prevent any potentially sensitive information (e.g. from metadata or breadcrumbs) appearing in our events it was decided that we should no longer include any raw data.

PR #1387, opened 2022-05-23
Pass NSException to the NSException handler with no extra processing

kstenerud:
## Goal

NSException is actually implemented using the C++ exception ABI. The Bugsnag C++ handler was doing some preprocessing of the NSException before passing it to the default C++ handler (which then calls the NSException handler). After this, the NSException handler would do the same processing over again.

Since both the C++ and NSException handlers call `bsg_kscrashsentry_beginHandlingCrash`, and this function checks for re-entry to detect a crash in the crash handler, it gets confused into thinking that a crash has happened in the crash handler when there's actually nothing amiss. The resulting report on the dashboard is still correct, but the messaging on the device is misleading, and the crash handler is doing unnecessary extra work.

## Design

Don't do any processing in the C++ handler if it has caught an NSException. Simply pass it along and let the NSException handler do everything (basically behaving the same way it would if C++ exception capture were disabled).

## Testing

Tested manually on real iphone, watch, macos, and simulator tvos.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
@@ -105,7 +105,6 @@ void __cxa_throw(void *thrown_exception, std::type_info *tinfo,
 static void CPPExceptionTerminate(void) {
     BSG_KSLOG_DEBUG("Trapped c++ exception");
 
-    bool isNSException = false;
     char descriptionBuff[DESCRIPTION_BUFFER_LENGTH];
     const char *name = NULL;
     const char *crashReason = NULL;
@@ -131,15 +130,13 @@ static void CPPExceptionTerminate(void) {
     try {
         throw;
     } catch (NSException *exception) {
-        BSG_KSLOG_DEBUG("Detected NSException. Recording details and letting "
-                        "the current NSException handler deal with it.");
-        isNSException = true;
-        // recordException() doesn't call beginHandlingCrash()
-        if (bsg_kscrashsentry_isNSExceptionHandlerInstalled() &&
-            bsg_kscrashsentry_beginHandlingCrash(bsg_ksmachthread_self())) {
-            bsg_recordException(exception);
-            bsg_kscrashsentry_endHandlingCrash();
+        if (bsg_g_originalTerminateHandler != NULL) {
+            BSG_KSLOG_DEBUG("Detected NSException. Passing to the current NSException handler.");
+            bsg_g_originalTerminateHandler();
+        } else {
+            BSG_KSLOG_DEBUG("Detected NSException, but there was no original C++ terminate handler.");
         }
+        return;
     } catch (std::exception &exc) {
         strlcpy(descriptionBuff, exc.what(), sizeof(descriptionBuff));
         crashReason = descriptionBuff;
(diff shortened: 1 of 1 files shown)


Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
```diff
@@ -131,15 +130,13 @@ static void CPPExceptionTerminate(void) {
...
-                        "the current NSException handler deal with it.");
-        isNSException = true;
-        // recordException() doesn't call beginHandlingCrash()
-        if (bsg_kscrashsentry_isNSExceptionHandlerInstalled() &&
-            bsg_kscrashsentry_beginHandlingCrash(bsg_ksmachthread_self())) {
-            bsg_recordException(exception);
-            bsg_kscrashsentry_endHandlingCrash();
+        if (bsg_g_originalTerminateHandler != NULL) {
+            BSG_KSLOG_DEBUG("Detected NSException. Passing to the current NSException handler.");
+            bsg_g_originalTerminateHandler();
```
nickdowell, line 135: How do NSExceptions now get captured?

Does `bsg_g_originalTerminateHandler` end up calling `bsg_ksnsexc_i_handleException` via `NSUncaughtExceptionHandler`?

kstenerud: The ObjC runtime registers a C++ exception handler that checks for NSException and does processing if it is (calling whatever NSUncaughtExceptionHandler is registered, or calling `abort()` if none is).

We tap into both the C++ exception handler and the NSException handler, but since we don't want to process twice, we check for NSException and just pass the exception along for the ObjC runtime to find.

nickdowell: What would happen if bsg_g_originalTerminateHandler is (for example) Crashlytics's NSUncaughtExceptionHandler?

Will this PR mean that crashes we previously reported in apps that use both would no longer be detected?

kstenerud: Hmm actually, if someone else injects themselves into the callback chain AND forgets to pass it along, we'd get an abort signal instead (if you hook the c++ exception callback and don't pass it along to the next handler, the ObjC runtime will never see it).

So, also capturing information at the C++ level would give an extra defence against poorly implemented third party software (but only if they hook the handler BEFORE we do; if they hook after us and forget to pass it on, all is lost).

kstenerud:
Closed in favour of https://github.com/bugsnag/bugsnag-cocoa/pull/1389

Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry_CPPException.mm
kstenerud: Nope, my initial testing was flawed, and further testing showed that this C++ code is not capturing NSException data after all. So we're better off going with the approach in this PR to ensure that `bsg_kscrashsentry_beginHandlingCrash` never gets called more than once for a single crash.

kstenerud:
Reopening because the other approach leaves open a concurrency risk.

nickdowell (approved):

PR #1388, opened 2022-05-23
Never collect threads when they're not supported

kstenerud:
## Goal

The library is still capturing minimal, but basically useless information about threads when the mach threads interface is not supported. Instead, just omit thread information altogether when they're not supported.

## Testing

Manual testing on simulators and real watch, iphone, mac.


kattrali (approved):
Looks good, though there will also need to be a note about the `sendThreads` config option having no effect for this case.

PR #1389, opened 2022-05-24
[Plat-8488] Fix false "crash in handler" messaging

kstenerud:
## Goal

Calling `bsg_kscrashsentry_endHandlingCrash` doesn't fully clear the crash handling state, so calling `bsg_kscrashsentry_beginHandlingCrash` a second time afterwards will issue a false alarm about a crash in the crash handler.

This PR ensures that the state is correct after calling `bsg_kscrashsentry_endHandlingCrash`, and explicitly checks that flag in `bsg_kscrashsentry_beginHandlingCrash` rather than relying on a thread ID comparison.

## Testing

Manual verification on all major devices.

### DIFF
--- a/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry.c
+++ b/Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry.c
@@ -229,7 +229,7 @@ bool bsg_kscrashsentry_beginHandlingCrash(const thread_t offender) {
         return true;
     }
 
-    if (offender == firstHandlingThread) {
+    if (bsg_g_context->handlingCrash) {
         BSG_KSLOG_INFO("Detected crash in the crash reporter. "
                        "Restoring original handlers.");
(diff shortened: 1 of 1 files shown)


Bugsnag/KSCrash/Source/KSCrash/Recording/Sentry/BSG_KSCrashSentry.c
```diff
@@ -229,7 +229,7 @@ bool bsg_kscrashsentry_beginHandlingCrash(const thread_t offender) {
         return true;
     }
 
-    if (offender == firstHandlingThread) {
+    if (bsg_g_context->handlingCrash) {
```
nickdowell, line 232: `if (offender == firstHandlingThread)` positively checks that the crash handler crashed, rather than just a concurrent crash outside the crash handler.

kstenerud: `offender == firstHandlingThread` is what leads to the false "crash in crash reporter" message, because the thread that captures the C++ or NSException is the offending thread. So when it goes through this path twice (once for C++, once for NSException), it triggers.

PR #1391, opened 2022-05-25
Record battery information on watchOS

kstenerud:
## Goal

Records the battery level and charging status when the app starts.

Since watchOS doesn't have a notification API for battery like iOS does, only the initial status is recorded.

## Testing

Manually tested on all platforms.

### DIFF
--- a/Bugsnag.xcodeproj/project.pbxproj
+++ b/Bugsnag.xcodeproj/project.pbxproj
@@ -843,7 +843,6 @@
 		CB28F0D3282A4B91003AB200 /* BugsnagErrorTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 008966B52486D43500DC48C2 /* BugsnagErrorTest.m */; };
 		CB28F0D4282A4B91003AB200 /* BugsnagEventFromKSCrashReportTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 008966C02486D43600DC48C2 /* BugsnagEventFromKSCrashReportTest.m */; };
 		CB28F0D5282A4B91003AB200 /* BugsnagDeviceTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 008966A42486D43400DC48C2 /* BugsnagDeviceTest.m */; };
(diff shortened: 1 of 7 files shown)


Bugsnag.xcodeproj/project.pbxproj
```diff
@@ -994,13 +999,13 @@
 		CBBDE972280069540070DCD3 /* BugsnagStacktrace.h in Headers */ = {isa = PBXBuildFile; fileRef = 0089684F2486DA9400DC48C2 /* BugsnagStacktrace.h */; };
 		CBBDE973280069540070DCD3 /* BugsnagStacktrace.m in Sources */ = {isa = PBXBuildFile; fileRef = 0089685C2486DA9500DC48C2 /* BugsnagStacktrace.m */; };
 		CBBDE9742800695A0070DCD3 /* BugsnagUser.m in Sources */ = {isa = PBXBuildFile; fileRef = 0089685F2486DA9500DC48C2 /* BugsnagUser.m */; };
-		CBBDE975280069670070DCD3 /* BugsnagFileStore.h in Headers */ = {isa = PBXBuildFile; fileRef = 008968E62486DAB800DC48C2 /* BugsnagFileStore.h */; };
-		CBBDE976280069670070DCD3 /* BugsnagSessionFileStore.m in Sources */ = {isa = PBXBuildFile; fileRef = 008968E52486DAB800DC48C2 /* BugsnagSessionFileStore.m */; };
+		CBBDE975280069670070DCD3 /* (null) in Headers */ = {isa = PBXBuildFile; };
+		CBBDE976280069670070DCD3 /* (null) in Sources */ = {isa = PBXBuildFile; };
```
nickdowell, line 1003: Broken Xcode project?

nickdowell:
Can we go back to something closer to the original version of this PR?

The latest version (with BSGHardware) adds over 1KB to the binary size whereas the original added 0.

nickdowell:
The naming convention for functions in Bugsnag is `UpperCamelCase`.

`snake_case` should be restricted to code that lives under `/Bugsnag/KSCrash`.

nickdowell (approved):

PR #1397, opened 2022-06-01
[Plat-8106] Remove calls to libc printf

kstenerud:
## Goal

Libc printf acquires locks, and so it not safe for use in async-safe code.

This PR replaces these calls with async-safe alternatives, and also disables low level logging by default so that it doesn't run in production (it doesn't hurt anything, but bloats the binary size, and is of no use in the field).

## Design

Low level number-to-string functions from the JSON encoder have now been exposed so that they can be used from elsewhere.

Added stb_sprintf to replace the few sprintf calls in the logger and other places in the codebase that can't be easily replaced with simpler alternatives.

All low level logging is now dependent upon the define `BSG_KSLOG_ENABLE_LOW_LEVEL_LOGGING`. When 0, no low level logging code is compiled.

## Testing

Added tests for the low level number-to-string functions.
Tested manually with `BSG_KSLOG_ENABLE_LOW_LEVEL_LOGGING` = 1 and = 0.


Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashC.h
```diff
@@ -93,6 +93,8 @@ void bsg_kscrash_reinstall(const char *const crashReportFilePath,
                            const char *const crashID);
 
 /** Set whether or not to print a stack trace to stdout when a crash occurs.
+ *
+ * Note: This has been removed, and now does nothing.
```
kattrali, line 97: Why not remove `bsg_kscrash_setPrintTraceToStdout()` and `config.printTraceToStdout`?

nickdowell, line 97: Can we remove this function altogether?

kstenerud: We could, but I was worried about it breaking someone's build since it's technically a public API.

kattrali: Oh lol Nick beat me to it!

kattrali: I don't think there's any way to import this via public Bugsnag headers (?)

Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSLogger.h
```diff
@@ -24,8 +24,25 @@
...
+
+
 #include "BugsnagLogger.h"
 
+/**
+ * Enables low-level logging.
+ *
+ * Keep this disabled in production as it increases the binary size.
+ */
+#ifndef BSG_KSLOG_ENABLE_LOW_LEVEL_LOGGING
```
nickdowell, line 42: Any reason not to shorten this to `BSG_KSLOG_ENABLED`?

kattrali (approved):

PR #1398, opened 2022-06-01
[PLAT-8506] Local watchOS E2E testing

nickdowell:
## Goal

Make it possible to run E2E scenarios on Apple Watch.

## Scope

Allows running the `barebone_tests` on watchOS.

## Design

Launches the fixture using Xcode 13's `xcdebug` command-line tool.

The `--environment` option does not seem to work, so maze runner's host name needs to be compiled into the fixture. The `Makefile` takes care of this by inserting the computer's IP address.

## Changeset

Adds a watchOS fixture - `watchOSTestApp` in `features/fixtures/watchos`

Amends `features/barebone_tests.feature` to be compatible with watchOS.

`BareboneTestUnhandledErrorScenario` has been changed to an uncaught NSException because it's not possible to detect Swift fatal errors.

## Testing

Run the tests with `make e2e_watchos`

An iPhone with paired Apple Watch must be connected and available in Xcode's devices window.

See [TESTING.md](https://github.com/bugsnag/bugsnag-cocoa/blob/nickdowell/watchos-e2e/TESTING.md#running-tests-on-apple-watch)


features/steps/app_steps.rb
```diff
@@ -10,15 +10,17 @@
...
     run_and_relaunch
   when 'macos'
     $scenario_mode = $last_scenario[:scenario_mode]
     execute_command($last_scenario[:action], $last_scenario[:scenario_name])
+  when 'watchos'
+    run_watchos_app
+    sleep 10 # we don't have a way to check if the app is still running
   else
-    raise "Unknown platform: #{platform}"
+    raise 'Unsupported platform'
```
twometresteve, line 23: Any reason to remove the printing of `platform`?  Couldn't that be a bit annoying for someone if it falls into this branch?

twometresteve (commented):
Test/CI approach looks good to me, but could we have something in TESTING.md detailing how to run the tests on watchOS (and to some extent advertise that it's possible)?

features/steps/app_steps.rb
nickdowell: Seemed unlikely someone wouldn't know which platform they're running tests on... but I can restore this

twometresteve: For me it was the fact that you're at the mercy of what `Maze::Helper.get_current_platform` returns.  It's not necessarily just what you pass in on the command line.

twometresteve (approved):

kstenerud (approved):

PR #1406, opened 2022-06-16
[PLAT-8545] Capture the current thread name on crash

kstenerud:
## Goal

Capturing thread names is async-safe when getting the current thread name. This can be useful when recording Objective-C and C++ exceptions because they are handled on the same thread that threw the exception.

Note: This won't capture the offending thread name for mach exceptions because they are handled on the mach exception handler thread.

## Testing

Added thread name checks for ObjC and C++ exception e2e tests.


features/unhandled_cpp_exception.feature
```diff
@@ -14,6 +14,7 @@ Feature: Thrown C++ exceptions are captured by Bugsnag
     And the event "severity" equals "error"
     And the event "unhandled" is true
     And the event "severityReason.type" equals "unhandledException"
+    And the event "threads.0.name" equals "BSG MAIN THREAD"
```
kattrali, line 17: Might be worth testing different values for the various cases to check that they are sensibly handled? for example, the following are valid:

```c
pthread_setname_np(thread, "");
pthread_setname_np(thread, "\nnext line?");
pthread_setname_np(current, "٥٦٧٨٩");
pthread_setname_np(current, "œ´¨ø“‘");
```

kattrali (approved):

nickdowell (changes requested):
The changes to `bsg_kscrw_i_writeThread` LGTM but I think adding more scenarios to the E2E is a bad idea due to additional run time required. Instead the thread name checks can be conditionalised - e.g.

```
And on iOS 13 and later, the event "threads.0.name" equals "потік"
```

features/steps/reusable_steps.rb
```diff
@@ -10,8 +10,18 @@
...
-Then(/^on (iOS|macOS) (\d+) and later, (.+)/) do |platform, version, step_text|
-  step(step_text) if platform.downcase == Maze::Helper.get_current_platform && Maze.config.os_version >= version
+Then(/^on (\w+) ([0-9.]+) and later, (.+)/) do |test_platform, test_version, step_text|
+  actual_platform = Maze::Helper.get_current_platform
+  actual_version = Maze.config.os_version
+
+  $logger.info "Detected actual platform/version: #{actual_platform} #{actual_version}"
+
+  unless test_platform.downcase == actual_platform && actual_version >= test_version.to_f
+    $logger.info "Skipping #{test_platform} #{test_version} check on #{actual_platform} #{actual_version}"
```
nickdowell, line 20: This logging seems unnecessary, is it debug code accidentally left in? @twometresteve 

```suggestion
  unless test_platform.downcase == actual_platform && actual_version >= test_version.to_f
```

twometresteve: I wanted it to be obvious what the scenario was actually testing and what it wasn't, because Cucumber turns it green whether a check was performed and passed or no check at all was performed.  I was trying to guard against the behaviour of `Maze::Helper.get_current_platform` changing subtly and undermining the entire step quietly.

twometresteve: ...although the one on L17 could be removed.

nickdowell: It just feels very odd to have such a noisy step - generally they produce no output unless there's an error.

Maybe $logger.debug instead ?

Or going back to the `Then(/^on (iOS|macOS) ...` way of doing it ensures only known platform names can be specified.

nickdowell (approved):

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -926,7 +927,17 @@ void bsg_kscrw_i_writeThread(const BSG_KSCrashReportWriter *const writer,
         writer->addBooleanElement(writer, BSG_KSCrashField_Crashed,
                                   isCrashedThread);
         writer->addBooleanElement(writer, BSG_KSCrashField_CurrentThread,
-                                  thread == bsg_ksmachthread_self());
+                                  isSelfThread);
+
+        // Fetching the current thread name is only safe as of libpthread-330.201.1
+        // which was used in ios+tvos 12 and macos 10.14
+        if(__builtin_available(iOS 12.0, tvOS 12.0, macOS 10.14, *)) {
```
nickdowell, line 934: `__builtin_available` is not async-safe...

some more info in this comment:

https://github.com/bugsnag/bugsnag-cocoa/blob/master/Bugsnag/KSCrash/Source/KSCrash/Recording/Tools/BSG_KSMach.c#L66-L69

nickdowell (changes requested):

Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSCrashReport.c
```diff
@@ -926,7 +928,19 @@ void bsg_kscrw_i_writeThread(const BSG_KSCrashReportWriter *const writer,
...
                                   isCrashedThread);
         writer->addBooleanElement(writer, BSG_KSCrashField_CurrentThread,
-                                  thread == bsg_ksmachthread_self());
+                                  isSelfThread);
+
+        // Fetching the current thread name is only safe as of libpthread-330.201.1
+        // which was used in ios+tvos 12 and macos 10.14
+        if (bsgos_available(BSGOS_IOS, 12, 0) ||
+            bsgos_available(BSGOS_TVOS, 12, 0) ||
+            bsgos_available(BSGOS_MACOS, 10, 14)) {
```
nickdowell, line 937: The complexity and binary size impact of the `bsgos_available` function(s) (760 bytes) feels a bit unjustified for this use.

How about a much simpler approach - checking the version of CoreFoundation to deduce the OS version? It was a commonly used approach before introduction of the `@available` intrinsic.

```suggestion
        if (kCFCoreFoundationVersionNumber >= kCFCoreFoundationVersionNumber_iOS_12_0) {
```

Annoyingly, Apple stopped adding defines for the version after iOS 9, so we'll need to declare this ourselves:

```
// Reference: http://iphonedevwiki.net/index.php/CoreFoundation.framework
#define kCFCoreFoundationVersionNumber_iOS_12_0 793.00
```

nickdowell (approved):

PR #1407, opened 2022-06-20
Disable file protection for BSGRunContext

nickdowell:
## Goal

Fix a potential segfault when accessing `bsg_runContext` after device is locked.

If an app uses the [Data Protection Entitlement](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_developer_default-data-protection) to set the default protection to `NSFileProtectionComplete`, the file mapping becomes invalidated once the underlying file gets encrypted - i.e. after locking the device. Dereferencing `bsg_runContext` then causes a segfault and crash the app.

## Changeset

Disables content / file protection of the memory mapped file, using the NSURL / NSURLFileProtection APIs.

Considered using lower-level APIs - `fcntl` with `F_GETPROTECTIONCLASS` and `F_SETPROTECTIONCLASS` is available but the [protection class definitions](https://github.com/apple-oss-distributions/xnu/blob/xnu-8020.121.3/bsd/sys/content_protection.h#L34-L50) are private to xnu and their mappings to corresponding NSURLFileProtectionType undocumented.

<details>

```c
/*
 * Protection classes vary in their restrictions on read/writability.
 * A is generally the strictest, and D is effectively no restriction.
 */

#define PROTECTION_CLASS_A 1 // -> NSFileProtectionComplete
#define PROTECTION_CLASS_B 2 // -> NSFileProtectionCompleteUnlessOpen
#define PROTECTION_CLASS_C 3 // -> NSFileProtectionCompleteUntilFirstUserAuthentication
#define PROTECTION_CLASS_D 4 // -> NSFileProtectionNone
#define PROTECTION_CLASS_E 5
#define PROTECTION_CLASS_F 6
```
</details>

## Testing

Reproduced the crash locally by adding the Data Protection Entitlement and a UIBackgroundTask to a sample app. After locking the device, the app would crash upon completion of the background task (it appears that content won't get protected while a UIBackgroundTask is active.)

Was unable to reproduce the crash with this fix in place.


Bugsnag/Helpers/BSGRunContext.m
```diff
@@ -378,9 +378,19 @@ static void LoadLastRunContext(int fd) {
 
 /// Truncates or extends the file to the size of struct BSGRunContext,
 /// maps it into memory, and sets the `bsg_runContext` pointer.
-static void ResizeAndMapFile(int fd) {
+static void ResizeAndMapFile(const char *path, int fd) {
```
kstenerud, line 381: This feels weird passing in the path of a file that's already been opened. What about a function `openWithProtectionDisabled`, which takes a path and returns an fd?

PR #1416, opened 2022-06-27
Ignore OOMs that occur while the app is inactive

nickdowell:
## Goal

Ignore OOMs that occur in `UIApplicationStateInactive`, restoring behaviour previously introduced by #394.

Pull request #1307 (in v6.16.4) inadvertently stopped rejecting these OOMs.

## Changeset

Adds `BSGRunContext.inActive`.
Observes `UIApplicationDidBecomeActiveNotification` / `WillResignActiveNotification`.
Checks `isActive` in `BSGRunContextWasKilled()`.

Enables state tracking on tvOS, since OOMs are also reported there.

## Testing

Adds an E2E scenario that simulates an OOM occurring in `UIApplicationStateInactive` and verifies that no error is reported.

### DIFF
--- a/Bugsnag/Helpers/BSGRunContext.m
+++ b/Bugsnag/Helpers/BSGRunContext.m
@@ -38,7 +38,11 @@
 #pragma mark Forward declarations
 
 static uint64_t GetBootTime(void);
+static bool GetIsActive(void);
 static bool GetIsForeground(void);
+#if TARGET_OS_IOS || TARGET_OS_TV
+static UIApplication * GetUIApplication(void);
+#endif
 static void InstallTimer(void);
 static void UpdateAvailableMemory(void);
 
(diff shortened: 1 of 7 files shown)


Bugsnag/Helpers/BSGRunContext.m
```diff
@@ -66,6 +70,14 @@ static void InitRunContext() {
         uuid_copy(bsg_runContext->machoUUID, image->uuid);
     }
     
+    if ([NSThread isMainThread]) {
+        bsg_runContext->isActive = GetIsActive();
+    } else {
+        dispatch_async(dispatch_get_main_queue(), ^{
```
kattrali, line 76:
```suggestion
        dispatch_sync(dispatch_get_main_queue(), ^{
```

should this block to wait for the context to be updated, or was the intent to update the context async (in which case more should move into the block (like updating the timestamp?)

nickdowell: This is intentionally async, didn't want to block the calling thread.

The check is there because `GetIsActive()` must be called on the main thread.

nickdowell: True, the timestamp could be a little behind, but I don't think it matters 🙃

kattrali (approved):

PR #1417, opened 2022-06-28
Remove unused static variables from BugsnagClient

nickdowell:
Removes some unused variables from BugsnagClient.

Sorts import statements.


kattrali (approved):
nice!

PR #1418, opened 2022-06-28
Run jazzy without bundler

nickdowell:
## Goal

Fix the [Update Docs](https://github.com/bugsnag/bugsnag-cocoa/actions/workflows/update_docs.yml) action that started failing with release 6.18.0 (apparently due to a [change](https://github.com/bugsnag/bugsnag-cocoa/commit/48c94b94778e5306a22bd6fbe704ceed86e4db27#diff-89cade48462044ee1b672dc5f4c3ec250fbd29effcd8932096a23c1283c6731fR212) in `Gemfile.lock`)

## Changeset

Runs `jazzy` without Bundler.

Removes `jazzy` from `Gemfile` (it's pre-installed on GitHub's macOS images.)

## Testing

See successful dry run: https://github.com/bugsnag/bugsnag-cocoa/runs/7095074190?check_suite_focus=true


kattrali (commented):
how odd - is the jazzy removal actually necessary, or does it the version of bundler or jazzy or something else need to be pinned?

kattrali:
I'm not opposed to removing bundler (generally the opposite), though it would make sense to still cache the gem to avoid extra work. This is still possible by caching the result of calling `gem install` (inside $GEM_HOME), but slightly more effort to grab the right paths than using bundler.

There's also the potential of future breakages between the ruby version used and the latest version of jazzy, which is harder to track without a Gemfile.lock pinning versions.

In this particular case it seems like preventing arbitrary updates to the bundler version is ideal.

nickdowell:
Now uses the pre-installed version of `jazzy` provided by Github

twometresteve (approved):

PR #1423, opened 2022-07-05
[Plat-5196] Fix init order

kstenerud:
## Goal

Some parts of the library were being used before initialization.

## Design

Make sure the init functions are called before anything uses them. Also, don't allow arbitrary configs to be passed in to the lower levels when there's always a single global instance in reality.

## Testing

Reran all tests.


Bugsnag/Client/BugsnagClient.m
```diff
@@ -165,7 +165,8 @@ - (instancetype)initWithConfiguration:(BugsnagConfiguration *)configuration {
         if (!_configuration.user.id) { // populate with an autogenerated ID if no value set
             [_configuration setUser:[BSG_KSSystemInfo deviceAndAppHash] withEmail:_configuration.user.email andName:_configuration.user.name];
         }
-        
+        [_configuration validate];
+
```
nickdowell, line 169: This change is causing the unit tests to fail.

By design, `validate` is called from `start` so that an exception won't be thrown too early.

nickdowell (approved):

PR #1425, opened 2022-07-06
[PLAT-6059] Fix empty traces in some edge cases

kstenerud:
## Goal

The code was incorrectly returning 0 stack trace entries in some error edge cases when it should have been returning the current count. This is a rare edge case, but it may have been contributing to our CI flakes.

## Testing

Reran all tests.


nickdowell (approved):
This does appear to fix the issue!

https://buildkite.com/bugsnag/bugsnag-cocoa/builds/5920

PR #1428, opened 2022-07-07
[PLAT-8619] Capture userInfo from NSExceptions

nickdowell:
## Goal

Capture `userInfo` from Objective-C exceptions, both uncaught and those provided to `notify()`.

NSExceptions raised by Apple's frameworks don't usually have a userInfo, but those raised in user or 3rd party code might.

## Changeset

Overhauls `BugsnagClient` `notify:` methods to reduce code duplication, make implementation easier to follow, and capture `userInfo` from exceptions.

Adds metadata to handled exceptions in a structure that matches what is already reported for unhandled ones:

```json
"metaData": {
  "error": {
    "nsexception": {
      "name": "NSRangeException",
      "userInfo": {
        "NSUnderlyingError": "Error Domain=ErrorDomain Code=0 \"(null)\""
      }
    },
    "reason": "*** -[__NSArray0 objectAtIndex:]: index 42 beyond bounds for empty NSArray",
    "type": "nsexception"
  },
}
```

Updates `BSG_KSCrashSentry_NSException.m` to capture the `userInfo`, and `BSG_KSCrashReport.c` to include it in crash reports.

Removes `BugsnagEvent.error` and updates `-[BugsnagEvent initWithKSCrashReport:]` to explicitly set the `error` metadata from crash reports.

## Testing

Amends barebones test scenarios to verify `context` and `metaData.error`.


nickdowell:
Metadata structure changes have been reverted to avoid breaking customers' existing filters.

PR #1440, opened 2022-07-19
E2E fixes

nickdowell:
## Goal

Fix E2E flakes

## Changeset

* Disables `MallocScribble` etc. on macOS 10.13 because CFNetwork was intermittently crashing with it enabled.
* Ignore background app hangs in `CABackingStoreCollectBlocking` - this was intermittently causing two hangs to be reported resulting in failure of that scenario.

## Testing

Passed full CI run - https://buildkite.com/bugsnag/bugsnag-cocoa/builds/5999

### DIFF
--- a/features/fixtures/shared/scenarios/ReportBackgroundAppHangScenario.swift
+++ b/features/fixtures/shared/scenarios/ReportBackgroundAppHangScenario.swift
@@ -5,6 +5,12 @@ class ReportBackgroundAppHangScenario: Scenario {
     override func startBugsnag() {
         self.config.appHangThresholdMillis = 1_000
         self.config.reportBackgroundAppHangs = true
+        self.config.addOnSendError { event in
+            !event.errors[0].stacktrace.contains { stackframe in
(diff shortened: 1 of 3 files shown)


features/fixtures/shared/scenarios/ReportBackgroundAppHangScenario.swift
```diff
@@ -5,6 +5,12 @@ class ReportBackgroundAppHangScenario: Scenario {
     override func startBugsnag() {
         self.config.appHangThresholdMillis = 1_000
         self.config.reportBackgroundAppHangs = true
+        self.config.addOnSendError { event in
+            !event.errors[0].stacktrace.contains { stackframe in
+                // CABackingStoreCollectBlocking is known to hang for several seconds upon entering the background
+                stackframe.method == "CABackingStoreCollectBlocking"
```
kstenerud, line 11: If `CABackingStoreCollectBlocking` is present, won't it omit the event that maze runner is waiting for?

kstenerud (approved):

PR #1442, opened 2022-07-20
[PLAT-8726] Use `device.id` as fallback `user.id` for events & sessions

nickdowell:
## Goal

Makes behaviour of `user.id` consistent with other platform, crash errors, and our [online documentation](https://docs.bugsnag.com/platforms/ios/automatically-captured-data/#user-information).

After calling `Bugsnag.setUser(nil, withEmail: nil, andName: nil)`, events and sessions should have their `user.id` set to the default Bugsnag-generated device ID.

## Changeset

Adds `-[BugsnagUser withUserId]` to get a user object with the fallback `id` if required.

Updates `BugsnagSessionTracker` and `BugsnagClient` to use this method to populate sessions and events with a `user.id`.

Renames `-[BugsnagUser initWithUserId:name:emailAddress:]` to `-[BugsnagUser initWithId:name:emailAddress:]`.

Removes unused method `+[BugsnagUser userWithUserId:name:emailAddress:]`.

Simplifies `-[BugsnagConfiguration getPersistedUserData]`.

## Testing

Amends existing and adds a new scenario to verify that fallback `user.id` is sent when expected, while retaining ability to nullify it via callbacks.

### DIFF
--- a/Bugsnag/Payload/BugsnagUser.m
+++ b/Bugsnag/Payload/BugsnagUser.m
@@ -8,7 +8,7 @@
 
 #import "BugsnagUser+Private.h"
 
-#import "BugsnagCollections.h"
+#import "BSG_KSSystemInfo.h"
 
 @implementation BugsnagUser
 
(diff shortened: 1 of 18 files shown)


Bugsnag/Payload/BugsnagUser.m
```diff
@@ -43,4 +38,14 @@ - (NSDictionary *)toJson {
...
 
+- (BugsnagUser *)withId {
+    if (self.id) {
+        return self;
+    } else {
+        return [[BugsnagUser alloc] initWithId:[BSG_KSSystemInfo deviceAndAppHash]
+                                          name:self.name
+                                  emailAddress:self.email];
+    }
+}
```
kstenerud, line 49: Why can't we just set this in the constructor when nil is passed as the user ID?

nickdowell: I've left it this way to be consistent with the Android notifier where after calling `Bugsnag.setUser(null, null, null)`, `Bugsnag.getUser().id` will be null but events & sessions will have the fallback value.

It's probably not _that_ important a behaviour, but I thought consistency would be best.

kstenerud (approved):

PR #1460, opened 2022-08-12
[PLAT-4594] Remove unnecessary use of @synchronized

nickdowell:
## Goal

Remove unnecessary `@synchronized` statements to improve performance and reduce code size.

## Changeset

Removes `@synchronized` from configuration property accessors. These properties are declared `nonatomic` and not expected to be called concurrently, so synchronization is not required.

Removes `@synchronized` from `BugsnagEvent.severity` accessors. Events are only accessible from a single thread (when passed to an `OnError` / `OnSendError` block), so again synchronization is not required.

Adds `@synchronized` to `-[BugsnagClient user]` and `-[BugsnagClient setUser:withEmail:andName:]` to ensure state will be consistent after concurrent calls.

Refactors user persistence for simplicity and to remove complexity from `BugsnagConfiguration`.

## Testing

Covered by existing unit and E2E tests.

Removes `NSUserDefaultsStub` and associated code. This was originally added in #949 to fix unit test flakes. Further investigation has revealed the root cause to be stale data in the simulator(s). After erasing simulator contents and settings the tests pass.


kattrali (approved):
A lot has changed architecturally in the last several years, its nice to see these things tidied up. :+1:

PR #1461, opened 2022-08-15
[PLAT-8797] Fix os version conditional step on watchOS

nickdowell:
## Goal

Fix E2E tests on watchOS.

## Changeset

Fixes the conditional `on (iOS|macOS|watchOS), (.+)` step.

Refactors platform check into `platform_matches?` function.

## Testing

Verified with locally connected Apple watch.

### DIFF
--- a/features/steps/reusable_steps.rb
+++ b/features/steps/reusable_steps.rb
@@ -6,24 +6,20 @@ When('I ignore invalid {word}') do |type|
   Maze.config.captured_invalid_requests.delete(type.to_sym)
 end
 
-Then(/^on (iOS|macOS|watchOS), (.+)/) do |platform, step_text|
-  step(step_text) if platform.downcase == Maze::Helper.get_current_platform
+def platform_matches?(platform)
+  Maze::Helper.get_current_platform == platform.downcase
 end
 
-Then(/^on (\w+) ([0-9.]+) and later, (.+)/) do |test_platform, test_version, step_text|
-  actual_platform = Maze::Helper.get_current_platform
-  unless %w[ios macos].include? actual_platform
-    raise "Unexpected platform #{actual_platform}, this step can only handle ios and macos"
-  end
+Then(/^on (iOS|macOS|watchOS), (.+)/) do |platform, step_text|
+  step(step_text) if platform_matches?(platform)
+end
 
-  actual_version = Maze.config.os_version
-  next unless test_platform.downcase == actual_platform && actual_version >= test_version.to_f
-
-  step(step_text)
+Then(/^on (iOS|macOS|watchOS) ([0-9.]+) and later, (.+)/) do |platform, version, step_text|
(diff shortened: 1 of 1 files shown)


features/steps/reusable_steps.rb
```diff
@@ -6,24 +6,20 @@
...
-
-  actual_version = Maze.config.os_version
-  next unless test_platform.downcase == actual_platform && actual_version >= test_version.to_f
+Then(/^on (iOS|macOS|watchOS), (.+)/) do |platform, step_text|
+  step(step_text) if platform_matches?(platform)
+end
 
-  step(step_text)
+Then(/^on (iOS|macOS|watchOS) ([0-9.]+) and later, (.+)/) do |platform, version, step_text|
+  step(step_text) if platform_matches?(platform) && version >= Maze.config.os_version.to_s
```
twometresteve, line 18: I don't think you want to be doing string comparison for the versions:
```
irb(main):007:0> a = "10"
=> "10"
irb(main):008:0> b = "9"
=> "9"
irb(main):009:0> a >= b
=> false
irb(main):010:0> b >= a
=> true
```

features/steps/reusable_steps.rb
```diff
@@ -6,24 +6,20 @@
...
 
-Then(/^on (iOS|macOS|watchOS), (.+)/) do |platform, step_text|
-  step(step_text) if platform.downcase == Maze::Helper.get_current_platform
+def platform_matches?(platform)
+  Maze::Helper.get_current_platform == platform.downcase
 end
 
-Then(/^on (\w+) ([0-9.]+) and later, (.+)/) do |test_platform, test_version, step_text|
-  actual_platform = Maze::Helper.get_current_platform
-  unless %w[ios macos].include? actual_platform
```
twometresteve, line 15: So was the problem here that it didn't account for watchOS?  Could we not have just allowed for that without refactoring it all, introducing a new problem?

features/steps/reusable_steps.rb
nickdowell: Fixed 👍

features/steps/reusable_steps.rb
nickdowell: This code was checking the wrong thing, there is no need for the step to verify the return value from `Maze::Helper.get_current_platform`.

This step is now consistent with the other OS-conditional steps.

twometresteve (approved):

features/steps/reusable_steps.rb
nickdowell: Also I think using floats to represent OS versions is a bug - there are often more than 2 components to a version.

twometresteve: Yeah true, although the whole mechanism is broken in a sense, when the OS version comes from a command line argument.  However, if what's here is working for you then that's all good.

twometresteve (approved):

PR #1466, opened 2022-08-17
Fix watchOS E2E when run without `--os-version`

nickdowell:
## Goal

Fix watchOS E2E when run without `--os-version`

The [Makefile](https://github.com/bugsnag/bugsnag-cocoa/blob/b6b9fd27b1d5e78aadeb226dbec6db3e53bdf3f8/Makefile#L147-L151) is unable to pass `--os-version` because we don't have a way to determine the OS version of the connected watch.

## Changeset

Skips version check unless platform matches.

Changes step definition to clarify that watchOS is not a supported target platform.

## Testing

Verified by running `make e2e_watchos`.

### DIFF
--- a/features/steps/reusable_steps.rb
+++ b/features/steps/reusable_steps.rb
@@ -14,9 +14,10 @@ Then(/^on (iOS|macOS|watchOS), (.+)/) do |platform, step_text|
   step(step_text) if platform_matches?(platform)
 end
 
-Then(/^on (iOS|macOS|watchOS) ([0-9.]+) and later, (.+)/) do |platform, version, step_text|
+Then(/^on (iOS|macOS) ([0-9.]+) and later, (.+)/) do |platform, version, step_text|
+  next unless platform_matches?(platform)
   raise 'Maze.config.os_version is not defined' if Maze.config.os_version.nil?
-  step(step_text) if platform_matches?(platform) && Gem::Version.new(Maze.config.os_version) >= Gem::Version.new(version)
+  step(step_text) if Gem::Version.new(Maze.config.os_version) >= Gem::Version.new(version)
 end
 
 Then(/^on !(iOS|macOS|watchOS), (.+)/) do |platform, step_text|


features/steps/reusable_steps.rb
```diff
@@ -15,8 +15,9 @@ def platform_matches?(platform)
 end
 
 Then(/^on (iOS|macOS|watchOS) ([0-9.]+) and later, (.+)/) do |platform, version, step_text|
+  next unless platform_matches?(platform)
```
twometresteve, line 18: I'm confused as to how this works and what is being tested (which has always concerned me when Cucumber turns a step green even if it does absolutely nothing).

When running `make e2e_watchos`, it looks like `Maze.config.os` should be `watchos`, so why doesn't `platform_matches?` return true here and it barf on the next line?

In general I remain nervous about this step, as a subtle change in the behaviour of `Maze::Helper.get_current_platform` would lead to this step quietly doing nothing yet reporting a pass state. That routine derives the value from Appium capabilities when running on a device farm, which is potentially out of our control.  A change in value  I would be happier if `platform_matches?` explicitly checked that the value returned from `Maze::Helper.get_current_platform` was in a known/acceptable list of values, failing loudly if it was not.

Having said that.... perhaps I am making an argument here for Maze Runner to be making that check itself?

nickdowell: Ah yes - it would barf if a scenario included a step like `on watchOS 8 and later, the event ...`

Right now we have no need to do that, so I didn't think it was a problem that needs solving. We'd have to figure out a way to determine the watchOS version.

As for checking that `Maze::Helper.get_current_platform` is one of the expected values - there are many other places in `bugsnag-cocoa/features` that make this assumption. If this assumption were broken, the E2E run would likely not succeed.

As you suggest, I think Maze Runner should be making that check itself!

nickdowell: Changed `Then(/^on (iOS|macOS|watchOS) ([0-9.]+) and later, (.+)/)`
to `Then(/^on (iOS|macOS) ([0-9.]+) and later, (.+)/)`

twometresteve (approved):

PR #1468, opened 2022-08-18
E2E: Disable unexpectedly quit dialog on macOS 11

nickdowell:
## Goal

Fix intermittent E2E `did not GET /command` flakes on macOS 11.

These were caused by the "macOSTestApp quit unexpectedly" dialog stealing focus from the fixture. Becoming foreground is what triggers the fixture to GET /command.

## Changeset

Unloads the `com.apple.ReportCrash` service while running tests on macOS 11. It was found through manual experimentation that the usual preferences (configurable through `CrashReporterPrefs.app`) have no effect on this OS version.

On other versions the `UseUNC` option is still used because it is sometimes useful to inspect Apple's crash reports.

## Testing

Verified locally and through a [dedicated CI run](https://buildkite.com/bugsnag/bugsnag-cocoa/builds/6174#0182b048-6033-4219-9300-70244ae7bce4).

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -52,29 +52,19 @@ BeforeAll do
   end
 end
 
-AfterAll do
-  $after_all_blocks.each(&:call) unless $after_all_blocks.nil?
-end
-
 # Disables the "macOSTestApp quit unexpectedly" dialog to prevent focus being stolen from the fixture.
 def disable_unexpectedly_quit_dialog
   if Maze.config.os_version.floor == 11
     # com.apple.CrashReporter defaults seem to be ignored on macOS 11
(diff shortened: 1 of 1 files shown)


features/support/env.rb
```diff
@@ -55,6 +52,31 @@
   end
 end
 
+AfterAll do
```
twometresteve, line 55: You might want to use `at_exit` (a Ruby rather than Cucumber thing) instead, as it should still get called in a Ctrl-C or Buildkite agent termination situation.

twometresteve (approved):
LGTM - would it be worth me adding a generalization of this to Maze Runner so that, eg Unity, can make use of it?

features/support/env.rb
nickdowell: Nice 👍

PR #1471, opened 2022-08-24
Simplify running E2E tests on macOS

nickdowell:
## Goal

Make it easier to run maze-runner when targeting macOS.

## Changeset

Moves OS version detection to `env.rb`, removing need to pass `--os-version` on command line.

Stops using `--app` command line argument (`Maze.config.app`) - passing anything other than `macOSTestApp` has never been supported.

## Testing

Tested by running locally.

### DIFF
--- a/.buildkite/pipeline.full.yml
+++ b/.buildkite/pipeline.full.yml
@@ -257,8 +257,6 @@ steps:
       - bundle install
       - bundle exec maze-runner
         --os=macos
-        --os-version=11.0
-        --app=macOSTestApp
         --fail-fast
         --order=random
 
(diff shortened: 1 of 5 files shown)


twometresteve (approved):
LGTM, only thing I wondered is if you wanted to run the e2e tests on macOS 11, as I don't think that is covered by basic CI and from memory there is some 11-specific test harness logic somewhere?

nickdowell:
Updated the BuildKite pipelines to omit the --app and --os-version options, and ran the full suite to ensure macOS 11 is still ok.

twometresteve (approved):

PR #1474, opened 2022-09-01
[PLAT-8863] Fix crash when reporting internal errors during `+[Bugsnag start]`

nickdowell:
## Goal

Fix a crash that could occur when reporting internal errors during `+[Bugsnag start]`.

## Changeset

Initializes `BSGInternalErrorReporter` AFTER `bsg_runContext` to ensure the latter is non-null.

Considered moving `BSGRunContextInit` to a `+initialize` method instead, but that would hinder testing - it would not be possible to set `bsg_runContext` to `NULL` and have it re-initialized when starting a new `BugsnagClient`.

## Testing

Adds a unit test case that reproduced the problem and verifies the fix.

### DIFF
--- a/Bugsnag/Client/BugsnagClient.m
+++ b/Bugsnag/Client/BugsnagClient.m
@@ -214,18 +214,15 @@ __attribute__((annotate("oclint:suppress[too many methods]")))
 }
 
 - (void)start {
-    if (self.configuration.telemetry & BSGTelemetryInternalErrors) {
-        BSGInternalErrorReporter.sharedInstance = [[BSGInternalErrorReporter alloc] initWithDataSource:self];
-    } else {
-        bsg_log_debug(@"Internal error reporting was disable in config");
-    }
-
     // Called here instead of in init so that a bad config will only throw an exception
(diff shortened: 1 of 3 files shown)


Bugsnag/Client/BugsnagClient.m
```diff
@@ -235,6 +232,12 @@ - (void)start {
 
     [self computeDidCrashLastLaunch];
 
+    if (self.configuration.telemetry & BSGTelemetryInternalErrors) {
+        BSGInternalErrorReporter.sharedInstance = [[BSGInternalErrorReporter alloc] initWithDataSource:self];
+    } else {
+        bsg_log_debug(@"Internal error reporting was disable in config");
```
tomlongridge, line 238: Minor (pre-existing) typo, probably not worth a CI re-run unless you have other changes:

```suggestion
        bsg_log_debug(@"Internal error reporting was disabled in config");
```

nickdowell: Will fix in the release 👍

PR #1476, opened 2022-09-01
[PLAT-8849] Add `BugsnagInternals.h`

nickdowell:
## Goal

Make it easier for dependent platform notifiers to integrate with bugsnag-cocoa and clarify which APIs are used by them to aid refactoring.

## Changeset

Moves all interfaces used by dependent notifiers into `BugsnagInternals.h`, and removes their existing declarations from `+Private.h` files.

Makes `BugsnagClient.breadcrumbs` an array of breadcrumbs rather than exposing the `BugsnagBreadcrumbs` class.

Adds `BugsnagClient.session` and `-[BugsnagClient registerSession:...]` to avoid exposing the `BugsnagSessionTracker` class.

Removes the `-[BugsnagClient collect...]` methods - `@bugsnag/react-native` has now been updated to call the same methods as other notifiers.

Removes some unnecessary and unused APIs from the `Bugsnag` class - dependents should instead directly access the `BugsnagClient` instance.

## Testing

Tested with dependent notifiers:
* https://github.com/bugsnag/bugsnag-cocos2dx/pull/22
* https://github.com/bugsnag/bugsnag-flutter/pull/168
* https://github.com/bugsnag/bugsnag-js/pull/1810
* https://github.com/bugsnag/bugsnag-unity/pull/608
* https://github.com/bugsnag/bugsnag-unreal/pull/184


Bugsnag/BugsnagSessionTracker.m
```diff
@@ -163,19 +164,30 @@ - (void)startNewSession {
...
         return;
     }
+    [self registerSession:[[NSUUID UUID] UUIDString] startedAt:[NSDate date] handledCount:0 unhandledCount:0];
+}
 
+- (void)registerSession:(NSString *)sessionId startedAt:(NSDate *)startedAt handledCount:(NSUInteger)handledCount unhandledCount:(NSUInteger)unhandledCount {
+    if (!sessionId || !startedAt) {
+        self.currentSession = nil;
+        BSGSessionUpdateRunContext(nil);
+        return;
```
kstenerud, line 174: Why is this behaviour needed? It feels weird to clear data when a null is passed in one of two params...

kstenerud (approved):

Bugsnag/BugsnagSessionTracker.m
nickdowell: This is to continue supporting how bugsnag-unity calls the API

kstenerud: Could we get a quick comment about it so that it doesn't become Chesterson's fence? :)

nickdowell: After chatting to @rich-bugsnag about what bugsnag-unity uses this API for, I've tried to clean this up:

* https://github.com/bugsnag/bugsnag-cocoa/blob/6775064f61694be5badf55f6b6d25894319b337e/Bugsnag/Client/BugsnagClient.m#L449-L452
* https://github.com/bugsnag/bugsnag-unity/blob/634751779ce8ba4605c13c9796a894c504cccadc/src/BugsnagUnity.m#L365-L372

kstenerud: 👍

kstenerud (approved):

PR #1480, opened 2022-09-13
Use Appium to get iOS E2E logs

nickdowell:
## Goal

Make it easier to investigate test failures by including device syslog in `maze_output`.

## Changeset

Uses Appium's [get_log](https://appium.io/docs/en/commands/session/logs/get-log/) method to fetch the device log for each scenario.

@twometresteve once proven, this probably should find its way into maze-runner.
On Android, I think `get_log('logcat')` should work.

## Testing

Verified logs are populated when running locally.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -91,20 +91,11 @@ Maze.hooks.before do |_scenario|
   Maze.config.captured_invalid_requests = Set[:errors, :sessions, :builds, :uploads, :sourcemaps]
 
   $started_at = Time.now
-
-  if Maze.config.os == 'ios' && Maze.config.farm == :local
-    begin
-      $logger_pid = Process.spawn(
-        'idevicesyslog', '-u', Maze.config.device_id, '-m', 'iOSTestApp',
-        %i[err out] => File.open('device.log', 'w')
-      )
-    rescue Errno::ENOENT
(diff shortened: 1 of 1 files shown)


twometresteve (commented):
Were you expecting at least one of the pipeline steps to upload files as buildkite artifacts?

nickdowell:
> Were you expecting at least one of the pipeline steps to upload files as buildkite artifacts?

Only failed scenarios have their logs uploaded, so I wasn't expecting to see any in a successful build.

twometresteve (approved):
Nice

features/support/env.rb
```diff
@@ -130,17 +119,17 @@ def skip_below(os, version)
       FileUtils.mv '/tmp/kscrash.log', path
     end
   when 'ios'
-    if $logger_pid
-      Process.kill 'TERM', $logger_pid
-      Process.waitpid $logger_pid
-      $logger_pid = nil
-      FileUtils.mv 'device.log', path
+    # get_log can be slow (1 or 2 seconds) on device farms
+    if scenario.failed || Maze.config.farm == :local
```
twometresteve, line 123:
```suggestion
    if scenario.failed? || Maze.config.farm == :local
```

twometresteve (approved):

PR #1503, opened 2022-11-09
Relax multiplatform stacktrace checks

Cawllec:
## Goal

We have some potential issues with MacOS tests running on Intel and ARM architectures, where stacktraces can differ in small ways.

This change adds a new test step that verifies that a specific method appears in the stacktrace, without worrying about what specific stackframe it's in.

Note - I've kept the check validating the entire stack method, as the makes it more explicit which frame methods are reliable on each platform, even if the check could be relaxed to just ensuring a specific string is present.

### DIFF
--- a/features/crashprobe.feature
+++ b/features/crashprobe.feature
@@ -139,7 +139,7 @@ Feature: Reporting crash events
     And the exception "message" matches "Attempted to dereference (garbage|null) pointer"
     And the exception "errorClass" equals "EXC_BAD_ACCESS"
     And the "method" of stack frame 0 equals "objc_msgSend"
-    And the "method" of stack frame 1 equals one of:
+    And the stacktrace contains one of the following methods:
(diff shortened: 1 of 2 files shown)


twometresteve (approved):

features/crashprobe.feature
```diff
@@ -139,7 +139,7 @@ Feature: Reporting crash events
     And the exception "message" matches "Attempted to dereference (garbage|null) pointer"
     And the exception "errorClass" equals "EXC_BAD_ACCESS"
     And the "method" of stack frame 0 equals "objc_msgSend"
-    And the "method" of stack frame 1 equals one of:
+    And the stacktrace contains one of the following methods:
       | ARM   | __29-[ReleasedObjectScenario run]_block_invoke |
```
nickdowell, line 143: This line could have been removed

Cawllec: You're absolutely right, I'll get a second PR up to remove it.  Apologies for jumping the gun!

Issue #1511, opened 2022-12-15
Calling NSProcessInfo.thermalState from an app with bugsnag installed crashes with _os_unfair_lock_recursive_abort


ericcj:
Apparently calling NSProcessInfo.thermalState can trigger NSProcessInfoThermalStateDidChangeNotification so the NoteThermalState receiver needs to dispatch its call to access NSProcessInfo.thermalState on the object sent to it or it can trigger recursion that crashes the app.  This appears to only happen ~10% of the time but we're not certain: 

<img width="751" alt="Screen Shot 2022-12-15 at 9 19 59 AM" src="https://user-images.githubusercontent.com/53106/207884585-22687d58-29ef-4f64-8d23-ac7db3d65c55.png">

bcherry:
Note that all cases of this that we have seen so far are on iOS 15.1.0

johnkiely1:
Hi, thanks for raising, we're going to take a look into addressing this as soon as we can.

johnkiely1:
This should now be resolved in [v6.25.2](https://github.com/bugsnag/bugsnag-cocoa/releases/tag/v6.25.2).

PR #1514, opened 2023-01-06
Plat 9367 thermal state workaround

kstenerud:
## Goal

A bug from iOS 15.0.2 to 15.1.1 causes `ThermalStateDidChangeNotification` to on rare occasions be posted from within `-[NSProcessInfo thermalState]`, which triggers a recursive call for any listeners since they will in turn call `-[NSProcessInfo thermalState]` and trigger recursion detection in `_os_unfair_lock` (`_os_unfair_lock_recursive_abort`).

See https://github.com/bugsnag/bugsnag-cocoa/issues/1511

Ref: https://github.com/Tencent/matrix/blob/master/matrix/matrix-iOS/Matrix/WCCrashBlockMonitor/CrashBlockPlugin/Main/BlockMonitor/WCBlockMonitorMgr.mm#L493

## Design

This PR defers the `-[NSProcessInfo thermalState]` call to a background thread as a workaround.

## Testing

Manually tested

### DIFF
--- a/Bugsnag/Helpers/BSGRunContext.m
+++ b/Bugsnag/Helpers/BSGRunContext.m
@@ -258,11 +258,16 @@ static void NoteThermalState(__unused CFNotificationCenterRef center,
                              __unused CFNotificationName name,
                              const void *object,
                              __unused CFDictionaryRef userInfo) {
-#pragma clang diagnostic push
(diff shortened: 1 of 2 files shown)


lemnik (approved):
LGTM

PR #1515, opened 2023-01-10
[PLAT-9362] fix for data race and nullability issues

kstenerud:
## Goal

PLAT-9362

Fixed a data race when getting the host memory reading.

There were also warnings about incorrect nullability invariant expectations in the code, which are also now fixed.

## Testing

Tested manually, reran all unit and e2e tests.


lemnik (approved):
LGTM

PR #1519, opened 2023-01-23
Refactor test fixture to read Maze Runner address from disk

twometresteve:
## Goal

Avoid hard-coding the Maze Runner address so that we can use device farms other than BrowserStack.

## Design

iOS 12 and above will be moving to BitBar, where it will need to dynamically determine the Maze Runner address by reading it from a file - `fixture_config.json`, which is push onto the device by Maze Runner (using Appium).

## Changeset

Some light refactoring to allow the passing of the base Maze Runner address in at the appropriate point.  I've also been able to make one or two small improvements - as noted by my own comments.

## Testing

Covered by a full CI run on BrowserStack.  Testing on BitBar will follow in a separate PR.


features/fixtures/shared/scenarios/NetworkBreadcrumbsScenario.swift
```diff
@@ -10,18 +10,12 @@ import Foundation
...
 @available(iOS 10.0, macOS 10.12, *)
 class NetworkBreadcrumbsScenario : Scenario {
-    
-    lazy var baseURL: URL = {
-        var components = URLComponents(string: Scenario.mazeRunnerURL.absoluteString)!
-        components.port = 9340 // `/reflect` listens on a different port :-((
-        return components.url!
-    }()
-    
+
```
twometresteve, line 13: I added the `/reflect` endpoint to the server that runs on 9339 to prevent us from having to do this.

features/support/env.rb
```diff
@@ -121,12 +121,6 @@ def skip_below(os, version)
       FileUtils.mv '/tmp/kscrash.log', path
     end
   when 'ios'
-    # get_log can be slow (1 or 2 seconds) on device farms
```
twometresteve, line 124: Removed as Maze Runner now also does this and there is the suspicion that they interfere with each other.

features/fixtures/ios/iOSTestApp/ViewController.swift
```diff
@@ -46,6 +51,43 @@ class ViewController: UIViewController {
         Scenario.clearPersistentData()
     }
 
+    func loadMazeRunnerAddress() -> String {
+
+        let bsAddress = "http://bs-local.com:9339"
+        
+        // Only iOS 12 and above will run on BitBat for now
```
kstenerud, line 58:
```suggestion
        // Only iOS 12 and above will run on BitBar for now
```

features/fixtures/macos/macOSTestApp/MainWindowController.m
```diff
@@ -92,6 +92,7 @@ - (IBAction)useDashboardEndpoints:(id)sender {
 }
 
 - (IBAction)executeMazeRunnerCommand:(id)sender {
+    Scenario.baseMazeAddress = @"http://localhost:9339";
```
kstenerud, line 95: So on mac we never load the address from disk?

kstenerud (approved):

features/fixtures/macos/macOSTestApp/MainWindowController.m
twometresteve: Yes, that's right.  Currently (and I think always) we only run the macOS e2e tests on the same machine that Maze Runner is running on, so the address is always localhost.

features/fixtures/ios/iOSTestApp/ViewController.swift
twometresteve: I make this typo more frequently than I care to mention!  I've corrected it on my next branch (following on immediate) to avoid an extra CI run.

kstenerud: 💿🦇

PR #1521, opened 2023-02-06
Prepare pipeline for running e2e tests on BitBar

twometresteve:
## Goal

Prepare for running e2e tests on BitBar.

## Design

There are two side to this PR:
- Make it easier to run e2e tests only on BitBar (allowing us to set up a regular schedule)
- Fix or skip failing tests

The first point is achieved by moving all BrowserStack steps from the basic pipeline into a separate file.  By default the CI pipeline will behave in exactly the same way and run tests on BrowserStack, but by either setting a pipeline environment variable or specific commit message, we can chose to run tests on BitBar.  (Either just BitBar or both BitBar and BrowserStack).

A few test scenarios fail on BitBar and require further work/investigation for which I have raised separate tickets.  Some failures I was able to deal with under this PR by changing the way we run some checks.

## Testing

BitBar-based testing covered in the run-up to creating this PR by setting the commit message.  BrowserStack-based testing is covered by a full CI run.

### DIFF
--- a/.buildkite/pipeline.bb.yml
+++ b/.buildkite/pipeline.bb.yml
@@ -50,3 +50,4 @@ steps:
     concurrency: 5
     concurrency_group: 'bitbar-ios-15'
     concurrency_method: eager
+


.buildkite/pipeline.full.yml
```diff
@@ -73,13 +73,13 @@ steps:
       queue: opensource
     plugins:
       artifacts#v1.5.0:
-        download: ["features/fixtures/ios/output/iOSTestApp.ipa"]
+        download: "features/fixtures/ios/output/ipa_url.txt"
```
twometresteve, line 76: This approach allows the app to be uploaded to BrowserStack just once for a CI run (the text file just includes a unique reference for the upload).  The change was made some time ago, but got missed on this particular step.

features/steps/cocoa_steps.rb
```diff
@@ -44,45 +44,6 @@ def request_fields_are_equal(key, index_a, index_b)
   )
 end
 
-def check_device_model(field, list)
```
twometresteve, line 47: BitBar name their devices differently, so there's no way to get a reliable handle on the model name using Appium.  I've therefore removed all of this functionality and now just match the fields against a regex.  If the regex matches it seems very unlikely that the notifier will have generated an incorrect value.

.buildkite/pipeline.bb.yml
```diff
@@ -0,0 +1,52 @@
+steps:
+
+  - label: 'BitBar Poc - iOS 15 - batch 1'
+    depends_on: "cocoa_fixture"
+    timeout_in_minutes: 90
+    agents:
+      queue: opensource
+    plugins:
+      artifacts#v1.9.0:
+        download: "features/fixtures/ios/output/iOSTestApp.ipa"
```
kstenerud, line 10: Should this be `ipa_url.txt` as well?

twometresteve: Eventually, yes, but we're only uploading the app to BrowserStack at the moment.  We can do the same for BitBar in time, but it's more effort than I wanted to put in at this stage.

kstenerud (approved):

PR #1526, opened 2023-03-01
[Plat-9721] fix race condition in BSGRunContext

kstenerud:
## Goal

`BSGRunContext.InitRunContext()` was spawning a BG task that calls `BSGRunContextUpdateMemory()`, which `InitRunContext()` is also calling, resulting in a race condition.

## Design

Change the order to call `BSGRunContextUpdateMemory()` before spawning the BG task.

Also changed all functions defined as `func()` to `func(void)` because xcode cloud no longer allows it.

## Testing

Tested manually.

### DIFF
--- a/Bugsnag/Helpers/BSGHardware.h
+++ b/Bugsnag/Helpers/BSGHardware.h
@@ -18,11 +18,11 @@
 #pragma mark Device
 
 #if TARGET_OS_IOS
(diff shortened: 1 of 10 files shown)


robert-smartbear (approved):

lemnik (approved):
LGTM

PR #1527, opened 2023-03-02
[Plat-9721] race condition  in hostMemoryFree

kstenerud:
## Goal

Another race condition identified by Apple. Use atomics like for the other fields of this type.


lemnik (approved):
LGTM

PR #1532, opened 2023-03-07
[PLAT-9724] Disable oom test on ios 16

kstenerud:
## Goal

The OOM scenario flakes often on ios 16, so disable it on that OS for now.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -69,6 +69,14 @@ def skip_below(os, version)
   skip_this_scenario("Skipping scenario") if Maze::Helper.get_current_platform == os and Maze.config.os_version < version
 end
 
+def skip_between(os, version_lo, version_hi)
+  skip_this_scenario("Skipping scenario") if Maze::Helper.get_current_platform == os and Maze.config.os_version >= version_lo and Maze.config.os_version <= version_hi
+end
+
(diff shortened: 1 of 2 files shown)


features/support/env.rb
```diff
@@ -69,6 +69,14 @@ def skip_below(os, version)
...
 end
 
+def skip_between(os, version_lo, version_hi)
+  skip_this_scenario("Skipping scenario") if Maze::Helper.get_current_platform == os and Maze.config.os_version >= version_lo and Maze.config.os_version <= version_hi
+end
+
+Before('@skip_ios_16') do |scenario|
+  skip_between('ios', 16, 16.99)
+end
+
```
twometresteve, line 79: I guess this is fine, but `Maze.config.os_version.floor == 16` might have saved a couple of lines of code.

twometresteve (approved):

features/support/env.rb
kstenerud: I figured this would be more versatile because it could skip a specific sub-version (e.g. `skip_between('ios', 16.1, 16.199)`, although it turns out you can't specify this in our current host...)

twometresteve: Yeah, I'm happy to keep it like it anyway.

PR #1554, opened 2023-04-28
[Plat-9934] Don't generate URLSession on config copy

kstenerud:
## Goal

The configuration object was generating a fresh URLSession whenever none was provided, but since many internal functions end up calling along the same path, the client ended up creating many sessions. This PR cleans up the behaviour a bit.

## Design

Prevent `BugsnagConfiguration` from pre-emptively taking action. I've added a new method `sessionOrDefault`, which calls on a common internal shared session if there was no configured session. This code should ideally be situated elsewhere, but for now we want to keep any changes shallow while we decide how to handle sessions long-term (see PLAT-10067).

## Testing

Outward behaviour should be unaffected, so existing tests are sufficient.

### DIFF
--- a/Bugsnag/Client/BugsnagClient.m
+++ b/Bugsnag/Client/BugsnagClient.m
@@ -341,7 +341,7 @@ BSG_OBJC_DIRECT_MEMBERS
 }
 
 - (void)sendLaunchCrashSynchronously {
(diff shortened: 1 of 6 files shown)


tomlongridge (approved):
LGTM - pending changelog (worth pointing out the change, even if it shouldn't be noticeable)

PR #1563, opened 2023-06-09
[PLAT-10188] Log Cocoa performance library startup times to MR

robert-smartbear:


Gemfile
```diff
@@ -4,7 +4,7 @@ gem 'cocoapods'
 
 # A reference to Maze Runner is only needed for running tests locally and if committed it must be
 # portable for CI, e.g. a specific release.  However, leaving it commented out would mean quicker CI.
-gem 'bugsnag-maze-runner', '~> 7.0'
+gem 'bugsnag-maze-runner', '~> 7.30.2'
```
twometresteve, line 7: Why this change?  It means that you won't automatically get 7.31.0 and greater in future, but maybe that's what you want.

twometresteve (commented):
I can't make much comment about the Objective-C changes, but are you sure they are not responsible for the repeated test failures on CI?

PR #1565, opened 2023-06-22
Fix visionOS support

msmollin:
## Goal

This fixes a build error when targeting visionOS (xrOS) devices in Xcode 15 beta 2
Screenshot 2023-06-21 at 11 05 13 PM

## Design

Per the docs listed here:

https://developer.apple.com/documentation/visionos/bringing-your-app-to-visionos#Isolate-features-that-are-unavailable-in-visionOS

>When separating code for visionOS and iOS, understand that conditional checks for iOS also return true in visionOS, so place visionOS conditions first and execute the iOS code only if visionOS isn’t present.

## Changeset

Added a `!TARGET_OS_XR` condition on the front per recommendations from Apple docs.

## Testing

Imported my fork into my project and it built whereas it did not before.

### DIFF
--- a/Bugsnag/Breadcrumbs/BSGNotificationBreadcrumbs.m
+++ b/Bugsnag/Breadcrumbs/BSGNotificationBreadcrumbs.m
@@ -413,7 +413,7 @@ static NSString *nullStringIfBlank(NSString *str) {
 
 #pragma mark -
 
-#if TARGET_OS_IOS
+#if !TARGET_OS_XR && TARGET_OS_IOS
 
 - (void)orientationDidChange:(NSNotification *)notification {
     UIDevice *device = notification.object;
--- a/Bugsnag/Client/BugsnagClient.m
+++ b/Bugsnag/Client/BugsnagClient.m
@@ -435,7 +435,7 @@ BSG_OBJC_DIRECT_MEMBERS
     BSGGetDevice().batteryMonitoringEnabled = FALSE;
 #endif
 
-#if TARGET_OS_IOS
+#if !TARGET_OS_XR && TARGET_OS_IOS
     [[UIDEVICE currentDevice] endGeneratingDeviceOrientationNotifications];
 #endif
 }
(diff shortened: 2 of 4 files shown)


msmollin:
This likely can't merge until Xcode 15 is generally available but figured I'd put this up to help folks trying out the betas.

marcpalmer:
I think this could land pre-Xcode 15 release with a check for the definition of the macro? Wouldn't something like this work:

```
#ifndef TARGET_OS_XROS
  #define TARGET_OS_XROS 0
#endif
```

If this does not ship until after Xcode 15 is GM nobody can use bugsnag in the interim without forking, and even after Xcode 15 is GM, many people will still be using Xcode 14 so we need a fix that works for any Xcode release?

msmollin:
@marcpalmer that's a good point. My objc is fairly rusty so thanks for the direction. Added that [here](https://github.com/bugsnag/bugsnag-cocoa/pull/1565/files#diff-d82695cc410a0f04a221273d8733628da2beae4128c98c3303c047c0ed37e638R28).

johnkiely1:
HI @msmollin @marcpalmer , Thanks for this PR, adding official support for VisionOS is definitely something we are interested in doing, however do need to do some additional work to fully understand the scope of changes involved. Going to leave this open for now and we will revisit soon. Sounds like forking the repo is the most reasonable workaround for now.

matthargett:
With Xcode 15.2 now officially (no longer a beta) released with visionOS support, should this be revisited? Sentry also has  a PR up that adds visionOS support.

clr182:
Hi @matthargett 
Thank you for reaching out on this issue. 
VisionOS is a platform we do aim to support at some time in the future. However, as for our previous message, we still do not have a clear ETA on a release. 
Once we have more information to share, we will be sure to update this thread!

tomlongridge:
As of [v6.29.0](https://github.com/bugsnag/bugsnag-cocoa/releases/tag/v6.29.0) we have fixed the build issues with visionOS. We do not intend on having official support for it as a platform, but want to ensure our SDK at least compiles against it.

PR #1578, opened 2023-09-07
[Plat-10822] duplicate -lc++

kstenerud:
## Goal

Xcode 15 now warns about duplicate library flags during the linking phase.

Since c++ is now inferred by the compiler, adding `-lc++` hasn't been necessary in Xcode for the past few versions.

This PR removes the manual `-lc++` flag in the project and podspec.

## Testing

Reran all tests to ensure that this doesn't break any builds.

### DIFF
--- a/Bugsnag.podspec.json
+++ b/Bugsnag.podspec.json
@@ -42,7 +42,7 @@
     "-fvisibility=hidden"
   ],
   "libraries": [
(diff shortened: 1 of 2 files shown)


robert-smartbear (approved):

robert-smartbear:
This PR is outdated. Closing

Issue #1582, opened 2023-09-14
App crashes when trying to start Bugsnag on Watch app


gpambrozio:
### Describe the bug
When trying to start Bugsnag on a Watch app (not an extension) the app crashes when trying to call `[WKExtension sharedExtension]` as this only works on extensions, not on a watch app that can run as a standalone app. 

### Steps to reproduce
1. On latest XCode, create a new Apple Watch app.
2. Follow instructions to [integrate Bugsnag](https://docs.bugsnag.com/platforms/watchos/)
3. Run app 
4. XCode will stop at the crash mentioned above

### Environment
* Bugsnag version: 6.27.2
* watchOS version(s): 9
* Simulator or physical device: Any watch
* Xcode version: 14.3.1

### Crash screenshot:

![CleanShot 2023-09-14 at 11 44 29](https://github.com/bugsnag/bugsnag-cocoa/assets/672135/5a5bb821-88cd-40a7-8e50-f685c60a6a62)

clr182:
Hi @gpambrozio 
We appreciate your feedback and thank you for opening a PR to address this issue. 
We will be sure to review your PR once priorities allow.

PR #1583, opened 2023-09-14
Fixes issue when using Bugsnag on a WatchOS app

gpambrozio:
The code was trying to access an extension but on latest WatchOS we can have a standalone app. Related to [this issue](https://github.com/bugsnag/bugsnag-cocoa/issues/1582)

## Goal

Not crash when using Bugsnag on a standalone Apple watch app.

## Design

Just added some checks to see if code is running on an app or extension.

## Testing

See steps in [related issue](https://github.com/bugsnag/bugsnag-cocoa/issues/1582). With this branch the crash does not happen.

### DIFF
--- a/Bugsnag/Helpers/BSGRunContext.m
+++ b/Bugsnag/Helpers/BSGRunContext.m
@@ -109,8 +109,13 @@ static bool GetIsActive(void) {
 #endif
 
 #if TARGET_OS_WATCH
-    WKExtension *ext = [WKExtension sharedExtension];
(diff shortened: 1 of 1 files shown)


chickdan:
+1 to this. I noticed the SDK works fine with watch targets created prior to Xcode 14, but new targets and consolidated targets will crash.

tomlongridge:
Thanks @gpambrozio - the fix looks good. We're going to carry it on in #1588 to allow CI to run and to merge it down.

PR #1588, opened 2023-10-11
Fixes issue when using Bugsnag on a WatchOS app

tomlongridge:
The code was trying to access an extension but on latest WatchOS we can have a standalone app.


tomlongridge (commented):
LGTM

kstenerud (approved):

PR #1593, opened 2023-10-23
[Plat-11007] null user components

kstenerud:
## Goal

Setting user fields to nil wouldn't clear out their values when serializing a report because [user toJson] would produce dictionaries with the null fields missing, and then when merged with the existing data the old data would remain (because there's no new field value to override the old).

This PR ensures that the user object's produced dictionary always contains all keys, even if they've been set to null.

## Testing

Added unit tests.


robert-smartbear (approved):

kstenerud:
Replaced with https://github.com/bugsnag/bugsnag-cocoa/pull/1599

PR #1601, opened 2023-11-02
Harmonize fixture logging

kstenerud:
## Goal

This PR harmonizes all of the logging in the CI fixtures so that they all use the same function names (`logInfo`, `logWarn`, `logError`) in Swift and Objective-C, and support optional varargs. These log functions all go to the same "channel" that prefixes `bugsnagci <level>` so that the fixture messages can easily be filtered by common tooling.

### DIFF
--- a/features/fixtures/shared/scenarios/NSExceptionShiftScenario.m
+++ b/features/fixtures/shared/scenarios/NSExceptionShiftScenario.m
@@ -1,4 +1,5 @@
 #import "Scenario.h"
+#import "Logging.h"
 
 @interface NSExceptionShiftScenario : Scenario
 @end
--- a/features/fixtures/shared/scenarios/OOMScenario.m
+++ b/features/fixtures/shared/scenarios/OOMScenario.m
@@ -58,7 +58,7 @@ void onCrashHandler(const BSG_KSCrashReportWriter *writer) {
     [Bugsnag setContext:@"OOM Scenario"];
     [NSNotificationCenter.defaultCenter addObserverForName:UIApplicationDidReceiveMemoryWarningNotification object:nil
                                                      queue:nil usingBlock:^(NSNotification *note) {
-        logInfo(@"*** Received memory warning");
+        logDebug(@"*** Received memory warning");
     }];
     // Delay to allow session payload to be sent
     [self performSelector:@selector(consumeAllMemory) withObject:nil afterDelay:2];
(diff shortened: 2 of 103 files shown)


features/fixtures/shared/scenarios/NSExceptionShiftScenario.m
```diff
@@ -1,4 +1,5 @@
 #import "Scenario.h"
+#import "Logging.h"
```
twometresteve, line 2: Do we really need to include the header in all the scenario files, when there are no other changes?

features/fixtures/shared/scenarios/OOMScenario.m
```diff
@@ -66,11 +67,11 @@ - (void)run {
 - (void)consumeAllMemory {
     struct utsname system = {0};
     uname(&system);
-    NSLog(@"*** Device = %s", system.machine);
+    logInfo(@"*** Device = %s", system.machine);
```
twometresteve, line 70: From my perspective, logs like these are too low-level to have the same prefix.  What I'd really like to do is come to a device log and be able to quickly get a medium-level context of what's going on with the e2e tests.  So (for me) these things are all really valid:
- Launching the app
- Finding reading the fixture config file
- Polling for commands
- Starting scenarios
- etc

But things that are very scenario specific I would rather leave to normal logging, as not to dilute the power of being able to focus in on an area very quickly.  Maybe we could have another standard prefix for the more detailed stuff?

features/fixtures/shared/scenarios/Scenario.m
```diff
@@ -218,28 +213,29 @@ + (void)clearPersistentData {
 }
 
 + (void)executeMazeRunnerCommand:(void (^)(NSString *scenarioName, NSString *scenarioMode))preHandler {
-    NSLog(@"%s", __PRETTY_FUNCTION__);
+    logInfo(@"%s", __PRETTY_FUNCTION__);
```
twometresteve, line 216: Same for me here - worthy of logging, for sure, but breaks the level of abstraction that I had been aiming for.

features/fixtures/shared/scenarios/NSExceptionShiftScenario.m
kstenerud: Scenario.h is currently the bridging header, so anything it sees, Swift sees, which causes problems with the naming clash. We could try hoisting the things Swift needs out into another header file, but I ran out of time. I can look again later.

features/fixtures/shared/scenarios/OOMScenario.m
kstenerud: We could add a "debug" channel for stuff like this...

twometresteve (approved):

PR #1603, opened 2023-11-03
Fix Compliation Errors for Xcode 15

neil477:
## Goal
- Get Bugsnag-Cocoa Library to compile on Xcode 15.0.1 and above.

## Design
- Initial Error `Darwin.POSIX.sys.types' appears within extern "C" language linkage specification`. The `#include <sys/types.h>` contains C++ code. When nested in `extern "C"` block the code is expected be only C, so the solution is to move the include outside of the extern. This seems to be a new error with the complier that ships with XCode 15. 

- Second Error after moving the include `Declaration of 'size_t' must be imported from module 'Darwin.C.stddef' before it is required`, this just necessitated including `<stddef.h>` in the extern C again.

## Changeset
- Needed to change the order of include statements and add missing header for C code.

## Testing
- tested compiling and running Bugsnag code in our app.

### DIFF
--- a/Bugsnag/include/Bugsnag/BSG_KSCrashReportWriter.h
+++ b/Bugsnag/include/Bugsnag/BSG_KSCrashReportWriter.h
@@ -32,11 +32,13 @@
 #define HDR_BSG_KSCrashReportWriter_h
 
 #ifdef __cplusplus
+#include <sys/types.h>
+
 extern "C" {
 #endif
 
 #include <stdbool.h>
-#include <sys/types.h>
+#include <stddef.h>
 
 /**
  * Encapsulates report writing functionality.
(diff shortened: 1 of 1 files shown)


samueljim:
LGTM
The same change is what I did in order to fix the code on my project

clr182:
Hi,

Thank you for opening this PR. We do have an item on our backlog aimed at fixing this issue. We don't have an ETA on the release of this, but we'll be sure to update you once we have more info.

samueljim (commented):
LGTM

tomlongridge:
@neil477 – thanks for this fix. Just to let you know we've included part of it in #1608, which has now been released in [v6.27.3](https://github.com/bugsnag/bugsnag-cocoa/releases/tag/v6.27.3).

We were unable to reproduce the second issue you mention with the `<stddef.h>` include. Perhaps you can give it a re-test on the latest release and let us know if you see any issues?

PR #1625, opened 2024-02-02
Add support for visionOS

bcherry:
## Goal

Currently you cannot use Bugsnag in a native visionOS app. We'd like to be able to offer a visionOS version of a native iOS app which uses Bugsnag.

## Design

I modeled this off the way that tvOS is integrated, for the most part, as a separate framework.

## Changeset

Add compiler macros to support visionOS builds.
Add visionOS framework target.

## Testing

Testing has been lightweight so far, I'd imagine that Bugsnag themselves would like to own serious testing of new platform support.


tomlongridge:
These changes have now been merged into #1660 to provide some basic support for visionOS in the latest release (v6.29.0). Thanks for your input @bcherry.

PR #1628, opened 2024-02-12
Add auto retries on app-hang errors

Cawllec:
## Goal

To enable currently unstable tests to be retried without manual effort, we're adding automatic retries in the event the app-hang tests fail due to an appium error.

Note: This doesn't currently retry if the test fails due to no errors being received.  While this could very well be caused by an appium issue with the app-hang tests, it could also present due to changes in the notifier code, and should be (minimally) investigated.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -335,6 +335,8 @@ steps:
       automatic:
         - exit_status: -1  # Agent was lost
           limit: 2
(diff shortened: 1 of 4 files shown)


Cawllec:
Note: This relies on a change to maze-runner, which hasn't yet been released.  Once that's done, this will be updated to point back to the main branch of MR.

twometresteve (approved):

PR #1654, opened 2024-05-23
[Plat-11927] fixture backgrounding

kstenerud:
## Goal

Add support for backgrounding the fixture app in e2e tests.

### DIFF
--- a/features/steps/app_steps.rb
+++ b/features/steps/app_steps.rb
@@ -28,6 +28,14 @@ When("I relaunch the app after a crash") do
   relaunch_crashed_app
 end
 
+When('I switch to the web browser for {int} second(s)') do |duration|
+  execute_command "background", [duration.to_s]
+end
+
+When('I switch to the web browser') do
+  execute_command "background", ["-1"]
+end
+
 #
 # Setting scenario mode
 #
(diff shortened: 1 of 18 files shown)


features/delivery.feature
```diff
@@ -107,6 +107,8 @@ Feature: Delivery of errors
     Given I set the HTTP status code for the next request to 500
     And I run "AutoSessionScenario"
     And I wait to receive a session
+    And I background the app for 0 seconds
+    And I wait for 10 seconds
```
twometresteve, line 111: I think these are debug lines no longer needed?  The tests are failing because the first isn't defined.

features/steps/app_steps.rb
```diff
@@ -28,6 +28,18 @@
   relaunch_crashed_app
 end
 
+When('I switch to the web browser for {int} seconds') do |duration|
+  execute_command "background", [duration.to_s]
+end
+
+When('I switch to the web browser 1 second') do |duration|
+  execute_command "background", ["1"]
+end
```
twometresteve, line 37: There's a missing "for" in the sentence here, but all you need to do is put brackets around the "s" of seconds in the previous step and it makes the "s" optional.

features/delivery.feature
kstenerud: Yup, no longer needed.

features/delivery.feature
```diff
@@ -107,6 +107,7 @@ Feature: Delivery of errors
     Given I set the HTTP status code for the next request to 500
     And I run "AutoSessionScenario"
     And I wait to receive a session
+    And I switch to the web browser for 0 seconds
```
twometresteve, line 110: Not sure if you have a commit coming, but this still needs to be removed at some point.

twometresteve (approved):
Nice to see working!

PR #1661, opened 2024-06-13
[Plat-12269] Prevent inlining of stack trace entries that are going to be pruned

kstenerud:
## Goal

We prune the top of the stack trace so that our stack trace gathering code doesn't show in customer traces. However, if the optimizer inlines any of the functions/methods along the path, it breaks the pruning.

## Design

Force everything along the trace gathering path to NOT be optimized, so that they are never folded or inlined.


lemnik (approved):
LGTM

PR #1663, opened 2024-06-18
Plat 12250 build xcframeworks

kstenerud:
## Goal

This PR adds the ability to build the binary xcframework from the command line.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,9 @@ build_ios_static: ## Build the static library target
 build_carthage: ## Build the latest pushed commit with Carthage
 	@./scripts/build-carthage.sh
 
+build_xcframework: ## Build as a multiplatform xcframework
+	@./scripts/build-xcframework.sh
+
 build_swift: ## Build with Swift Package Manager
 	@swift build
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -89,6 +89,9 @@ build_ios_static: ## Build the static library target
 build_carthage: ## Build the latest pushed commit with Carthage
 	@./scripts/build-carthage.sh
 
+build_xcframework: ## Build as a multiplatform xcframework
```
twometresteve, line 92: nit: The Make target calls it xcframework but the script just uses framework.  Maybe less likely to trip people up if we us framework for both?  I can change it on the PR to add it to CI, so long as you agree.

kstenerud: Oh, it's actually building an xcframework, so best to settle on that I guess

twometresteve: Yeah, I don't mind at all which one is used, I'm just likely to forget and get them wrong if they aren't the same - I'll chang eit to xc when I get to it no problem.

kstenerud: Everything should be xcframework now

twometresteve (approved):

PR #1667, opened 2024-06-27
[PLAT-12346] Remove inlined frame from the Bugsnag notify call chain entirely so t…

kstenerud:
## Goal

Remove the call to functions that are getting inlined so that we don't have to worry about the size of the bugsnag portion of the stack (which will change if the call gets inlined).

**Note**: Do not merge this yet. We will also be trying a different approach.


kstenerud:
Superseded by https://github.com/bugsnag/bugsnag-cocoa/pull/1668

PR #1682, opened 2024-07-23
[Plat-12473] bad header

kstenerud:
There was an extra copy of `BugsnagCorrelation.h` in the public headers directory that was causing warnings to appear when building an application:

`bugsnag-cocoa/Bugsnag/include/Bugsnag/Bugsnag.h:393:1: warning: umbrella header for module 'Bugsnag' does not include header 'BugsnagCorrelation.h'`

This extra header has now been removed.


robert-smartbear (approved):

twometresteve (commented):
No comments from me.

PR #1684, opened 2024-07-26
Run e2e tests against XcFramework test fixture

twometresteve:
## Goal

Run e2e tests against a test fixture built with the BugSnag XcFramework.

## Design

The test fixture for iOS was duplicated and modified to include the XcFramework for the notifier.  A subset of tests will be run against this, on the earliest and latest iOS versions, for a full build.

## Changeset

While preparing this change I checked that the whole set of e2e tests pass against the fixed, so there are some changes to tests that won't actually be run as part of the cut-down set.   

## Testing

Covered by a full CI run.


.buildkite/pipeline.yml
```diff
@@ -17,10 +18,10 @@ steps:
       - make build_xcframework
       - make build_swift
       - make build_ios_static
+      - zip -r Bugsnag.xcframework.zip build/xcframeworks/products/Bugsnag.xcframework
```
kstenerud, line 21: This should also have the `--symlinks` flag so that it doesn't actually store the symlinked files twice.

features/release/barebone_tests.feature
```diff
@@ -85,10 +85,11 @@ Feature: Barebone tests
...
     And the event "usage.config" is not null
-    And the event "usage.config.staticallyLinked" equals the platform-dependent boolean:
-      | ios     | true  |
-      | macos   | @null |
-      | watchos | @null |
+    # TODO
+#    And the event "usage.config.staticallyLinked" equals the platform-dependent boolean:
+#      | ios     | true  |
+#      | macos   | @null |
+#      | watchos | @null |
```
kstenerud, line 92: Is this still a TODO for this PR, or for a later PR?

kstenerud (approved):

features/release/barebone_tests.feature
twometresteve: Have moved this to another feature file now, which isn't being run against the xcFramework fixture.

PR #1686, opened 2024-08-01
Re-enable app-hang tests

Cawllec:
## Goal

These don't seem to be an issue running on BrowserStack devices right now, so after re-running them a couple of times we should be able to re-enable them.


Cawllec:
We're not enabling iOS 13 app-hang tests for now, ticketed here: https://smartbear.atlassian.net/browse/PLAT-12554

twometresteve (approved):

PR #1693, opened 2024-09-05
Use direct project references instead of cocoapods for the macos fixture

kstenerud:
## Goal

Remove the macos fixture's dependence on Cocoapods so that all fixture targets use the same build methodology.


twometresteve:
The Buildkite build is failing when trying to `pod install`, I'm guessing it's just a matter of removing it from the step in `pipeline.yml` assuming the command line build process is the same otherwise.

twometresteve (approved):
LGTM

PR #1697, opened 2024-09-27
Add Buildkite analytics collectors to pipeline

Cawllec:
I've left out the MacOS 10.* tests, as those machines will need to be updated manually with the test analytics token, and I'm not quite sure the extent necessary yet.  I'm going to ticket that separately.


joshedney (approved):
LGTM!

PR #1704, opened 2024-11-08
Use a XIB project for macos xcframework so that we can reuse the existing macos fixture code

kstenerud:


features/fixtures/shared/scenarios/CxxUnexpectedScenario.mm
```diff
@@ -17,7 +17,7 @@ @implementation CxxUnexpectedScenario
 
 - (void)run {
     [[NSThread mainThread] setName:@"BSG MAIN THREAD"];
-    std::unexpected();
```
kstenerud, line 20: Note: `std::unexpected` has been deprecated and later removed.

twometresteve (approved):

PR #1715, opened 2024-12-11
Feature flags performance improvements

robert-smartbear:
## Goal

Reduce the time it takes to add a feature flag

## Design

Currently feature flags are stored in the same file as metadata which means that each time a feature flag is added, all that data has to be sanitized and saved. By saving the feature flags into separate files we reduce the processing required significantly.

## Changeset

Removed the feature flags from metadata object and created a separate file store for them.

## Testing

E2E tests


kstenerud (approved):

saad352:
> Currently feature flags are stored in the same file as metadata which means that each time a feature flag is added, all that data has to be sanitized and saved. By saving the feature flags into separate files we reduce the processing required significantly.

@robert-smartbear this design choice is resulting into slow Applaunch. 

When Bugsnag is init on the App start. It tries to clear the Feature Flags [here](https://github.com/bugsnag/bugsnag-cocoa/blob/f4348c29cc881ce80e4e22a46f29d7f5a18e924b/Bugsnag/FeatureFlags/BSGPersistentFeatureFlagStore.m#L102) in a for loop and if you have too many Feature flags. This result in slow down of the app launch.

We have ~200 Feature flags and 42% of slow launches in our app are happening due to this.

This deleting part should be moved to background thread? There could be side effects of that, new Feature flags might get added while the earlier ones are being removed. Thus causing inconsistency or other problems. But I believe those can be avoided by various other techniques.

Or other option could be to write Feature Flags in a new directory on every app init of Bugsnag while previous one is deleted in the background. Offcourse for this some state needs to be managed by SDK. 

Can you please fix this at your earliest convenience. I am afraid we can not do much on our end to mitigate it except minimizing the number of FeatureFlags which is not a practical option for us right now.

PR #1716, opened 2024-12-12
[PLAT-13234] disable fixture stripping

kstenerud:
## Goal

The xcframework macos fixture had application stripping turned on, which makes the notifier unable to symbolicate on-device (and this breaks our e2e tests).

Turn stripping off in this target to match the non-xcframework target.


features/release/barebone_tests.feature
```diff
@@ -112,7 +112,7 @@ Feature: Barebone tests
     And the "isPC" of stack frame 0 is null
     And the "isLR" of stack frame 0 is null
     # TODO: Skipped pending PLAT-13234
```
twometresteve, line 114: The comment here and below need to be removed now.

twometresteve (dismissed):
Ha, that was nice and simple!  I've confirmed locally that the tests now pass, just the comments to be removed as noted.

PR #1764, opened 2025-04-16
[plat 14105] move cocoa e2e to isolated queue

joshedney:
## Goal

Move cocoa macos 14 tests to isolated queue

## Testing
Covered by CI


twometresteve (commented):
LGTM as long as a Cocoa engineer is content with the time increase (I would suggest increasing them to 2900 and 5700 - IMO the test is really just checking that the times are plausible and judging that the code is most likely using the correct mechanism).

robert-smartbear (approved):

PR #1805, opened 2025-06-10
Restore part of the E2E tests

DariaKunoichi:
Set System and AppState to be default monitors in KSCrash.
Fixed timestamp - calculated from microseconds.
Added `binary_arch` and `clang_version` fields in KSCrash report.
Restored system name to be always `macOS` of OSX.
Restored saving disc info from statfs just for macOS.


features/app_hangs.feature
```diff
@@ -44,7 +42,7 @@ Feature: App hangs
 
     And the error payload field "events.0.device.osName" equals the platform-dependent string:
       | ios   | iOS    |
-      | macos | Mac OS |
+      | macos | macOS |
```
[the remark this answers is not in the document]
DariaKunoichi: Previously it was "Mac OS" because old KSCrash version had it hardcoded, now it's fine so the new "macOS" is used.
https://github.com/bugsnag/bugsnag-cocoa/blob/a3f56c6191f0d8ae6b2407aad026ce3580298ce1/Bugsnag/KSCrash/Source/KSCrash/Recording/BSG_KSSystemInfo.m#L317

features/release/unhandled_mach_exception.feature
```diff
@@ -11,7 +11,7 @@ Feature: Bugsnag captures an unhandled mach exception
     And I wait to receive an error
     Then the error is valid for the error reporting API
     And the event "exceptions.0.errorClass" equals "EXC_BAD_ACCESS"
-    And the event "exceptions.0.message" equals "Attempted to dereference garbage pointer 0xdeadbeef."
+    And the event "exceptions.0.message" starts with "Attempted to dereference garbage pointer 0xdeadbeef."
```
robert-smartbear, line 14: Why are we changing this?

DariaKunoichi: Because the message now is longer, and have more details.

PR #1841, opened 2025-09-02
Correct IPHONEOS_DEPLOYMENT_TARGET for xcframework fixture

twometresteve:
## Goal

Corrects IPHONEOS_DEPLOYMENT_TARGET for xcframework fixture.

## Design

We didn't notice this before because the XcFramework test fixture is only built on a full CI run.

## Testing

Covered by a full CI run.


DariaKunoichi (approved):
I saw the warning before but did not know it was this significant

twometresteve:
> I saw the warning before but did not know it was this significant

Yes, it's strange - there are still some similar warnings in the build log, I'm not sure why it was a hard failure in this particular case.

PR #1850, opened 2025-09-14
Remote SDK Config: Download and store

robert-smartbear:
## Goal

Allow the Cocoa SDK to be remotely and directly configured to discard errors based on selected conditions. This would provide both us and our customers a way to regulate the volume or errors their apps produce.

## Changeset

- Added a new JSON based configuration file for the SDK
- Retrieval of the remote configuration (if it is available)
- Timeouts / TTL for the remote configuration
- A new configuration option to set the remote configuration endpoint (or opt-out if set to nil)
- New configuration options to set the remote configuration automatic update interval and tolerance

## Testing

Tests will be added in a separate PR

### DIFF
--- a/Bugsnag/Configuration/BugsnagConfiguration.m
+++ b/Bugsnag/Configuration/BugsnagConfiguration.m
@@ -445,6 +445,15 @@ static NSURLSession *getConfigDefaultURLSession(void) {
     return self.endpoints.sessions.length ? [NSURL URLWithString:self.endpoints.sessions] : nil;
 }
 
+- (NSURL *)configurationURL {
+    NSString *configurationUrlString = self.endpoints.configuration;
+    if (!configurationUrlString) {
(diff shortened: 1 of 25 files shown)


Bugsnag/Configuration/BugsnagConfiguration.m
```diff
@@ -188,6 +188,8 @@ - (instancetype)initWithApiKey:(NSString *)apiKey {
     _maxPersistedSessions = 128;
     _maxStringValueLength = 10000;
     _autoTrackSessions = YES;
+    _remoteConfigUpdateInterval = 86400;
+    _remoteConfigUpdateTolerance = 7200;
```
DariaKunoichi, line 192: what are those values? seconds? bytes?

DariaKunoichi (approved):
LGTM with small comment

PR #1852, opened 2025-09-22
Discard processor and rules

robert-smartbear:
## Goal

Allow the Cocoa SDK to be remotely and directly configured to discard errors based on selected conditions. This would provide both us and our customers a way to regulate the volume or errors their apps produce.

## Changeset

- Added event discard processor
- Creating discard rules based on remote config
- Applying discard rules before sending the payload
- Implemented rules: ALL, ALL_HANDLED

## Testing

Tests will be added in a separate PR

### DIFF
--- a/Bugsnag/DiscardProcessor/Factory/BSGEventDiscardRuleFactory.h
+++ b/Bugsnag/DiscardProcessor/Factory/BSGEventDiscardRuleFactory.h
@@ -7,7 +7,7 @@
 //
 
 #import "../Source/Model/BSGEventDiscardRule.h"
-#import "BSGRemoteConfiguration.h"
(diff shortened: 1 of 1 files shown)


Bugsnag/RemoteConfig/Store/Model/BSGRemoteConfiguration.m
```diff
@@ -132,9 +125,6 @@ - (NSDictionary *)toJson {
     if (self.configurationTag) {
         result[ConfigurationTagKey] = self.configurationTag;
     }
-    if (self.appVersion) {
```
DariaKunoichi, line 135: why is app version removed?

DariaKunoichi (approved):

PR #1859, opened 2025-09-28
Remote config tests

robert-smartbear:
Fixes and E2E tests for Remote Config

### DIFF
--- a/features/release/remote_config_discard.feature
+++ b/features/release/remote_config_discard.feature
@@ -9,7 +9,9 @@ Feature: Remote config discard rules are applied
      | property | body                  | @features/support/config/no_rules.json     |
      | property | status                | 200                                        |
      | header   | Cache-Control         | max-age=604800                             |
-    And I run "RemoteConfigBasicScenario" and relaunch the crashed app
+    And I run "RemoteConfigBasicScenario" 
+    And on macOS, I wait for 10 seconds
+    And I relaunch the app after a crash
     And I configure Bugsnag for "RemoteConfigBasicScenario"
     And I wait to receive 2 errors
     And the received errors match:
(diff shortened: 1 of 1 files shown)


Bugsnag/RemoteConfig/Handler/BSGRemoteConfigHandler.m
```diff
@@ -105,6 +108,7 @@ - (BOOL)isRemoteConfigEnabled {
 }
 
 - (void)updateRemoteConfig {
+    self.isLoadingRemoteConfig = YES;
```
DariaKunoichi, line 111: should the access to this var be synchronized?

Bugsnag/RemoteConfig/Handler/BSGRemoteConfigHandler.m
```diff
@@ -105,16 +108,26 @@ - (BOOL)isRemoteConfigEnabled {
 }
 
 - (void)updateRemoteConfig {
+    @synchronized (self) {
+        self.isLoadingRemoteConfig = YES;
+    }
```
lemnik, line 113: Given how `isLoadingRemoteConfig` guards `updateRemoteConfig` shouldn't the `if` check also be here? That way the check itself can be guarded by the same `synchronized` block and remove any possible races?

Bugsnag/RemoteConfig/Service/BSGRemoteConfigService.m
```diff
@@ -92,31 +136,36 @@ - (void)downloadRemoteConfigWithRequest:(NSURLRequest *)request
...
+        
+        NSString *etag;
+        NSDate *expiryDate;
+        if ([response isKindOfClass:[NSHTTPURLResponse class]]) {
+            NSHTTPURLResponse *httpResponse = (NSHTTPURLResponse *)response;
+            etag = httpResponse.allHeaderFields[ETagHeader];
+            NSString *cacheControl = httpResponse.allHeaderFields[CacheControlHeader];
+            
+            expiryDate = [strongSelf expiryDateFromCacheControl:cacheControl];
+            if (httpResponse.statusCode == 304) {
```
lemnik, line 152: nit: could this be a named constant?

Bugsnag/RemoteConfig/Handler/BSGRemoteConfigHandler.m
lemnik: I mean that the "indirectly" is a problem, if we're using locks to ensure safety we shouldn't be relying on the caller to establish that lock, at least not without a clear naming convention (such as `updateRemoteConfigUnderLock`). The locks are re-entrant so re-locking them should be relatively cheap.

lemnik (dismissed):

lemnik (approved):

PR #1861, opened 2025-09-30
Fixed state breadcrumbs flakey test fix

robert-smartbear:
## Goal

There seems to be an issue with handling commands after the fixture returns from foreground on iOS 18.0+

## Changeset

The test schedules a notify call to be executed after the app returns to foreground

### DIFF
--- a/features/release/breadcrumbs.feature
+++ b/features/release/breadcrumbs.feature
@@ -41,7 +41,6 @@ Feature: Attaching a series of notable events leading up to errors
     Then the event has a "manual" breadcrumb named "Cache locked"
 
   # TODO: Flaky on iOS 18 - see PLAT-14585
-  @skip_ios_18
   @skip_below_ios_13
   @skip_macos
   Scenario: State breadcrumbs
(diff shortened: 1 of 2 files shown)


features/release/breadcrumbs.feature
```diff
@@ -41,7 +41,6 @@ Feature: Attaching a series of notable events leading up to errors
     Then the event has a "manual" breadcrumb named "Cache locked"
 
   # TODO: Flaky on iOS 18 - see PLAT-14585
```
twometresteve, line 43: This line should be removed now

features/release/breadcrumbs.feature
```diff
@@ -50,12 +49,12 @@ Feature: Attaching a series of notable events leading up to errors
     And I invoke "notify_error"
     And I wait to receive an error
     And I discard the oldest error
+    And I invoke "notify_error_on_foreground"
     # Now we know that the backgrounding will occur at an appropriate time
     And I switch to the web browser for 2 seconds
     # Give iOS sufficient time to raise the notification for foregrounding the app
     And I make the test fixture wait for 1 second
```
twometresteve, line 56: We don't need this line and comment now (the fixture may fail to get the command it tries to send anyway).

twometresteve (commented):
Looks great - just a couple of minor improvements that could be made and I'm also soak-testing it (with my suggested removal) across many iOS 18/26 devices to avoid us coming back to this any time soon.

PR #1874, opened 2025-10-29
Codesign Xcframework Test Fixture

joshedney:
## Goal

Sign XcFramework fixture with an ad-hoc certificate so that it can get past the macOS Gatekeeper

## Testing

Covered by CI


scripts/build-xcframework.sh
twometresteve: This script is called by the `release` target in the makefile so signing the framework here would need a changelog entry, version bump, etc.  As we only want this for test purposes I would create a new script under `features/scripts` and just call that from the CI pipeline.  Or perhaps we could guard it with a new "code_sign" parameter defaulting to false?  That could be easier given that there are steps in the script following the code signing.

twometresteve (approved):

PR #1932, opened 2026-05-22
PLAT-16356 Update bugsnag-maze-runner version to 11.2

SB-khushbum:
## Goal

upgrade maze-runner version to latest 11.2


SB-adityap (approved):
Done

PR #1941, opened 2026-08-18
Release v6.37.0

SB-meiyalaganr:
## Goal

Allow the Cocoa SDK to be remotely and directly configured to discard errors based on selected conditions. 
Fixed a thread-safety issue in BSGPreventInlining that could cause rare crashes.
Fixed a crash after shared NSURLSession is invalidated.
Removed custom code signing settings that contained a redundant signing style and an incorrect development team ID.

## Design

These changes adopt a server-driven remote configuration approach for the Cocoa SDK while hardening its internals — using synchronization primitives (locks/atomics) to make BSGPreventInlining thread-safe, sanitizing debug symbol paths to prevent local file-system details from leaking into builds, and stripping redundant custom code-signing settings from the Xcode project to rely on Xcode's automatic signing instead.

## Testing

Tested with Unit Testing and Maze Runner Test Scenarios.


SB-shivamgarg (approved):
LGTM

SB-jigneshR (approved):

PR #1943, opened 2026-08-19
Meiy/plat 16585 Exclude data from cocoa sdk

SB-meiyalaganr:
## Goal

Add file backup support metadata to SDK-managed storage paths, allowing customers to configure whether SDK files are included in or excluded from device backups via isExcludedFromBackup.

## Design

SDK applies the configured backup metadata setting on fresh install, normalizes existing paths on relaunch when config changes, preserves metadata across file rewrites/replacements, and ensures crash-generated files inherit the correct backup flag on next launch.

## Testing

Unit tests cover metadata application and normalization logic. End-to-end tests validate fresh install, config change + relaunch, file rewrite preservation, crash file delivery, and full SDK lifecycle scenarios across true/false settings.


Bugsnag/Client/BugsnagClient.m
```diff
@@ -213,14 +214,21 @@ @interface BugsnagClient () <BSGAppHangDetectorDelegate>
 __attribute__((annotate("oclint:suppress[too many methods]")))
 #endif
 
+static void BSGApplyFileBackupSupportToExistingFiles(BSGFileLocations *fileLocations) {
+    // When customers upgrade from older SDK versions, Bugsnag files may already
+    // exist without backup metadata. Update only known SDK-owned paths so the
+    // scan does not include unrelated Application Support contents.
+    NSString *rootDirectory = fileLocations.events.stringByDeletingLastPathComponent;
+    [BSGFilesystem applyFileBackupSupportToPathAndContents:rootDirectory];
+}
```
sb-gaurava, line 223: BSGFilesystem` declares and implements both `needsFileBackupSupportReconciliation` and `markFileBackupSupportReconciled`, but neither is called here. As a result, `applyFileBackupSupportToPathAndContents:` runs a **full recursive directory enumeration and `setxattr` pass on every single app launch**, regardless of whether the backup policy has changed since last run.
**Impact:**
1. **Dead code** — `needsFileBackupSupportReconciliation` / `markFileBackupSupportReconciled` are unreachable code paths today.
2. **Startup latency / watchdog risk (`0x8badf00d`)** — On devices with large accumulated event or breadcrumb queues, synchronously enumerating and calling `setResourceValue:forKey:` on every file inside `initWithConfiguration:` (main thread) can block long enough to trigger a watchdog termination.
**Suggested fix:**
```objc
static void BSGApplyFileBackupSupportToExistingFiles(BSGFileLocations *fileLocations) {
    if ([BSGFilesystem needsFileBackupSupportReconciliation]) {
        // Narrow scan to SDK root only; does not include unrelated Application Support contents.
        NSString *rootDirectory = fileLocations.events.stringByDeletingLastPathComponent;
        [BSGFilesystem applyFileBackupSupportToPathAndContents:rootDirectory];
        [BSGFilesystem markFileBackupSupportReconciled];
    }
}

This ensures the expensive directory walk only runs on first install or when fileBackupSupport actually changes between launches, which is the exact use case the reconciliation methods were designed for.

File: Bugsnag/Helpers/BSGFilesystem.m [hunk 2]
```diff
@@ -42,4 +74,92 @@ + (nullable NSError *)rebuildPath:(NSString *)path {
...
+
++ (BOOL)writeData:(NSData *)data toFile:(NSString *)path options:(NSDataWritingOptions)options error:(NSError * __autoreleasing *)error {
+    BOOL success = [data writeToFile:path options:options error:error];
+    if (success) {
+        // File replacement can drop resource values, so reapply after each write.
+        bsg_log_debug(@"[File backup support] Wrote SDK file; reapplying backup policy: %@", path);
+        [self applyFileBackupSupportToPath:path];
+    }
+    return success;
+}
```
sb-gaurava, line 192: Redundant setxattr syscall on every individual file write

Apple's HFS+/APFS extended-attribute inheritance means that once NSURLIsExcludedFromBackupKey is set on a parent directory (e.g. breadcrumbs/, events/, sessions/) via ensurePathExists:, all files created within that directory automatically inherit the backup exclusion — no per-file setResourceValue:forKey: call is needed at write time.

Recommendation:

Keep applyFileBackupSupportToPath: in ensurePathExists: and on the specific root files that go through atomic replacement (e.g. metadata JSON, feature flags state).

Evaluate skipping the per-write reapplication for child files under directories already tagged (e.g. individual breadcrumb entries under breadcrumbs/), reducing syscall pressure on the hot write path.

Bugsnag/Helpers/BSGFilesystem.m
```diff
@@ -42,4 +74,92 @@ + (nullable NSError *)rebuildPath:(NSString *)path {
...
+    bsg_log_debug(@"[File backup support] Reconciling existing SDK path and contents: %@", path);
+    [self applyFileBackupSupportToPath:path];
+
+    BOOL isDirectory = NO;
+    if (![NSFileManager.defaultManager fileExistsAtPath:path isDirectory:&isDirectory] || !isDirectory) {
+        return;
+    }
+
+    NSDirectoryEnumerator<NSString *> *enumerator =
+    [NSFileManager.defaultManager enumeratorAtPath:path];
```
sb-gaurava: enumeratorAtURL:includingPropertiesForKeys:options:errorHandler: over enumeratorAtPath:

enumeratorAtPath: returns NSString relative paths and requires constructing a full NSURL per iteration inside applyFileBackupSupportToPath:.

enumeratorAtURL:includingPropertiesForKeys:options:errorHandler: returns NSURL objects directly and can pre-fetch NSURLIsExcludedFromBackupKey via includingPropertiesForKeys:, which lets you skip the setResourceValue:forKey: call entirely for paths that already have the correct value — reducing I/O on the already-problematic startup reconciliation walk.

This is particularly worth doing here since this method is called on the full SDK root directory tree.

Bugsnag/Client/BugsnagClient.m
SB-meiyalaganr: resolved please check

SB-jigneshR: @SB-meiyalaganr : Please address

SB-meiyalaganr: Resolved. The full recursive backup metadata reconciliation is now guarded by needsFileBackupSupportReconciliationForDirectory:fileBackupSupport:, and markFileBackupSupportReconciledForDirectory:fileBackupSupport: records completion after the scan. This prevents running applyFileBackupSupportToPathAndContents: on every app launch while still handling SDK upgrades and fileBackupSupport config changes.

SB-meiyalaganr (commented):
resolved

SB-jigneshR (approved):

Issue #1944, opened 2026-08-24
Build settings should be centralized in .xcconfig files


schriftgestalt:
There are a lot build settings scattered over the build targets. Most are redundant but some seem to be mistakes or are inconsistent.
This is related to #1856 

Will list the inconsistencies here:

The Watch targets have
`CLANG_CXX_LANGUAGE_STANDARD = gnu++17`
`GENERATE_INFOPLIST_FILE = YES`
all other targets have:
`CLANG_CXX_LANGUAGE_STANDARD = gnu++14`
`GENERATE_INFOPLIST_FILE = NO`

Only the the iOS framework target has:
`MODULEMAP_FILE = $(PROJECT_DIR)/Framework/module.modulemap`
this probably can be removed

the iOS and macOS have:
`CLANG_MODULES_AUTOLINK = NO`
all others (including the test targets have 
`CLANG_MODULES_AUTOLINK = YES`

only the WatchOS framework has
`SWIFT_EMIT_LOC_STRINGS = YES`
`CLANG_WARN_DIRECT_OBJC_ISA_USAGE = YES_ERROR
`CLANG_WARN_OBJC_ROOT_CLASS = YES_ERROR`

only the test targets have:
`SWIFT_VERSION = 4.2`

hannah-smartbear:
Hi @schriftgestalt,

Thanks for sharing this. As mentioned over on the PR that you've raised, our engineers will take a look and we'll let you know of any updates over on the PR.

PR #1945, opened 2026-08-24
move build setting into .xcconfig files

schriftgestalt:
- add BugsnagDebug.xcconfig
- remove most manually set build settings form the targets

see #1944

This makes it much easier to keep the settings in sync and the values should only be edited there (e.g. the version numbers)


hannah-smartbear:
Hi @schriftgestalt,

Thanks for taking the time to raise this PR. Our engineers will take a look at it, and leave any comments here accordingly.

PR #1951, opened 2026-08-30
Avoid eager Mach image enumeration during startup

### LINKED ISSUE #1950: Bugsnag.start synchronously enumerates every loaded Mach-O image during launch
### Describe the bug

`Bugsnag.start()` synchronously enumerates every Mach-O image already loaded in the process. This work runs in the application launch path and scales with the number of loaded images.

### Steps to reproduce

1. Integrate Bugsnag Cocoa 6.37.0 into an application with a substantial number of dynamic frameworks.
2. Start Bugsnag during application launch.
3. Profile `Bugsnag.start()` or add a signpost around it.
4. Observe `_dyld_register_func_for_add_image`, `add_image`, `bsg_mach_headers_populate_info`, and `dladdr` beneath the synchronous startup call.
5. Increase the number of loaded frameworks and observe the duration increase.

### Expected behavior

Starting Bugsnag should not perform work proportional to every loaded image on the launch thread. Binary-image reporting and address lookup should remain complete and crash safe, including for images loaded after Bugsnag starts.

### Suggested mitigation

Modern KSCrash removed this eager callback-populated linked list in [kstenerud/KSCrash#655](https://github.com/kstenerud/KSCrash/pull/655). Its [current binary-image cache](https://github.com/kstenerud/KSCrash/blob/master/Sources/KSCrashRecordingCore/KSBinaryImageCache.c) keeps `dyld_all_image_infos` as the source of truth, iterates the live dyld image array when writing reports, and uses a bounded allocation-free cache for address lookups.

### Environment

* Bugsnag version: 6.37.0
* CocoaPods version: not applicable to source benchmark
* Carthage version: not applicable
* OS version: macOS 26.5.2 for the source benchmark; the affected code is shared by Apple platforms
* Simulator or physical device: local macOS process
* Xcode version: 26.6
* Swift version: not applicable

naftaly:
Fixes #1950.

Bugsnag currently walks every loaded Mach-O image when it starts, adding launch work that grows with the size of the app. This uses dyld's live image list instead and resolves image details lazily, following the approach used by current KSCrash.

The change is deliberately narrow and preserves compact crash reports and images loaded after startup. Since this is a specific, measurable launch regression, would you please consider it for a quick bug-fix release once the draft is ready?

Known limitation: The image list can theoretically change while we read it, and unloaded images remain cached. These cases are extremely rare in normal apps and match current KSCrash behavior, so we accept the small risk to avoid slowing startup.

### DIFF
--- a/Bugsnag/Helpers/BSGRunContext.m
+++ b/Bugsnag/Helpers/BSGRunContext.m
@@ -67,9 +67,9 @@ static void InitRunContext(void) {
     // Make sure the images list is populated.
     bsg_mach_headers_initialize();
 
-    BSG_Mach_Header_Info *image = bsg_mach_headers_get_main_image();
-    if (image && image->uuid) {
-        uuid_copy(bsg_runContext->machoUUID, image->uuid);
+    BSG_Mach_Header_Info image;
+    if (bsg_mach_headers_get_main_image(&image) && image.uuid) {
+        uuid_copy(bsg_runContext->machoUUID, image.uuid);
     }
 
     NSString *bundleVersion = [[[NSBundle mainBundle] infoDictionary] objectForKey:@"CFBundleVersion"];
--- a/Bugsnag/Helpers/BSGTelemetry.m
+++ b/Bugsnag/Helpers/BSGTelemetry.m
@@ -22,7 +22,11 @@ static NSNumber *_Nullable IntegerValue(NSUInteger actual, NSUInteger defaultVal
 }
 
 static BOOL IsStaticallyLinked(void) {
-    return bsg_mach_headers_get_self_image() == bsg_mach_headers_get_main_image();
+    BSG_Mach_Header_Info selfImage;
+    BSG_Mach_Header_Info mainImage;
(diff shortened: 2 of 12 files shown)


SB-meiyalaganr:
@naftaly HI, Thanks for sharing.
We have been investigating the startup-performance issue reported in #1950. We tested the changes from your closed PR #1951 and observed a significant improvement in Match-image initialisation performance.

Before continuing with the implementation, could you please share why PR #1951 was closed? 
Were there any known correctness, crash-safety, or compatibility concerns with the proposed approach? 
We would also appreciate your recommendations on whether we should continue from your implementation or use a newer KSCrash version as the reference. 

cc: @SB-jigneshR

naftaly:
The implementation is incomplete and would require a lot more of my time than I can spend building it in to Bugsnag when it's already present in KSCrash 2.6.0 which is already released. I would suggest upgrading to KSCrash 2.6.0+ instead.

hannah-smartbear:
Hi @naftaly,

Thanks for letting us know.

Our engineers are continuing to work on this and we will let you know of any updates over on issue #1950.

PR #1956, opened 2026-09-08
Updating ChangeLogs

SB-meiyalaganr:
Update change logs v6.38.0


SB-meiyalaganr:
Only Change log info need to add in this PR, better release branch added chnagelog info there raised another PR for that. so this is not needed now closing.

